PR #268, opened 2023-09-23
Feature#170

albertmart000:
Implementación converter usando modelmapper.


MiquelDebon:
- Created a Generic Mapper using ModelMapper 3.1.1
- Implement this generic mapper in classes that use it
- Mapper Tested

PR #542, opened 2024-06-17
Feature#471a

ecs90:
Created zmqClient in Score micro, zmqServer in Challenge micro, made a request with the testing values using idChallenge and idLanguage as parameters. Also created necessary dtos (request and response)


jonatanvicente:
It seems it was made a good work. 
However, some config params are confusing (address and adress2). ZeroMQ allows communication over the same port in some patterns (in this situation we prefer using two, one for each channel). By that, names a bit more explicit are desirable (for example, address_server, address_client, etc).

Please, rename them.

PR #544, opened 2024-06-25
Update contributors.md

Alalilacias:
Added contributors that joined on the 25/06/2024


Alalilacias (commented):
Edited it so only my name was added.

PR #555, opened 2024-06-27
Feature#543

ecs90:
Test to verify if payload is too large.

Added test, properties and exception handler


jonatanvicente:
It doesn't work.
Config options added must be working (with both verbs PUT / POST), but a manual test allows send bigger request (attached screenshot).
May be necessary other options. In fact, config option "max-http-form-post-size" added is referring to the data encoded as key-value pairs sent, and represents the input fields of a form.... it seems it doesn't be what we need.

Pasted image

PR #556, opened 2024-06-27
Feature#535

danidiazd:
in UserSolutionService line 44,
added count from scoredtos.size to shows a real count.


danidiazd:
Added test to try function GetChallengeByUserID

PR #563, opened 2024-06-27
Feature#535

danidiazd:
Fix bug in micro-user


jonatanvicente:
This fix does not contain any test.
UserSolutionServiceImpTest contains several tests to test functionalities as addSolution (with many options), markAsBookmarked, etc.
It's not possible to accept any PR without tests over the new code, is a basic good practice.
Please, make the tests and request a new PR.

PR #574, opened 2024-07-23
Feature#471a

ecs90:
Renamed addresses as address_client and address_server depending on wich side is calling


jonatanvicente:
There are some conflicts on this PR

PR #624, opened 2024-10-29
Carlos Galan name and GitHub were added

camilo-arthuro:


jonatanvicente:
To include your user into contributors.txt you need change only that file. Review this, please :)

PR #645, opened 2024-11-07
Feature#579v0

nuriagrau:
- Implementation consists over ZMQ queues, both server and client at User Micro (client) and Score Micro (server).

  1.ZMQClient class in Micro User.

  2.ZMQServer class in Micro Score.

  3.Created or fixed the ZMQ Dto's in both Micros.

- When user micro get the results, persist them at MongoDB (score and errors fields; modify status as well).

- When request to Score is executed, status automatically should be 'SCORE_PENDING'. When the response is received, status will change to 'ENDED'.

    5.UserSolutionServiceImpl in Micro User

    6.Enums in Micro User

-Unitary testing for mqClient,  Service and Repository (Micro User).
-Unitary testing for mqServer, helper (Micro Score).


nuriagrau:
unwanted changed configuration files

PR #653, opened 2024-11-13
Feature#622

sebastiansdb:
Funcionando lo que pide la card. Se podrian agregar mas controles de errores


jonatanvicente:
This PR contains a conflicting file, please review this

PR #671, opened 2024-12-01
Feature#616

Gatchan1:
modified the ChallengeStatisticTotalDto DTO: removed count attribute and modified JSON field name for DTO attribute scorePending.


jonatanvicente:
The code does not working.
An invocation to endpoint throws 500-Internal Server Error.
Result must be (fake at least) at is showed at user.statistics.out.json
Pasted image (2)
Pasted image

PR #673, opened 2024-12-02
Feature#622

imoyaf:
Send two lists:
- Challenges completed. Array that contains elements with atts uuid_challenge and score of each one
- Challenges saved. Array that contains uuid_challenge saved by user
Testing. DTO classes and converter created.


imoyaf:
Testing coverage insufficient

PR #674, opened 2024-12-02
Feature#622

imoyaf:
Send two lists:
    Challenges completed. Array that contains elements with atts uuid_challenge and score of each one
    Challenges saved. Array that contains uuid_challenge saved by user

DTO classes and converter created.
Method created
Testing for getCompletedAndSavedChallengesStatistics method and converter methods.


jonatanvicente:
It seems the code is working properly, but some files contain wrong names (DTO names are different than other DTO global names). Please, review it.

Issue #622, opened 2024-10-29
User Business Logic: Statistic List


jonatanvicente:
**Please, keep an eye to branch feature#622**

**RELATIVE ESTIMATION:2**

We need to send two lists **filtered by user**:
    - Challenges completed. Array that contains elements with atts uuid_challenge and score of each one
    - Challenges saved. Array that contains uuid_challenge saved by user
    
* **Status available are**: 
     - **EMPTY** (user only set bookmarked to true)
     - **STARTED** (user start to resolve the challenge and save pieces of code, but not final response)
     - **SENT** (user has sent his solution; he cannot modify this solution).
     - **SCORE_PENDING** (Score micro has been requested to get scoring)
     - **ENDED** (score micro has sent final score).

**IMPLEMENTATION NOTES:**

- May be you need repository layer in addition
- Please, check performance when using lambda implementating this task (we don't recommend more than 2 levels of nesting, lower is desired)
- As usually, add testing you need, please. It's better if you want to use TDD

PR #698, opened 2024-12-18
Feature#687v0

nuriagrau:
It's necessary linking endpoint that throws statistic list (/itachallenge/api/v1/user/{idUser}/challenges/language/{idLanguage}/statistics/list) with service layer.
That is because Controller Layer contains fake data (hardcoded).

It depends on task https://github.com/IT-Academy-BCN/ita-challenges-backend/issues/622

### DIFF
--- a/itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
+++ b/itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
@@ -35,8 +35,12 @@ public class UserController {
 
     private static final Logger log = LoggerFactory.getLogger(UserController.class);
 
-    @Autowired
+
     private IUserSolutionService userScoreService;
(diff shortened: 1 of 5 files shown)


nuriagrau:
Added method 
**Mono<ChallengesListsDto> getCompletedAndSavedChallengesStatistics(String userId, String languageId)**
in **IUserSolutionService** and **UserSolutionServiceImp** as agreed with Ignasi from **feature#622** only for linking and testing purposes.

**UserController** modified including this method and test modified to fit new implementation.

ivilarop:
We are closing PR since the scoring service is no longer available with RoadMap 2025. In the future, they may be reopened.

Related archiving task:
https://github.com/IT-Academy-BCN/ita-challenges-backend/issues/731

PR #742, opened 2025-01-21
Onboarding update

Dxmrt:
Created branch and added name to contributors


ivilarop (changes requested):
I have now renamed the "feature" branch that was in origin and that is causing the conflict. Now you can create the feature/-onboarding branch.

PR #743, opened 2025-01-21
Onboarding update

amarinite:
He modificado el archivo contributors.md en la rama albamarquez-onboarding. Las guidelines especificaban que se creara una rama feature/<yourname>-onboarding, el problema es que aparentemente al ya existir una rama feature Git no nos permite crear sub-ramas bajo feature/ (entiendo que habría que modificar el enunciado de la tarea o lidiar con la rama feature)


ivilarop (changes requested):
I have now renamed the "feature" branch that was in origin and that is causing the conflict. Now you can create the feature/<yourname>-onboarding branch.

PR #782, opened 2025-01-30
Feature/763 challenge creation service method

Gatchan1:
Endpoint done (POST "/itachallenge/api/v1/challenge/challenges"), service method done, testing done for both;
Created the ChallengeCreateFormDto for the info that comes from the frontend.
Created an enum (DifficultyLevel) in the backend in order to restrict the "level" field in the mentioned Dto.
Added a new custom exception (ChallengeAlreadyExistsException) and handled both this one and HttpMessageNotReadableException (in the GlobalExceptionHandler).

Updated the postman collection with this endpoint and also updated the changelog.


itachallenge-challenge/src/main/resources/application.yml
```diff
@@ -9,8 +9,8 @@ spring:
     enabled: true
   data:
     mongodb:
-      #uri: mongodb://admin_challenge:BYBcMJEEWw5egRUo@localhost:27017/challenges?authSource=admin
-      uri: mongodb://test_challenge:SBM55CTYA8Ui32yC@87.106.229.175:27017/challenges?authSource=admin
+      uri: mongodb://admin_challenge:BYBcMJEEWw5egRUo@localhost:27017/challenges?authSource=admin
```
ivilarop: undo the change and leave it as it was

itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeController.java
```diff
@@ -225,6 +225,21 @@ public Mono<Map<String, Object>> addSolution(@Valid @RequestBody SolutionDto sol
...
+            summary = "Post a challenge providing the necessary data.",
+            description = "Sending the title, description, difficulty level, language and solution, a new challenge document will be inserted in the database.",
+            responses = {
+                    @ApiResponse(responseCode = "200", content = {@Content(schema = @Schema(implementation = ChallengeCreateFormDto.class), mediaType = "application/json")}),
+                    @ApiResponse(responseCode = "400", description = "Missing parameter(s)"),
+            }
+    )
+    public Mono<ResponseEntity<ChallengeDto>> addChallenge(@Valid @RequestBody ChallengeCreateFormDto createFormDto) {
+         return challengeService.addChallenge(createFormDto)
+                 .map(dto -> ResponseEntity.ok().body(dto));
```
ivilarop: .map(ResponseEntity::ok) is cleaner and more concise.

If you only return the DTO within ResponseEntity.ok, use .map(ResponseEntity::ok).
If you need to do more transformations inside .map(), use the explicit version with lambda.

itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/ChallengeCreateFormDto.java
```diff
@@ -0,0 +1,26 @@
...
+
+import com.itachallenge.challenge.enums.DifficultyLevel;
+import jakarta.validation.constraints.NotEmpty;
+import lombok.*;
+
+@AllArgsConstructor
+@Builder
+@Getter
+@Setter
+public class ChallengeCreateFormDto {
```
ivilarop: ChallengeCreateFormDto is functional, but "Form" is unnecessary and "CreateForm" is redundant.

ChallengeCreateDto is the best option because it is cleaner and clearer.

ivilarop: Done

itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeController.java
ivilarop: done

ivilarop (approved):

itachallenge-challenge/src/main/resources/application.yml
ivilarop: done

PR #808, opened 2025-02-24
Feature/775 user verification

davidcuadrado:
Added successful authentication logging and headers for front end.


amarinite (approved):
Headers are added both in AuthController and UserController, we should review if we could remove redundancies in future iterations

PR #810, opened 2025-02-27
Feature/85 add user role

jraporta:
Added a role attribute to the User model document.
Implemented a getUser endpoint to retrieve user details.

### DIFF
--- a/itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
+++ b/itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
@@ -1,6 +1,7 @@
 package com.itachallenge.user.controller;
 
 import com.itachallenge.user.annotations.ValidGithubUsername;
+import com.itachallenge.user.document.UserDocument;
 import com.itachallenge.user.service.UserService;
 import io.swagger.v3.oas.annotations.Operation;
 import io.swagger.v3.oas.annotations.Parameter;
@@ -34,6 +35,10 @@ public class UserController {
         return "Hello from ITA Challenge UserController!!!";
     }
 
+    /**
+     * @deprecated
+     * Database now stores more than just Mentors. Deletion should be handled upon completion of task #84 'Implementar roles de usuario'
+     */
     @Operation(
             summary = "Boolean validation for existing mentor",
             description = "Checks if a given GitHub username corresponds to an existing mentor in the database and returns a boolean response.",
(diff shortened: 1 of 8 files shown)


itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
```diff
@@ -98,6 +104,66 @@ public Mono<ResponseEntity<Boolean>> validateMentorExists(@RequestParam @ValidGi
...
+                            .header("X-Github-Username", githubUsername)
+                            .body(user);
+                })
+                .switchIfEmpty(Mono.fromCallable(() -> {
+                    log.warn("User not found: {}", githubUsername);
+                    return ResponseEntity
+                            .status(HttpStatus.NOT_FOUND)
+                            .header("X-Validation-Status", "Error")
+                            .header("X-Error-Message", "User not found")
+                            .body(null);
```
arnaufata, line 158: It might be better to return a body with a message rather than 'null'. For example:    .body(new ErrorResponse("User not found"));

itachallenge-user/src/main/java/com/itachallenge/user/service/UserService.java
```diff
@@ -13,8 +14,12 @@ public UserService(UserRepository userRepository) {
         this.userRepository = userRepository;
     }
 
-    public Mono<Boolean> isMentor(Mono<String> githubUsernameMono) {
+    public Mono<Boolean> exists(Mono<String> githubUsernameMono) {
         return githubUsernameMono.flatMap(userRepository::existsByUsername);
     }
 
+    public Mono<UserDocument> getUser(String githubUsername) {
```
arnaufata, line 21: Think about whether it would be necessary to add validation to prevent githubUsername from being null or empty.

arnaufata (approved):
I reviewed the PR and left some comments about error handling, but the current code works fine. I approve the PR!

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
jraporta: The header already contains this information, as does the 404 status. Additionally, the response is not received directly by the frontend but by the auth service, which already interprets the status code accordingly. Including a body may not be necessary in this case.

itachallenge-user/src/main/java/com/itachallenge/user/service/UserService.java
jraporta: Validation for githubUsername is managed by the @ValidGithubUsername annotation. It isn't working as expected, so may need to review or improve it to ensure proper validation.

ivilarop (approved):

PR #813, opened 2025-03-08
Feature/85 login return user details

jraporta:
The authenticate endpoint in the Auth microservice now returns a JWT token containing the encrypted Username and Role.


raulfabra20 (approved):

itachallenge-auth/src/main/resources/application.yml
ivilarop, line 1: why change ip for localhost?It is an IP of the docker environment

itachallenge-auth/build.gradle
ivilarop, line 1: Why delete mongodb dependencies?

ivilarop (changes requested):

itachallenge-auth/src/main/resources/application.yml
jraporta: I reverted the changes

itachallenge-auth/build.gradle
jraporta: The microservice doesn't have any repository, as the users are managed by the User Service, so the MongoDB dependencies weren't needed. I removed them to keep the dependencies clean and avoid unnecessary logs and warnings when starting the app.

ivilarop (approved):

PR #815, opened 2025-03-10
Link challenge to database mentor id

arnaufata:
- When a challenge is created, the frontend will now also send the username of the mentor who created it.

- This will allow us to check if the user exists, if it fulfills the ADMIN role, and retrieve its ID.

- To achieve this, we had to enable communication between the challenge microservice and the user microservice using a WebClient class.

- In this way, each challenge is associated with the mentor who created it using his/her ID.

### DIFF
--- a/itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/UserDto.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/UserDto.java
@@ -1,13 +1,13 @@
 package com.itachallenge.challenge.dto;
 
-import lombok.AllArgsConstructor;
-import lombok.Builder;
-import lombok.Data;
-import lombok.NoArgsConstructor;
+import lombok.*;
 
 import java.util.UUID;
 
-@Data
+@Getter
+@Setter
+@ToString
+@EqualsAndHashCode
 @Builder
 @NoArgsConstructor
 @AllArgsConstructor
--- a/itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/UserDto.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/UserDto.java
@@ -1,0 +1,18 @@
+package com.itachallenge.challenge.dto;
+
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+import java.util.UUID;
+
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class UserDto {
+    private UUID uuid;
+    private String username;
+    private String role;
+}
--- a/itachallenge-challenge/src/main/resources/application.yml
+++ b/itachallenge-challenge/src/main/resources/application.yml
@@ -66,3 +66,6 @@ validation:
 messages:
   errorMessage: "Invalid values."
 
+user:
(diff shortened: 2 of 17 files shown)


itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/UserDto.java
```diff
@@ -0,0 +1,18 @@
...
+
+import lombok.*;
+
+import java.util.UUID;
+
+@Getter
+@Setter
+@ToString
+@EqualsAndHashCode
+@Builder
```
jraporta, line 11: The @Data annotation can be used to replace the following annotations reducing the lines of code required:
@Getter
@Setter
@ToString
@EqualsAndHashCode
@Builder

itachallenge-challenge/src/main/resources/application.yml
```diff
@@ -66,3 +66,6 @@ validation:
 messages:
   errorMessage: "Invalid values."
 
+user:
+  service:
+    url: http://localhost:8764/itachallenge/api/v1/user
```
jraporta: should be the develop url

itachallenge-challenge/src/test/java/com/itachallenge/challenge/config/UserClientTest.java
```diff
@@ -0,0 +1,131 @@
...
+        when(responseSpec.bodyToMono(UserDto.class)).thenReturn(Mono.error(new RuntimeException("Server error in User Service")));
+
+        Mono<UserDto> result = userClient.getUserByUsername(username);
+
+        StepVerifier.create(result)
+                .expectError(RuntimeException.class)
+                .verify();
+    }
+
+    @Test
```
jraporta: It this test is testing the UserDto class it might be more appropiate to place it in a UserDtoTest class

jraporta (approved):

itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/UserDto.java
arnaufata: The @Data annotation causes problems with SonarQube as it considers the getters, setters, to String, EqualsAndHashCode and builder to be covered by the tests created. By making an annotation for each case, SonarQube does consider that the tests cover them.

itachallenge-challenge/src/main/resources/application.yml
arnaufata: The @Data annotation causes problems with SonarQube as it considers the getters, setters, to String, EqualsAndHashCode and builder to be covered by the tests created. By making an annotation for each case, SonarQube does consider that the tests cover them.

itachallenge-challenge/src/test/java/com/itachallenge/challenge/config/UserClientTest.java
arnaufata: Absolutely right!

itachallenge-challenge/src/main/resources/application.yml
arnaufata: I agree! I'll update it now.

ivilarop:
We're closing the PR because it's not applicable. At https://tree.taiga.io/project/luisvi-ita-challenges/task/132, we included the ID in the login response JWT, which is more optimal.

Issue #763, opened 2025-01-27
Challenge CRUD: Implement POST method for creating challenges


ivilarop:

ivilarop:
### Tasks:
- [x] Add logic in `ChallengeService` for creating challenges.
- [x] Add POST endpoint in `ChallengeController`.
- [x] Write unit tests for the POST method in `ChallengeService`.
- [x] Write unit tests for the POST endpoint in `ChallengeController`.
- [x] Update Postman collection with the POST endpoint.
- [x] Update documentation for the POST endpoint.

ivilarop:
5 puntos

Gatchan1:
Example of the json input needed for the POST endpoint:

> {
>     "challengeTitle": "títol",
>     "description": "descripció",
>     "level": "EASY",
>     "language": "Java",
>     "solution": "solution body (code)"
> }

Remember, mentors are supposed to enter the title and description in only one language (catalan).

- "language" needs to match an existing language name in the database (currently available languages are: "Javascript", "Java", "PHP" and "Python").
- "level" must match "EASY", "MEDIUM" or "HARD". (I created an enum (DifficultyLevel) in the backend in order to restrict the "level" field that comes through this endpoint).
- No field can be sent empty.
- Any of these discrepancies would result in a 400 Bad Request response.


---

The **error message** sent to the frontend when any or several fields are left empty is not a clean one. Improving it would apparently imply refactoring other endpoints/service methods/testing methods.
The other error messages look like this:
- "Invalid title: A challenge titled "Buscant els extrems" already exists"
- "Level "EASEY" is not valid"
- "Language "Llavascript" is not valid"

Gatchan1:
He dejado habilitada la uri de conexión a la database de mongo remota, tal y como estaba originalmente.

PR #817, opened 2025-03-18
Feature/126 manage favorite challenges

jraporta:
Create endpoint to add challenge to a User's favorite challenges.


Sh1ng0 (approved):
Good job, the code looks good and functional. Approved. 
Regarding security, I have some notes I'll review separately, but if I find anything critical, I'll mention it in another PR (Everything is alright in that regard and already talked about it with the reviewee)

itachallenge-challenge/src/main/java/com/itachallenge/challenge/util/JwtParser.java
ivilarop, line 1: Can you reuse JWTService by adding a method like extractClaim(token, claimName) or directly extractUuid(token)?

This way, you avoid duplicating logic, keep all token handling in one place, and ensure that signature validation is performed correctly.

itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeController.java
```diff
@@ -45,7 +46,10 @@ public class ChallengeController {
     private DiscoveryClient discoveryClient;
 
     @Autowired
-    IChallengeService challengeService;
+    private IChallengeService challengeService;
+
+    @Autowired
```
ivilarop, line 51: Can you reuse JWTService by adding a method like extractClaim(token, claimName) or directly extractUuid(token)?

itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeController.java
```diff
@@ -258,4 +262,38 @@ public Mono<ResponseEntity<DeleteResponseDto>> deleteOneChallenge(@PathVariable(
...
+    public Mono<ResponseEntity<FavoriteDto>> addChallengeToFavorite(
+            @PathVariable String challengeId,
+            @RequestHeader(name = "Authorization", required = false) String authHeader) {
+        String userId = getUserIdFromToken(authHeader);
+        return challengeService.addChallengeToFavorites(challengeId, userId)
+                .doOnError(error -> log.error("Error adding challenge to favorites {}", error.getMessage()))
+                .map(ResponseEntity::ok);
+    }
+
+    private String getUserIdFromToken(String authHeader) {
```
ivilarop, line 287: We could delegate all responsibility for token validation and parsing to a specialized service, i.e. your JWTService?

itachallenge-challenge/src/main/java/com/itachallenge/challenge/service/UserService.java
```diff
@@ -0,0 +1,61 @@
...
+    public UserService(
+            WebClient.Builder webClientBuilder,
+            @Value("${user.service.url}") String userServiceUrl) {
+        this.webClientBuilder = webClientBuilder;
+        this.userServiceUrl = userServiceUrl;
+    }
+
+    @Override
+    public Mono<Boolean> addChallengeToFavorites(String userId, String challengeId) {
+        String url = userServiceUrl + "/itachallenge/api/v1/user/users/" + userId + "/favorites/" + challengeId;
```
ivilarop: Do you take advantage of WebClient tools for secure route management?

AlexandraBonetCanela (commented):
Is Postman collection update missing?

itachallenge-challenge/src/main/java/com/itachallenge/challenge/exception/ChallengeNotFoundReturn404Exception.java
ivilarop, line 1: ChallengeNotFoundException.It's clear, direct, and standard. You'll handle the 404 with @ResponseStatus or in the GlobalExceptionHandler.

itachallenge-user/src/main/java/com/itachallenge/user/service/UserServiceImpl.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.Set;
+import java.util.UUID;
+import java.util.regex.Pattern;
+
+@Service
+public class UserServiceImpl implements UserService {
+
+    private final UserRepository userRepository;
+
+    private static final Pattern UUID_FORM = Pattern.compile("^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$", Pattern.CASE_INSENSITIVE);
```
ivilarop: UUID.fromString(...) now validates the UUID format. The regex is unnecessary and can be removed.

itachallenge-user/src/main/java/com/itachallenge/user/service/UserServiceImpl.java
```diff
@@ -0,0 +1,70 @@
...
+    @Override
+    public Mono<UserDocument> getUser(String githubUsername) {
+        return userRepository.findByUsername(githubUsername);
+    }
+
+    @Override
+    public Mono<Boolean> addChallengeToFavorites(String userId, String challengeId) {
+
+        return Mono.zip(validateUUID(userId), validateUUID(challengeId))
+                .flatMap(objects -> {
```
ivilarop: objects is not very descriptive — this can confuse the reader of the code.

itachallenge-user/src/main/java/com/itachallenge/user/service/UserServiceImpl.java
```diff
@@ -0,0 +1,70 @@
...
+
+        if (added) {
+            user.setFavoriteChallenges(favorites);
+            return userRepository.save(user).then(Mono.just(true));
+        }
+
+        return Mono.just(false);
+    }
+
+    private Mono<UUID> validateUUID(String id) {
```
ivilarop: If you only use it in this service: it's fine to leave it private in UserServiceImpl.

Consider renaming validateUUID to parseAndValidateUUID to reflect that it not only validates, but also converts.

itachallenge-challenge/src/main/resources/application.yml
ivilarop, line 1: Is it necessary to go through apisix if you are already making the call between microservices?

itachallenge-challenge/src/main/java/com/itachallenge/challenge/exception/CustomBadRequestException.java
ivilarop, line 1: Is the word custom necessary?

itachallenge-challenge/src/main/java/com/itachallenge/challenge/exception/CustomInternalServerErrorException.java
ivilarop, line 1: Is the word custom necessary?

itachallenge-user/src/main/java/com/itachallenge/user/exception/BadUUIDException.java
ivilarop, line 1: In Spring/WebFlux, custom exceptions are usually unchecked (RuntimeException) to flow better in reactive chains.

jraporta:
> Is Postman collection update missing?

Good point! I just updated the Postman collection.

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
```diff
@@ -76,25 +82,111 @@ public Mono<ResponseEntity<UserDocument>> getUser(@PathVariable @ValidGithubUser
...
                             .body(null));
                 });
     }
 
+    @Operation(
+            summary = "Add Challenge to User Favorite Challenges",
+            description = "Adds challenge to user favorites",
+            parameters = {
+                    @Parameter(
+                            name = "",
```
ivilarop: ??

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
```diff
@@ -76,25 +82,111 @@ public Mono<ResponseEntity<UserDocument>> getUser(@PathVariable @ValidGithubUser
...
+                            description = "User ID",
+                            required = true,
+                            in = ParameterIn.PATH
+                    ),
+                    @Parameter(
+                            name = "",
```
ivilarop: ??

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
```diff
@@ -76,25 +82,111 @@ public Mono<ResponseEntity<UserDocument>> getUser(@PathVariable @ValidGithubUser
...
+                    }
+                    log.info("User's '{}' favorites already contain Challenge '{}'", userId, challengeId);
+                    return ResponseEntity.ok()
+                            .header(X_FAVORITE_ADDED, FALSE)
+                            .header(X_FAVORITE_MESSAGE, "Challenge is already in favorites.")
+                            .body(false);
+                })
+                .onErrorResume(throwable -> {
+                    if (throwable instanceof NotFoundException) {
+                        log.error("No User not found with id: {}", userId);
```
ivilarop: This is not a technical "error", but an expected case → better with warn

ivilarop (changes requested):

ivilarop:
Please, add new functionality in the changelog, in the block of Issue #763

itachallenge-challenge/src/main/java/com/itachallenge/challenge/util/JwtParser.java
jraporta: I agree that duplicating the JWT logic is not ideal. However, this might be a temporary solution until a decision is made on how authentication will be handled in the app. Currently, no signature verification is performed when adding a challenge as a favorite. Additionally, JWTService belongs to another microservice, so using it from the challenge microservice would require creating a dedicated endpoint

itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeController.java
jraporta: As mentioned above, JWTService belongs to another microservice, so using it from the challenge microservice would require creating a dedicated endpoint. It necessary, it could be added as a future feature.

itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeController.java
jraporta: Replied in a previous comment.

itachallenge-challenge/src/main/java/com/itachallenge/challenge/exception/ChallengeNotFoundReturn404Exception.java
jraporta: The ChallengeNotFoundException is handled in the GlobalExceptionHandler, returning a 200 response. This is the current behavior when deleting a challenge. I believe a 404 response would be more appropriate, but since I shouldn't modify the expected behavior of the endpoint beyond the scope of this feature, I created a specific exception to return a 404 response. In a future PR, it might be worth standardizing the behavior across endpoints and using a single exception for consistency.

itachallenge-challenge/src/main/java/com/itachallenge/challenge/service/UserService.java
jraporta: Including the Authorization header may be necessary once the authentication strategy for the app is finalized. In any case, this endpoint should not be publicly accessible and should only be used by other microservices, which is something that also needs further consideration.
Adding the Authorization header would eliminate the need to pass the userId, delegating the responsibility of retrieving it to the User microservice. Currently, the Challenge microservice does not actually require the userId, so this approach could improve encapsulation.

itachallenge-challenge/src/main/resources/application.yml
jraporta: It is certainly not necessary, but I am not sure what url should be used to avoid this. I followed the same approach that is used in the Auth microservice to call the User microservice. It might be interesting to change it there too to avoid inconsistencies.

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
jraporta: corrected

itachallenge-user/src/main/java/com/itachallenge/user/exception/BadUUIDException.java
jraporta: corrected

itachallenge-user/src/main/java/com/itachallenge/user/service/UserServiceImpl.java
jraporta: corrected

itachallenge-user/src/main/java/com/itachallenge/user/service/UserServiceImpl.java
jraporta: corrected

itachallenge-user/src/main/java/com/itachallenge/user/service/UserServiceImpl.java
jraporta: corrected

itachallenge-challenge/src/main/java/com/itachallenge/challenge/util/JwtParser.java
ivilarop: I believe using a utility class like JwtParser spreads JWT-related logic in a way that breaks cohesion and makes future maintenance harder.

I propose replacing it with a proper JwtService inside the challenge microservice for the following reasons:

Security and scalability: Even though signature verification isn’t currently enforced, having a dedicated JwtService prepares us to handle proper token validation when needed (including verifying the signing key, expiration, etc.).

Cohesion: Encapsulating JWT logic inside a service, instead of a generic utility class, helps maintain a clean structure and adheres to the Single Responsibility Principle.

jraporta: I will apply the proposed improvements. However, this will imply sharing the secret string used to sign the JWT token between the different microservices... I will add it to the challenges microservice but it should be stored in a single place, probably passed as an environment variable.

ivilarop: I don't quite understand the secret thing. The current JwtParser isn't validating the token signature; it only extracts claims. What I propose is moving that logic to a JwtService in the challenges microservice to maintain cohesion and avoid loose utilities.

itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeController.java
ivilarop: Ok, fine, in the future we will see what options we have to unify the jwtService.

itachallenge-challenge/src/main/java/com/itachallenge/challenge/service/UserService.java
ivilarop: Just to clarify, when I asked about using WebClient tools for secure route management, I was actually referring to how the URL is currently being built in this method.

Concatenating strings to build the URL can be error-prone and harder to maintain. I’d suggest using UriComponentsBuilder instead, which is part of Spring and integrates nicely with WebClient. It allows you to build URLs in a safer and more readable way:

String url = UriComponentsBuilder.fromHttpUrl(userServiceUrl)
    .path("/itachallenge/api/v1/user/users/{userId}/favorites/{challengeId}")
    .buildAndExpand(userId, challengeId)
    .toUriString();

This avoids issues with slashes and makes it easier to manage changes in the path structure later on.

itachallenge-challenge/src/main/java/com/itachallenge/challenge/exception/ChallengeNotFoundReturn404Exception.java
ivilarop: Ok, now I understand... Okay, let's leave it like this, but please create a request in Taiga so we can audit this technical debt and we'll address it in future sprints. Thanks.

itachallenge-challenge/src/main/java/com/itachallenge/challenge/service/UserService.java
jraporta: Done

itachallenge-challenge/src/main/java/com/itachallenge/challenge/util/JwtParser.java
jraporta: I didn't understand it earlier, I retrieve the previous code from the JwtParser class.

ivilarop (approved):

PR #821, opened 2025-03-20
Feature/140 delete challenge from favorites

jraporta:
Add endpoint to remove a challenge from the user's favorite challenges

### DIFF
--- a/itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeController.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeController.java
@@ -284,6 +284,27 @@ public class ChallengeController {
                 .map(ResponseEntity::ok);
     }
 
+    @DeleteMapping("/challenges/{challengeId}/favorites")
+    @Operation(
+            operationId = "Remove a challenge from the User's favorites.",
+            summary = "Remove a challenge from favorites.",
+            description = "The ID Challenge sent through the URI is removed from the user's favorites. User Id is determined from the headers.",
+            responses = {
+                    @ApiResponse(responseCode = "200", content = {@Content(schema = @Schema(implementation = FavoriteDto.class), mediaType = "application/json")}),
+                    @ApiResponse(responseCode = "400", description = "Missing or invalid authorization header."),
+                    @ApiResponse(responseCode = "404", description = "The Challenge with given Id was not found."),
+                    @ApiResponse(responseCode = "500", description = "Internal Server Error")
+            }
+    )
+    public Mono<ResponseEntity<FavoriteDto>> removeChallengeFromFavorite(
+            @PathVariable String challengeId,
+            @RequestHeader(name = "Authorization", required = false) String authHeader) {
+        String userId = getUserIdFromToken(authHeader);
+        return challengeService.removeChallengeFromFavorites(challengeId, userId)
+                .doOnError(error -> log.error("Error removing challenge from favorites {}", error.getMessage()))
+                .map(ResponseEntity::ok);
+    }
+
     private String getUserIdFromToken(String authHeader) {
         if (authHeader == null || !authHeader.startsWith("Bearer ")) {
             log.warn("Missing or bad formatted Authorization header");
--- a/itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeController.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeController.java
@@ -4,6 +4,7 @@ import com.itachallenge.challenge.annotations.ValidGenericPattern;
 import com.itachallenge.challenge.config.PropertiesConfig;
 import com.itachallenge.challenge.dto.*;
 import com.itachallenge.challenge.exception.BadRequestException;
+import com.itachallenge.challenge.exception.JwtException;
 import com.itachallenge.challenge.service.IChallengeService;
 import com.itachallenge.challenge.service.JwtService;
 import io.swagger.v3.oas.annotations.Operation;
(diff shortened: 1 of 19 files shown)


itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeController.java
```diff
@@ -284,6 +284,27 @@ public Mono<ResponseEntity<FavoriteDto>> addChallengeToFavorite(
...
+            responses = {
+                    @ApiResponse(responseCode = "200", content = {@Content(schema = @Schema(implementation = FavoriteDto.class), mediaType = "application/json")}),
+                    @ApiResponse(responseCode = "400", description = "Missing or invalid authorization header."),
+                    @ApiResponse(responseCode = "404", description = "The Challenge with given Id was not found."),
+                    @ApiResponse(responseCode = "500", description = "Internal Server Error")
+            }
+    )
+    public Mono<ResponseEntity<FavoriteDto>> removeChallengeFromFavorite(
+            @PathVariable String challengeId,
+            @RequestHeader(name = "Authorization", required = false) String authHeader) {
```
ivilarop, line 303: Why is not required true?

itachallenge-challenge/src/main/java/com/itachallenge/challenge/service/ChallengeServiceImp.java
```diff
@@ -364,19 +364,51 @@ public Mono<FavoriteDto> addChallengeToFavorites(String challengeId, String user
...
+                                                Optional.ofNullable(challenge.getTimesFavorite()).orElse(0) == 0) {
+                                            return decreaseTimesFavorite(challenge);
+                                        }
+                                        return Mono.just(challenge);
+                                    })
+                                    .map(savedChallenge -> new FavoriteDto(false, savedChallenge.getTimesFavorite())));
+                });
+    }
+
+    private Mono<ChallengeDocument> increaseTimesFavorite(ChallengeDocument challenge) {
```
ivilarop: This improvement also applies to the creation endpoint (which I didn't notice in its PR).

The increaseTimesFavorite and decreaseTimesFavorite methods are now in the service, but since they modify the state of the object (ChallengeDocument), it's better to move them to the entity itself.

itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeController.java
```diff
@@ -284,6 +284,27 @@ public Mono<ResponseEntity<FavoriteDto>> addChallengeToFavorite(
...
+        String userId = getUserIdFromToken(authHeader);
+        return challengeService.removeChallengeFromFavorites(challengeId, userId)
+                .doOnError(error -> log.error("Error removing challenge from favorites {}", error.getMessage()))
+                .map(ResponseEntity::ok);
+    }
+
     private String getUserIdFromToken(String authHeader) {
```
ivilarop: I know this comes from a previous PR and I missed it due to timing issues, but I'm mentioning it now because it also impacts this endpoint:

The getUserIdFromToken(...) method is now in the controller, but the ideal would be to move it to the JWTService, right?

itachallenge-challenge/src/main/java/com/itachallenge/challenge/service/ChallengeServiceImp.java
```diff
@@ -364,19 +364,51 @@ public Mono<FavoriteDto> addChallengeToFavorites(String challengeId, String user
...
                                     })
                                     .map(savedChallenge -> new FavoriteDto(true, savedChallenge.getTimesFavorite())));
                 });
     }
 
-    private Mono<ChallengeDocument> updateTimesFavorite(ChallengeDocument challenge) {
+    @Override
+    public Mono<FavoriteDto> removeChallengeFromFavorites(String challengeId, String userId) {
+        Mono<UUID> challengeIdMono = validateUUID(String.valueOf(challengeId));
+        Mono<UUID> languageIdMono = validateUUID(String.valueOf(userId));
```
ivilarop, line 379: userIdMono ??

ivilarop:
I have a question: What happens if I delete a challenge? What effect does it have on my favorites?

ivilarop (changes requested):

jraporta:
> I have a question: What happens if I delete a challenge? What effect does it have on my favorites?

Right now, the list of favorite challenges is not updated. To fix this, the favorite list of each user should be iterated to find the id of the challenge and delete it.

itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeController.java
jraporta: The request is accepted even if authorization is not provided, and the response logic is handled in the code (returning a 400 Bad Request). If needed, required can be set to true, allowing Spring to manage the situation automatically. However, I am unsure of the exact response Spring would return in this case.

itachallenge-challenge/src/main/java/com/itachallenge/challenge/service/ChallengeServiceImp.java
jraporta: The validateUUID returning a userIdMono was already there, and is used by old code.

itachallenge-challenge/src/main/java/com/itachallenge/challenge/service/ChallengeServiceImp.java
jraporta: done

itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeController.java
jraporta: Done

itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeController.java
ivilarop: Okay, let's leave it like that. If we set it to true, Spring will automatically throw a MissingRequestHeaderException if the header is not included, which results in a 400 response (although the message may not be as explicit to the API consumer).

ivilarop:
> > I have a question: What happens if I delete a challenge? What effect does it have on my favorites?
> 
> Right now, the list of favorite challenges is not updated. To fix this, the favorite list of each user should be iterated to find the id of the challenge and delete it.


Ok, I'll add it to the requests to implement it in the future.

ivilarop (approved):

PR #825, opened 2025-03-27
Feature #197-Restore addSolution() method in UserController

g-lemoing:
This PR restores endpoint to add student solution, the endpoint returns mocked solution.

### DIFF
--- a/itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
+++ b/itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
@@ -29,6 +29,7 @@ public class UserController {
 
     private static final Logger log = LoggerFactory.getLogger(UserController.class);
     public static final String X_FAVORITE_ADDED = "X-Favorite-Added";
+    public static final String X_FAVORITE_DELETED = "X-Favorite-Deleted";
     public static final String X_FAVORITE_MESSAGE = "X-Favorite-Message";
     public static final String X_VALIDATION_STATUS = "X-Validation-Status";
     public static final String FALSE = "False";
@@ -215,6 +216,92 @@ public class UserController {
         );
 
         return Mono.just(ResponseEntity.status(HttpStatus.OK).body(userSolutionScoreDto));
+    }
+  
+    @Operation(
+            summary = "Delete Challenge from User Favorite Challenges",
+            description = "Deletes challenge from user favorites",
+            parameters = {
+                    @Parameter(
+                            name = "",
+                            description = "User ID",
+                            required = true,
+                            in = ParameterIn.PATH
+                    ),
+                    @Parameter(
+                            name = "",
+                            description = "Challenge ID",
+                            required = true,
+                            in = ParameterIn.PATH
+                    )
+            },
+            responses = {
+                    @ApiResponse(
+                            responseCode = "200",
+                            description = "No change, challenge was not in favorites",
+                            content = @Content(mediaType = "application/json")
+                    ),
+                    @ApiResponse(
+                            responseCode = "201",
+                            description = "Challenge deleted from favorites",
+                            content = @Content(mediaType = "application/json")
+                    ),
+                    @ApiResponse(
+                            responseCode = "400",
+                            description = "Bad Request. The provided IDs have a bad format",
+                            content = @Content(mediaType = "application/json")
+                    ),
+                    @ApiResponse(
+                            responseCode = "404",
+                            description = "Not found. No user is found with the provided user id.",
+                            content = @Content(mediaType = "application/json")
+                    ),
+                    @ApiResponse(
+                            responseCode = "500",
+                            description = "Internal server error. An unexpected error occurred.",
+                            content = @Content(mediaType = "application/json")
+                    )
+            }
+    )
+    @DeleteMapping("/users/{userId}/favorites/{challengeId}")
+    public Mono<ResponseEntity<Boolean>> deleteFromFavorites(@PathVariable String userId, @PathVariable String challengeId) {
+        return userService.deleteChallengeFromFavorites(userId, challengeId)
+                .map(deleted -> {
+                    if (Boolean.TRUE.equals(deleted)) {
+                        log.info("Challenge '{}' deleted from user '{}' favorites", challengeId, userId);
+                        return ResponseEntity.status(HttpStatus.CREATED)
+                                .header(X_FAVORITE_DELETED, "True")
+                                .header(X_FAVORITE_MESSAGE, "Challenge deleted from favorites.")
+                                .body(true);
+                    }
+                    log.info("No change, User's '{}' favorites doesn't contain Challenge '{}'", userId, challengeId);
+                    return ResponseEntity.ok()
+                            .header(X_FAVORITE_DELETED, FALSE)
+                            .header(X_FAVORITE_MESSAGE, "Challenge not found in user's favorites.")
+                            .body(false);
+                })
+                .onErrorResume(throwable -> {
+                    if (throwable instanceof NotFoundException) {
+                        log.error("No User not found with id: {}", userId);
+                        return Mono.just(ResponseEntity.status(HttpStatus.NOT_FOUND)
+                                .header(X_FAVORITE_DELETED, FALSE)
+                                .header(X_FAVORITE_MESSAGE, "User not found.")
+                                .body(false));
+                    }
+                    if (throwable instanceof BadUUIDException) {
+                        log.error("The provided IDs are not valid.");
+                        return Mono.just(ResponseEntity.status(HttpStatus.BAD_REQUEST)
+                                .header(X_FAVORITE_DELETED, FALSE)
+                                .header(X_FAVORITE_MESSAGE, "The provided IDs are not valid.")
+                                .body(false));
+                    }
+                    log.error("Unexpected error: {}", throwable.getMessage());
+                    return Mono.just(ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR)
+                            .header(X_FAVORITE_DELETED, FALSE)
+                            .header(X_FAVORITE_MESSAGE, "Unexpected server error.")
+                            .body(false));
+                });
+
     }
 
 }
--- a/itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
+++ b/itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
@@ -2,6 +2,8 @@ package com.itachallenge.user.controller;
 
 import com.itachallenge.user.annotations.ValidGithubUsername;
 import com.itachallenge.user.document.UserDocument;
+import com.itachallenge.user.dto.UserSolutionDto;
+import com.itachallenge.user.dto.UserSolutionScoreDto;
 import com.itachallenge.user.exception.BadUUIDException;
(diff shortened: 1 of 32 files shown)


itachallenge-user/src/main/java/com/itachallenge/user/dto/UserSolutionDto.java
Sh1ng0, line 1: With lombok @Data there's no need for Getter, Setter (or RequiredArgsConstructor for that matter) in turn making the code a bit more compact

Sh1ng0 (approved):
It looks good to me, reviwee said this code is somewhat provisional, and it gets the job done!

itachallenge-user/src/main/java/com/itachallenge/user/dto/UserSolutionDto.java
g-lemoing: Yes, you're right, José. But I had to remove @Data annotation to satisfy SonarQube filtering.

itachallenge-user/src/main/java/com/itachallenge/user/dto/UserSolutionScoreDto.java
ivilarop, line 1: Considering that we won't be working with solution scoring at the moment.

itachallenge-user/src/main/java/com/itachallenge/user/exception/UserGlobalExceptionHandler.java
ivilarop, line 1: Are these changes applicable to this PR? Is it related to the task objective?

itachallenge-user/src/main/java/com/itachallenge/user/dto/UserSolutionScoreDto.java
```diff
@@ -0,0 +1,30 @@
...
+    @JsonProperty(value ="uuid_challenge")
+    private String challengeId;
+
+    @JsonProperty(value ="uuid_language")
+    private String languageId;
+
+    @JsonProperty(value ="solution_text")
+    private String solutionText;
+
+    @JsonProperty(value = "score")
```
ivilarop: does not apply score

itachallenge-user/src/main/java/com/itachallenge/user/dto/UserSolutionDto.java
```diff
@@ -0,0 +1,38 @@
...
+    @JsonProperty(value ="uuid_challenge")
+    @GenericUUIDValid(message = "Invalid UUID")
+    private String challengeId;
+
+    @JsonProperty(value ="uuid_language")
+    @GenericUUIDValid(message = "Invalid UUID")
+    private String languageId;
+
+    @JsonProperty(value ="status")
+    private String status;
```
ivilarop: Is it clear at a functional level how the different states will be defined? Are they defined in the taiga task?

itachallenge-user/src/test/java/com/itachallenge/user/dto/UserSolutionScoreDtoTest.java
ivilarop, line 1: Considering that we won't be working with solution scoring at the moment

itachallenge-user/src/test/resources/application-test.yml
ivilarop, line 1: Why this change?

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
```diff
@@ -190,6 +193,33 @@ public Mono<ResponseEntity<Boolean>> addToFavorites(@PathVariable String userId,
...
+        final int score = 0;
+        UserSolutionScoreDto userSolutionScoreDto = new UserSolutionScoreDto(
+                userSolutionDto.getUserId(),
+                userSolutionDto.getChallengeId(),
+                userSolutionDto.getLanguageId(),
+                userSolutionDto.getSolutionText(),
+                score
+        );
+
+        return Mono.just(ResponseEntity.status(HttpStatus.OK).body(userSolutionScoreDto));
```
ivilarop: review userSolutionScoreDto

ivilarop (changes requested):

ivilarop:
Remember to include the changes in changelog.md. We don't have a final version yet, but add it to the first block there.

itachallenge-user/src/test/resources/application-test.yml
g-lemoing: This configuration variable is needed to validate UUID pattern of UserSolutionDto (request body of new endpoint).
I had to add it to be available for test environment.

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
g-lemoing: score attribute removed from UserSolutionScoreDto --> UserController modified consequently.

itachallenge-user/src/main/java/com/itachallenge/user/dto/UserSolutionDto.java
g-lemoing: 2 states will be defined in further PR: STARTED and ENDED, documented in task #197

itachallenge-user/src/main/java/com/itachallenge/user/dto/UserSolutionScoreDto.java
g-lemoing: OK, Attribute score removed from UserSolutionScoreDto

itachallenge-user/src/main/java/com/itachallenge/user/exception/UserGlobalExceptionHandler.java
g-lemoing: Maybe not directly related to the task objective but I detected while doing tests, that the exception MethodArgumentNotValidException was not caught and didn't return the correct HTTP error code. This code is to fix it.

itachallenge-user/src/main/java/com/itachallenge/user/dto/UserSolutionScoreDto.java
g-lemoing: Attribute score removed from UserSolutionScoreDto class.

itachallenge-user/src/test/java/com/itachallenge/user/dto/UserSolutionScoreDtoTest.java
g-lemoing: Adaptations applied to test class refering to removal of attribute score in UserSolutionScoreDto

itachallenge-user/src/main/java/com/itachallenge/user/dto/UserSolutionDto.java
ivilarop: Let's keep in mind that for now, and for this sprint, we would only consider the finished state, which represents the submit solution button. If we have time, we would optionally consider the started state if the save challenge button from the front end is working. If you have any questions, let us know.

itachallenge-user/src/main/java/com/itachallenge/user/dto/UserSolutionScoreDto.java
ivilarop: I was referring to whether this DTO makes sense. If we eliminate the score, wouldn't it be duplicated with UserSolutionDto?

itachallenge-user/src/main/java/com/itachallenge/user/exception/UserGlobalExceptionHandler.java
ivilarop: In Spring, when you annotate a parameter with @Valid @RequestBody, as in your addSolution method, and the DTO's validations are not met, Spring automatically throws a MethodArgumentNotValidException exception.

By default, this exception is typically handled by Spring's global exception handler, which returns a 400 Bad Request with a standard error body. However, if you don't have a custom handler, the message may be generic and not as useful to the frontend or users.

Is it a good practice to let Spring handle this itself? Please check it out and let me know.

itachallenge-user/src/main/java/com/itachallenge/user/dto/UserSolutionScoreDto.java
ivilarop: I was referring to whether this DTO makes sense. If we eliminate the score, wouldn't it be duplicated with UserSolutionDto?

itachallenge-user/src/test/java/com/itachallenge/user/dto/UserSolutionScoreDtoTest.java
ivilarop: I was referring to whether this DTO makes sense. If we eliminate the score, wouldn't it be duplicated with UserSolutionDto?

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
ivilarop: I was referring to whether this DTO makes sense. If we eliminate the score, wouldn't it be duplicated with UserSolutionDto?

ivilarop (changes requested):

itachallenge-user/src/main/java/com/itachallenge/user/exception/UserGlobalExceptionHandler.java
g-lemoing: In our case, I think the MethodArgumentNotValidException is clear enough to avoid creating a custom exception for this purpose.

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
g-lemoing: It makes sense because fields from both dtos are not exactly the same. But to make class names more meaningful, refactor has been added:
UserSolutionDto renamed to UserSolutionRequestDto,
UserSolutionScoreDto renamed to UserSolutionResponseDto.

itachallenge-user/src/main/java/com/itachallenge/user/dto/UserSolutionDto.java
g-lemoing: Only solution status "ENDED" will be left for this sprint but this modification will be included in next PR of task #198, because current PR does not validate solution status yet.

itachallenge-user/src/main/java/com/itachallenge/user/dto/UserSolutionScoreDto.java
g-lemoing: It makes sense because fields from both dtos are not exactly the same. But to make class names more meaningful, refactor has been added:
UserSolutionDto renamed to UserSolutionRequestDto,
UserSolutionScoreDto renamed to UserSolutionResponseDto.

ivilarop: Ok perfect! thks

itachallenge-user/src/main/java/com/itachallenge/user/exception/UserGlobalExceptionHandler.java
ivilarop: ok understood

itachallenge-user/src/main/java/com/itachallenge/user/dto/UserSolutionDto.java
ivilarop: ok

ivilarop (approved):

Sh1ng0:
And Lombok still has problems with sonar when it comes to the "Data" anotation

PR #826, opened 2025-04-01
Feature/#181 user favorites endpoint

paumaneja:
Create a new GET endpoint in the User microservice, which, given a user, returns a list of ChallengeId (ChallengeDto) that the user has saved as favorites.

### DIFF
--- a/itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
+++ b/itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
@@ -19,6 +19,9 @@ import org.springframework.validation.annotation.Validated;
 import org.springframework.web.bind.annotation.*;
 import reactor.core.publisher.Mono;
 
+import java.util.Set;
+import java.util.UUID;
+
 @RestController
 @Validated
 @RequestMapping(value = "/itachallenge/api/v1/user")
@@ -28,6 +31,8 @@ public class UserController {
     public static final String X_FAVORITE_ADDED = "X-Favorite-Added";
     public static final String X_FAVORITE_MESSAGE = "X-Favorite-Message";
     public static final String X_VALIDATION_STATUS = "X-Validation-Status";
+    public static final String X_ERROR_MESSAGE = "X-Error-Message";
+    public static final String X_GITHUB_USERNAME ="X-Github-Username";
     public static final String FALSE = "False";
 
     private final UserService userService;
@@ -83,7 +88,7 @@ public class UserController {
                     log.info("User found: {} (Role: {})", githubUsername, user.getRole());
                     return ResponseEntity.ok()
                             .header(X_VALIDATION_STATUS, "Success")
-                            .header("X-Github-Username", githubUsername)
+                            .header(X_GITHUB_USERNAME, githubUsername)
                             .body(user);
                 })
                 .switchIfEmpty(Mono.fromCallable(() -> {
@@ -91,14 +96,14 @@ public class UserController {
                     return ResponseEntity
                             .status(HttpStatus.NOT_FOUND)
                             .header(X_VALIDATION_STATUS, "Error")
-                            .header("X-Error-Message", "User not found")
+                            .header(X_ERROR_MESSAGE, "User not found")
                             .body(null);
                 }))
                 .onErrorResume(e -> {
                     log.error("Error retrieving user '{}': {}", githubUsername, e.getMessage());
                     return Mono.just(ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR)
                             .header(X_VALIDATION_STATUS, "Error")
-                            .header("X-Error-Message", "An error occurred retrieving user.")
+                            .header(X_ERROR_MESSAGE, "An error occurred retrieving user.")
                             .body(null));
                 });
     }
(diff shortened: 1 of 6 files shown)


itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
```diff
@@ -275,4 +280,55 @@ public Mono<ResponseEntity<Boolean>> deleteFromFavorites(@PathVariable String us
...
+                    @ApiResponse(responseCode = "400", description = "Invalid UUID format"),
+                    @ApiResponse(responseCode = "500", description = "Unexpected error")
+            }
+    )
+
+    @GetMapping("/users/{userId}/favorites")
+    public Mono<ResponseEntity<Set<UUID>>> getUserFavorites(@PathVariable String userId) {
+        return userService.getUserFavorites(userId)
+                .map(favorites -> {
+                    log.info("Retrieved {} favorite challenges for user {}", favorites.size(), userId);
```
Almami679, line 335: It's fine to use logs in the development environment, but when doing PR I would delete them.

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
```diff
@@ -275,4 +280,55 @@ public Mono<ResponseEntity<Boolean>> deleteFromFavorites(@PathVariable String us
...
+            parameters = {
+                    @Parameter(
+                            name = "userId",
+                            description = "UUID of the user",
+                            required = true,
+                            in = ParameterIn.PATH
+                    )
+            },
+            responses = {
+                    @ApiResponse(responseCode = "200", description = "List of favorite challengeIds by user"),
```
Almami679: I would make it clearer that the return is Set<UUID>

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
```diff
@@ -275,4 +280,55 @@ public Mono<ResponseEntity<Boolean>> deleteFromFavorites(@PathVariable String us
...
+    @GetMapping("/users/{userId}/favorites")
+    public Mono<ResponseEntity<Set<UUID>>> getUserFavorites(@PathVariable String userId) {
+        return userService.getUserFavorites(userId)
+                .map(favorites -> {
+                    log.info("Retrieved {} favorite challenges for user {}", favorites.size(), userId);
+                    return ResponseEntity.ok()
+                            .header(X_VALIDATION_STATUS, "Success")
+                            .body(favorites);
+                })
+                .onErrorResume(throwable -> {
```
Almami679, line 340: Shouldn't these cases be handled from the exception handler? Maybe you should use the buildErrorResponse() method.

Almami679 (changes requested):
Clear and short PR, not like mine ahha

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
paumaneja: You're right, but I left the logs out of consistency with the latest changes applied to everything related to Favorites that the previous colleague had implemented. They could be deleted but all the methods that had already been implemented and also those that @Alexandra is implementing related to Bookmarks should also be changed.

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
paumaneja: updated

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
paumaneja: The same as the change you mention in the logs. I implemented it like this for consistency with what was already implemented. If Ivan thinks it's better as you mention, it should be applied to the rest of the Favorites and Bookmarks methods.

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
Almami679: perfect, I didn't have any context for everything

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
Almami679: Ok, Ivan's turn.

itachallenge-user/src/main/java/com/itachallenge/user/service/UserServiceImpl.java
```diff
@@ -94,4 +94,14 @@ private Mono<UUID> parseAndValidateUUID(String id) {
         }
     }
 
+    @Override
+    public Mono<Set<UUID>> getUserFavorites(String userId) {
+        return parseAndValidateUUID(userId)
+                .flatMap(userUuid ->
+                        userRepository.findById(userUuid)
+                                .switchIfEmpty(Mono.error(new NotFoundException("User not found.")))
```
ivilarop: for more info:new NotFoundException("User not found with id: " + userId))

ivilarop (changes requested):

ivilarop:
add the issue to changelog.md

itachallenge-user/src/main/java/com/itachallenge/user/service/UserServiceImpl.java
paumaneja: updated

ivilarop:
> add the issue to changelog.md

Done!

ivilarop (approved):

PR #827, opened 2025-04-02
Add bookmark to User - User Microservice

AlexandraBonetCanela:
Implemented "A User can add a bookmark" functionality in the User Microservice, through a new endpoint (addToBookmarks).

Opted for mimicking the addToFavorite behaviour for better consistency and discarded the option of restoring the former deleted addBookmarks endpoint since addToFavorite is more completed and up to date with our current goals and restoring the former endpoint could create some discrepancy in behaviour between these two entites. 

Created a new set of tests for testing bookmark functionality and modified the ones that needed to too.

Left to do in the future:
Parametrized tests (so we could unify both Favorite and Bookmark test batches into a single one).

### DIFF
--- a/itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
+++ b/itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
@@ -119,13 +119,13 @@ public class UserController {
             description = "Adds challenge to user favorites",
             parameters = {
                     @Parameter(
-                            name = "User ID",
+                            name = "userId",
                             description = "User ID",
                             required = true,
                             in = ParameterIn.PATH
                     ),
                     @Parameter(
-                            name = "Challenge ID",
+                            name = "challengeId",
                             description = "Challenge ID",
                             required = true,
                             in = ParameterIn.PATH
@@ -230,13 +230,13 @@ public class UserController {
             description = "Adds challenge to user Bookmarks",
             parameters = {
                     @Parameter(
-                            name = "User ID",
+                            name = "userId",
                             description = "User ID",
                             required = true,
                             in = ParameterIn.PATH
                     ),
                     @Parameter(
-                            name = "Challenge ID",
+                            name = "challengeId",
                             description = "Challenge ID",
                             required = true,
                             in = ParameterIn.PATH
@@ -317,13 +317,13 @@ public class UserController {
             description = "Deletes challenge from user favorites",
             parameters = {
                     @Parameter(
-                            name = "",
+                            name = "userId",
                             description = "User ID",
                             required = true,
                             in = ParameterIn.PATH
                     ),
                     @Parameter(
-                            name = "",
+                            name = "challengeId",
                             description = "Challenge ID",
                             required = true,
                             in = ParameterIn.PATH
--- a/itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
+++ b/itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
@@ -273,19 +273,19 @@ public class UserController {
 
     @PostMapping("/users/{userId}/bookmarks/{challengeId}")
     public Mono<ResponseEntity<Boolean>> addToBookmarks(@PathVariable String userId, @PathVariable String challengeId) {
-        return userService.addChallengeToFavorites(userId, challengeId)
+        return userService.addChallengeToBookmarks(userId, challengeId)
                 .map(added -> {
                     if (Boolean.TRUE.equals(added)) {
                         log.info("Challenge '{}' added to user '{}' favorites", challengeId, userId);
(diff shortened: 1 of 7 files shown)


itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
```diff
@@ -190,6 +192,93 @@ public Mono<ResponseEntity<Boolean>> addToFavorites(@PathVariable String userId,
...
+                    )
+            }
+    )
+
+    @PostMapping("/users/{userId}/bookmarks/{challengeId}")
+    public Mono<ResponseEntity<Boolean>> addToBookmarks(@PathVariable String userId, @PathVariable String challengeId) {
+        return userService.addChallengeToBookmarks(userId, challengeId)
+                .map(added -> {
+                    if (Boolean.TRUE.equals(added)) {
+                        log.info("Challenge '{}' added to user '{}' favorites", challengeId, userId);
```
paumaneja: bookmark instead of favorites...

itachallenge-user/src/test/java/com/itachallenge/user/controller/UserControllerTest.java
```diff
@@ -177,6 +231,24 @@ void addToFavorites_WhenBadFormattedId_Returns404() {
...
+                .uri("/itachallenge/api/v1/user/users/" + userId + "/bookmarks/" + challengeId)
+                .exchange()
+                .expectStatus().isEqualTo(HttpStatus.BAD_REQUEST)
+                .expectHeader().valueEquals("X-Bookmark-Added", "False")
+                .expectHeader().valueEquals("X-Bookmark-Message", "The provided IDs are not valid.")
+                .expectBody(Boolean.class).isEqualTo(false);
+
+        verify(userService, times(1)).addChallengeToBookmarks(userId, challengeId);
+    }
+
```
paumaneja: the endpoint should return a 400 error, right? Now checking it I saw that Juanra also had it like that

paumaneja (changes requested):
minor changes

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
AlexandraBonetCanela: Correct! Applied changes

itachallenge-user/src/test/java/com/itachallenge/user/controller/UserControllerTest.java
AlexandraBonetCanela: Well Spotted! Changed!

paumaneja (approved):

itachallenge-user/src/main/java/com/itachallenge/user/document/UserDocument.java
ivilarop, line 1: I see that the favoriteChallenges part is missing from the toString

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
```diff
@@ -218,6 +220,93 @@ public Mono<ResponseEntity<UserSolutionResponseDto>> addSolution(
         return Mono.just(ResponseEntity.status(HttpStatus.OK).body(userSolutionResponseDto));
     }
   
+    @Operation(
+            summary = "Add Challenge to User Bookmark Challenges",
+            description = "Adds challenge to user Bookmarks",
+            parameters = {
+                    @Parameter(
+                            name = "User ID",
```
ivilarop: @Parameter(name = "...") should use userId and challengeId (the real names).Swagger uses name to match @PathVariable, if they don't match it can break the generated documentation.I think that in other implementations it is not like this but we have recently discovered it and from now on we will adapt it like this.

ivilarop (changes requested):

itachallenge-user/src/main/java/com/itachallenge/user/document/UserDocument.java
AlexandraBonetCanela: @ivilarop  Could you double check? Cause I do see favoriteChallenges sitting just above bookmarkChallenges inside the toString method
Captura de pantalla 2025-04-03 a les 10 13 31

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
AlexandraBonetCanela: Updated parameters for all endpoints

itachallenge-user/src/main/java/com/itachallenge/user/document/UserDocument.java
ivilarop: Correct! Sorry for the mistake.

ivilarop (approved):

PR #831, opened 2025-04-03
Feature/205 bookmark delete endpoint user microservice

AlexandraBonetCanela:
Added bookmark DELETE endpoint following same favorite endpoint fashion.

Also fix favorite DELETE endpoint 201 Created response to return 200.

### DIFF
--- a/itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
+++ b/itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
@@ -333,12 +333,7 @@ public class UserController {
             responses = {
                     @ApiResponse(
                             responseCode = "200",
-                            description = "No change, challenge was not in favorites",
-                            content = @Content(mediaType = "application/json")
-                    ),
-                    @ApiResponse(
-                            responseCode = "201",
-                            description = "Challenge deleted from favorites",
+                            description = "Challenge deleted from favorites or was not in favorites",
                             content = @Content(mediaType = "application/json")
                     ),
                     @ApiResponse(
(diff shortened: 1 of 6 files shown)


paumaneja:
I have no specific comment.
Clearer PR than the previous one and following the thread of what was previously implemented in this microservice

paumaneja (approved):

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
```diff
@@ -332,12 +333,7 @@ public Mono<ResponseEntity<Boolean>> addToBookmarks(@PathVariable String userId,
             responses = {
                     @ApiResponse(
                             responseCode = "200",
-                            description = "No change, challenge was not in favorites",
```
ivilarop, line 335: The change to the favorites endpoint (favorite DELETE endpoint) should have been addressed in a separate PR. It's technical debt.

itachallenge-user/src/main/java/com/itachallenge/user/service/UserService.java
```diff
@@ -15,8 +15,8 @@ public interface UserService {
 
     Mono<Boolean> deleteChallengeFromFavorites(String userId, String challengeId);
 
-    Mono<Set<UUID>> getUserFavorites(String userId);
-
+    Mono<Boolean> deleteChallengeFromBookmarks(String userId, String challengeId);
 
+    Mono<Set<UUID>> getUserFavorites(String userId);
```
ivilarop, line 18: be careful! getUserFavorites already existed and seems to have been moved without any functional changes. While the code is still correct, unnecessarily moving methods can make it harder to track history  and adds unnecessary noise to the PR.

ivilarop (changes requested):

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
AlexandraBonetCanela: Noted!

itachallenge-user/src/main/java/com/itachallenge/user/service/UserService.java
AlexandraBonetCanela: Applied changes

ivilarop (approved):

PR #833, opened 2025-04-08
refactor UserController to handle exceptions from GlobalExceptionHandler

paumaneja:
I've deleted quite a few lines of code, and now everything is clearer, but there's less information in case errors arise. Perhaps I could have just removed the .OnErrorResume...what do you think?
I've left the @Operations as they were; perhaps they could also be summarized.


paumaneja:
After discussing with the mentor, we've decided to close this pull request and start fresh with a new branch.

The initial refactor was done too quickly and without a clear, structured plan, which has led to too many follow-up changes and adjustments. Continuing with this PR would result in a messy history and unclear purpose.

We'll now approach the refactor incrementally and with cleaner commits to ensure better traceability and reviewability.

PR #836, opened 2025-04-09
refactor UserController and modified test

paumaneja:
Description:

The UserController will be refactored to remove the error handling logic with onErrorResume, delegating that responsibility to the already implemented UserGlobalExceptionHandler.

Tasks to be performed:

1. Remove onErrorResume blocks from the controller.

2. Ensure that all exceptions are correctly handled by the global handler.

3. Keep custom headers only if necessary.

4. Adjust the tests to:

- Inject the UserGlobalExceptionHandler into WebTestClient
- Do not expect headers previously handled by the controller
- Correctly validate the expected status codes and bodies.

5. Confirm that all functionality continues to work correctly after the refactoring.

### DIFF
--- a/itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
+++ b/itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
@@ -4,7 +4,6 @@ import com.itachallenge.user.annotations.ValidGithubUsername;
 import com.itachallenge.user.document.UserDocument;
 import com.itachallenge.user.dto.UserSolutionRequestDto;
 import com.itachallenge.user.dto.UserSolutionResponseDto;
-import com.itachallenge.user.exception.NotFoundException;
 import com.itachallenge.user.service.IUserSolutionService;
 import com.itachallenge.user.service.UserService;
 import io.swagger.v3.oas.annotations.Operation;
@@ -85,7 +84,6 @@ public class UserController {
     @GetMapping("/users/{githubUsername}")
     public Mono<ResponseEntity<UserDocument>> getUser(@PathVariable @ValidGithubUsername String githubUsername) {
         return userService.getUser(githubUsername)
-                .switchIfEmpty(Mono.error(new NotFoundException("User not found")))
                 .map(user -> {
                     log.info("User found: {} (Role: {})", githubUsername, user.getRole());
                     return ResponseEntity.ok()
--- a/itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
+++ b/itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
@@ -4,7 +4,6 @@ import com.itachallenge.user.annotations.ValidGithubUsername;
 import com.itachallenge.user.document.UserDocument;
 import com.itachallenge.user.dto.UserSolutionRequestDto;
 import com.itachallenge.user.dto.UserSolutionResponseDto;
-import com.itachallenge.user.exception.BadUUIDException;
 import com.itachallenge.user.exception.NotFoundException;
 import com.itachallenge.user.service.UserService;
 import io.swagger.v3.oas.annotations.Operation;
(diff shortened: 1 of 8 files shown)


itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
```diff
@@ -489,7 +360,7 @@ public Mono<ResponseEntity<Boolean>> deleteFromBookmarks(@PathVariable String us
             responses = {
                     @ApiResponse(responseCode = "200", description = "Set of favorite challengeIds by user"),
                     @ApiResponse(responseCode = "404", description = "User not found"),
-                    @ApiResponse(responseCode = "400", description = "Invalid UUID format"),
+                    @ApiResponse(responseCode = "400", description = "The provided IDs are not valid."),
```
AlexandraBonetCanela, line 360: Not sure this change should be in this PR

itachallenge-user/src/main/java/com/itachallenge/user/exception/BadRequestException.java
```diff
@@ -1,6 +1,6 @@
 package com.itachallenge.user.exception;
 
-public class BadRequestException extends Exception {
+public class BadRequestException extends RuntimeException {
```
AlexandraBonetCanela, line 3: Should this be in another PR?

AlexandraBonetCanela (commented):
The code looks good to me. I just have a couple of questions regarding the scope of this PR, but the rest is pretty clear. Nice one!

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
paumaneja: You're correct that it's not strictly part of the controller refactor. I made the update to unify the wording of all 400 errors across endpoints (we had both "Invalid UUID format" and "The provided IDs are not valid." being used)

itachallenge-user/src/main/java/com/itachallenge/user/exception/BadRequestException.java
paumaneja: As for changing BadRequestException to extend RuntimeException – this change was necessary for compatibility with Mono.error(...) in WebFlux. Checked exceptions (Exception) can't be propagated through reactive chains unless wrapped, so converting it to an unchecked exception (RuntimeException) ensures it works properly with our UserGlobalExceptionHandler.

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
```diff
@@ -91,27 +82,13 @@ public String test() {
     @GetMapping("/users/{githubUsername}")
     public Mono<ResponseEntity<UserDocument>> getUser(@PathVariable @ValidGithubUsername String githubUsername) {
         return userService.getUser(githubUsername)
+                .switchIfEmpty(Mono.error(new NotFoundException("User not found")))
```
ivilarop: You could move the logic from Mono.error(new NotFoundException(...)) that you still use (like in getUser(...)) to the service, so that the controller is even cleaner and focused only on the response.

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
```diff
@@ -4,7 +4,6 @@
 import com.itachallenge.user.document.UserDocument;
 import com.itachallenge.user.dto.UserSolutionRequestDto;
 import com.itachallenge.user.dto.UserSolutionResponseDto;
-import com.itachallenge.user.exception.BadUUIDException;
 import com.itachallenge.user.exception.NotFoundException;
```
ivilarop: With the change to the service this import should be removed

ivilarop (changes requested):

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
paumaneja: I actually moved the switchIfEmpty(Mono.error(...)) logic into the service layer already — both in getUser(...) and getUserFavorites(...).

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
paumaneja: Deleted

ivilarop (approved):

PR #839, opened 2025-04-21
Delete of unnecesary Autowired - AuthService

RustyGearBox:
I deleted an unnecessary `@Autowired` annotation from the construtor.


ivilarop (approved):

AlexandraBonetCanela (approved):
It looks OK to me.

PR #841, opened 2025-04-21
Delete of unnecesary Autowired - DocumentController

RustyGearBox:
When using constructor injection, Spring automatically injects the dependencies without requiring the @Autowired annotation. So We can safely remove it from the constructor.

### DIFF
--- a/itachallenge-document/src/main/java/com/itachallenge/document/controller/DocumentController.java
+++ b/itachallenge-document/src/main/java/com/itachallenge/document/controller/DocumentController.java
@@ -35,7 +35,6 @@ public class DocumentController {
     @Value("${spring.application.name}")
     private String appName;
 
-    @Autowired
     public DocumentController(OpenApiConfig openApiConfig, DocumentService documentService) {
         this.openApiConfig = openApiConfig;
         this.documentService = documentService;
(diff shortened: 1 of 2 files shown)


ivilarop (approved):

paumaneja:
I noticed that env is still being injected using @Autowired, while the other dependencies (openApiConfig and documentService) are injected via constructor.
Since the goal of the PR is to remove @Autowired, I was wondering if there's any particular reason why env was kept this way — maybe something I'm missing?

Also, would using @RequiredArgsConstructor with all three fields marked as final be a viable alternative here?

paumaneja (commented):
I noticed that env (lines 29-30) is still being injected using @Autowired, while the other dependencies (openApiConfig and documentService) are injected via constructor.
Since the goal of the PR is to remove @Autowired, I was wondering if there's any particular reason why env was kept this way — maybe something I'm missing?

Also, would using @RequiredArgsConstructor with all three fields marked as final be a viable alternative here?

itachallenge-document/src/main/java/com/itachallenge/document/controller/DocumentController.java
```diff
@@ -25,20 +22,12 @@ public class DocumentController {
 
     private final OpenApiConfig openApiConfig;
     private final DocumentService documentService;
-
-    @Autowired
     private Environment env;
 
-    @Value("${spring.application.version}")
-    private String version;
```
ivilarop, line 33: why is it deleted?

itachallenge-document/src/main/java/com/itachallenge/document/controller/DocumentController.java
```diff
@@ -25,20 +22,12 @@ public class DocumentController {
...
-
-    @Value("${spring.application.name}")
-    private String appName;
```
ivilarop, line 36: why is it deleted?

ivilarop (changes requested):

ivilarop:
We're closing this PR and will address it in future sprints.

PR #842, opened 2025-04-24
Feature/271_Un_usuario_puede_guardar_única_solución_en_draft

Almami679:
Se ha actualizado el SolutionDocument y la lógica del SolutionService para usar un objeto de SolutionDocument en lugar de List<SolutionDocument>

Los archivos modificados son:
```
itachallenge-user/src/main/java/com/itachallenge/user/document/UserSolutionDocument.java
itachallenge-user/src/main/java/com/itachallenge/user/service/UserSolutionServiceImpl.java
'''

### DIFF
--- a/itachallenge-user/src/main/java/com/itachallenge/user/document/UserSolutionDocument.java
+++ b/itachallenge-user/src/main/java/com/itachallenge/user/document/UserSolutionDocument.java
@@ -33,7 +33,7 @@ public class UserSolutionDocument {
     private ChallengeStatus status;
 
     @Field("solution")
-    private List<SolutionAttemptDocument> solutionAttemptDocument;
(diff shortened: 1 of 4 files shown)


Sh1ng0 (approved):
After reviewing this PR many times I honestly see no way of improving this simple refactor, a list for an object, it doesn't need to go on the changelog either, it is approved from my side...
By the way the SolutionAttemptDocument class is a new one, added by Gwen a day ago, in case of any confusion.

ivilarop:
The PR name is incorrect. It does not follow the best practices guide.

ivilarop:
Have you updated/regularized the database so that only one solution is considered?

Almami679:
PR name changed, document now manages a document instead of a list, mongo will save them as it should

ivilarop (approved):

PR #843, opened 2025-04-25
Feature/205: Add delete bookmark endpoint challenge microservice

AlexandraBonetCanela:
This PR centers on the second endpoint needed in order to successfully delete a challenge from bookmarks _(delete bookmark endpoint from User Microservice is already in development)._

The functionality is at follows:

- Checks if the Challenge exists
- Sends a request to the deteleBookmark endpoint in UserMicroservice to delete the challenge from the user's bookmarks
- Decreases the counter of the total_bookmarked of said challenge.
- Returns a BookmarkDTO (that contain if user bookmarked the challenge and the updated total times the challenge has been bookmarked)

**UPDATED FILES:**

- CHANGELOG
- Wiki endpoints page

[Taiga ticket:](https://tree.taiga.io/project/luisvi-ita-challenges/task/254)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html). 
-
+* Issue #843: Added DELETE endpoint in the Challenge Microservice to unbookmark a challenge.
 * Issue #829: Added bookmark POST endpoint in the Challenge Microservice to bookmark a challenge.
 * Issue #827: Added bookmark POST endpoint in the User Microservice to bookmark a challenge.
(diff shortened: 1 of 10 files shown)


paumaneja:
[Technical Debt Proposal / Future Improvement]

Since we knew that favorites and bookmarks functionalities would be very similar, it would have been ideal to design a more generic and reusable solution from the start. This would have helped us minimize duplication and made future developments more scalable.

Looking ahead, we could consider:

- Define a shared structure to represent user-challenge relations generically.

- Create a common service layer to manage the shared business logic.

- Reuse request and response models (DTOs) where possible.

- Unify and simplify the API patterns and request handling.

- Parameterize and consolidate the related tests to avoid duplicated test code.

This is not blocking for now, but I recommend we document it as technical debt to be addressed in future sprints if priorities allow. It would bring long-term benefits to maintainability and code quality.

ivilarop (approved):

PR #844, opened 2025-04-27
feature/256 solvedCounter: Creation and modification of supporting files

RustyGearBox:
The next support files were created or modified to continue with the implemetacion of the funcionality of the solvedCounter.

1. Update of **ChallengeDocument**:
  · New field  `int timesSolved` 
  · New method `increaseTimesSolved()`

2. Update of **ChallengeDto**:
  · New field `int timesSolved`

3. Creation of **SolvedDto**
  · New fields `Boolean isSolved` and `int timesSolved`

4. Update of **UserChallengeActionType (enum)**
  · New value `SOLVED`

5. Update of **DocumentToDtoConverter**
  · Added `getTimesSolved` to the method `convertDocumentToDto()`

There are tests that have been corrected due to file modifications.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,6 @@
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
-* Issue #844: Added value SOLVED to enum UserChallengeActionType
 * Issue #843: Added DELETE endpoint in the Challenge Microservice to unbookmark a challenge.
 * Issue #829: Added bookmark POST endpoint in the Challenge Microservice to bookmark a challenge.
 * Issue #827: Added bookmark POST endpoint in the User Microservice to bookmark a challenge.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 ## CHANGELOG
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
-and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html). 
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+* Issue #844: Added value SOLVED to enum UserChallengeActionType
 * Issue #843: Added DELETE endpoint in the Challenge Microservice to unbookmark a challenge.
 * Issue #829: Added bookmark POST endpoint in the Challenge Microservice to bookmark a challenge.
 * Issue #827: Added bookmark POST endpoint in the User Microservice to bookmark a challenge.
--- a/itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/SolvedDto.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/SolvedDto.java
@@ -1,0 +1,14 @@
+package com.itachallenge.challenge.dto;
+
+import lombok.*;
+
+@AllArgsConstructor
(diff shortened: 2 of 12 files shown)


itachallenge-challenge/src/main/java/com/itachallenge/challenge/enums/UserChallengeActionType.java
```diff
@@ -2,5 +2,6 @@
 
 public enum UserChallengeActionType {
     BOOKMARKS,
-    FAVORITES
+    FAVORITES,
+    SOLVED
```
Almami679, line 6: This change should be added to the changelog file

itachallenge-challenge/src/test/java/com/itachallenge/challenge/document/ChallengeTest.java
```diff
@@ -172,13 +198,44 @@ void getTagsTest() {
...
                 List.of(tag.getIdTag())
         );
 
         assertTrue(challenge.getTags().contains(tag.getIdTag()));
         assertEquals(1, challenge.getTags().size());
     }
 
+    @Test
+    void increaseTimesSolved_whenTimesSolvedIsNull_shouldSetToOne() {
+        // Given
```
Almami679: delete all unnecessary comments

Almami679 (approved):
Remember to update the changelog

itachallenge-challenge/src/main/java/com/itachallenge/challenge/enums/UserChallengeActionType.java
RustyGearBox: I have added this change to the CHANGELOG

itachallenge-challenge/src/test/java/com/itachallenge/challenge/document/ChallengeTest.java
RustyGearBox: I have deleted the comments

itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/SolvedDto.java
ivilarop, line 1: This flow is triggered only when the user clicks "Finish," which means the challenge has been solved.
The existing endpoint for uploading the solution already saves the status as ENDED, so it's not necessary to include an isSolved field.
A SolvedDto is also not required, since we're not modeling a new entity; we just want to update a counter.
Following the same pattern as timesFavorite or timesBookmark, the most logical approach is to add timesSolved directly to the ChallengeDto and the document.
At the flow level, the frontend only calls the user service, which saves the solution and then internally calls the challenge service to increment the counter.

CHANGELOG.md
ivilarop, line 1: The change in CHANGELOG.md would be best added to the latest PR where the endpoint that updates timesSolved is implemented, as this doesn't yet expose actual functionality.
This will make the changelog clearer, grouping the entire action (endpoint creation + counter update) into a single entry.

ivilarop (changes requested):

itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/SolvedDto.java
RustyGearBox: The SolvedDto is needed so that, in the future, the FrontEnd team can display which challenges have been marked as solved, similar to how Bookmarks and Favorites are handled.

CHANGELOG.md
RustyGearBox: Ok, I have deleted the issue 844, it will be added at the last PR.

ivilarop (approved):

PR #849, opened 2025-04-28
Feature/255 get user bookmarks user microservice

AlexandraBonetCanela:
This feature mimics the get user favorites and its aim is to retrieve all the challenges bookmarked by a single user.

**FILES UPDATED:**

- CHANGELOG
- Wiki endpoints

[Taiga Ticket here](https://tree.taiga.io/project/luisvi-ita-challenges/task/255)

### DIFF
--- a/itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
+++ b/itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
@@ -370,4 +370,32 @@ public class UserController {
                     return ResponseEntity.ok().body(favorites);
                 });
     }
+
+    @Operation(
+            summary = "Gets challenges marked as bookmarks by a user",
+            description = "Returns a set of challenge IDs that the specified user has marked as bookmarked",
+            parameters = {
+                    @Parameter(
+                            name = "userId",
+                            description = "UUID of the user",
+                            required = true,
+                            in = ParameterIn.PATH
+                    )
+            },
+            responses = {
+                    @ApiResponse(responseCode = "200", description = "Set of bookmarked challengeIds by user"),
+                    @ApiResponse(responseCode = "404", description = "User not found"),
+                    @ApiResponse(responseCode = "400", description = "The provided IDs are not valid."),
+                    @ApiResponse(responseCode = "500", description = "Unexpected error")
+            }
+    )
+
+    @GetMapping("/users/{userId}/bookmarks")
+    public Mono<ResponseEntity<Set<UUID>>> getUserBookmarks(@PathVariable String userId) {
+        return userService.getUserBookmarks(userId)
+                .map(bookmarks -> {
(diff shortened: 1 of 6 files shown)


itachallenge-user/src/test/java/com/itachallenge/user/service/UserServiceImplTest.java
```diff
@@ -727,4 +727,69 @@ void getUserFavorites_WhenInvalidUUID_ReturnsBadUUIDException() {
...
+        String invalidUUID = "invalid-uuid";
+
+        userService.getUserBookmarks(invalidUUID)
+                .as(StepVerifier::create)
+                .expectErrorMatches(error ->
+                        error instanceof BadUUIDException &&
+                                error.getMessage().equals("Invalid ID format"))
+                .verify();
+    }
+
```
g-lemoing, line 794: Suggestion: Have you considered to declare and share variables at the beginning of the class and create a setup() method to lighten the global weight of this class?

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
```diff
@@ -370,4 +370,32 @@ public Mono<ResponseEntity<Set<UUID>>> getUserFavorites(@PathVariable String use
...
+    )
+
+    @GetMapping("/users/{userId}/bookmarks")
+    public Mono<ResponseEntity<Set<UUID>>> getUserBookmarks(@PathVariable String userId) {
+        return userService.getUserBookmarks(userId)
+                .map(bookmarks -> {
+                    log.info("Retrieved {} bookmarked challenges for user {}", bookmarks.size(), userId);
+                    return ResponseEntity.ok().body(bookmarks);
+                });
+    }
```
g-lemoing, line 400: You could also put your log.info in the service instead.

g-lemoing (commented):
Good job! I've added a couple of suggestions to your code.
In your taiga task, you state the purpose of this endpoint but it would be nice to put also the steps of the changes you implement (although in that case, it's quite simple).

itachallenge-user/src/test/java/com/itachallenge/user/service/UserServiceImplTest.java
AlexandraBonetCanela: I'm gonna leave it for now because these tests need a refactoring already but very good point!

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
AlexandraBonetCanela: The convention is to put the log in the controller in this project so I will leave it as it is for now.

g-lemoing (approved):

ivilarop (approved):

PR #850, opened 2025-04-28
feature/282 Servicio y repositorio para obtener recursos por challengeId

Sh1ng0:
Se implementa la lógica de servicio y repositorio para recuperar los recursos asociados a un challengeId.
No se incluye el controlador (irá en una PR posterior).

+Añadido el método getResourcesByChallengeId(UUID challengeId) en el servicio.

+Ampliado el repositorio con findByChallengeIdsContaining(UUID challengeId).

+Añadido test para validar que el repositorio filtra correctamente los recursos.

+Manejo básico de errores y validación de entrada (challengeId != null).

### DIFF
--- a/itachallenge-challenge/src/main/java/com/itachallenge/challenge/service/ResourceServiceImpl.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/challenge/service/ResourceServiceImpl.java
@@ -11,6 +11,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.cache.annotation.CacheEvict;
 import org.springframework.stereotype.Service;
+import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
 import java.util.*;
@@ -134,5 +135,21 @@ public class ResourceServiceImpl implements IResourceService {
                     return savedDto;
                 })
                 .doOnError(error -> log.error("Error occurred when creating resource {}", error.getMessage()));
+    }
+
+
+    public Flux<ResourceDto> getResourcesByChallengeId(UUID challengeId) {
+
+        if (challengeId == null) {
+            return Flux.error(new IllegalArgumentException("Challenge ID cannot be null"));
+        }
+
+
+        return resourceRepository.findByChallengeIdsContaining(challengeId)
+                .map(resourceDoc -> resourceConverter.convertDocumentToDto(resourceDoc, ResourceDto.class))
+                .onErrorResume(error -> {
+                    log.error("Error fetching resources: {}", error.getMessage());
+                    return Flux.error(new RuntimeException("Server error while fetching resources"));
+                });
     }
 }
--- a/itachallenge-challenge/src/main/java/com/itachallenge/challenge/service/ResourceServiceImpl.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/challenge/service/ResourceServiceImpl.java
@@ -5,6 +5,9 @@ import com.itachallenge.challenge.dto.ChallengeListDto;
 import com.itachallenge.challenge.dto.ResourceDto;
 import com.itachallenge.challenge.enums.AssociationType;
 import com.itachallenge.challenge.enums.Topic;
+import com.itachallenge.challenge.exception.BadRequestException;
+import com.itachallenge.challenge.exception.InternalServerErrorException;
+import com.itachallenge.challenge.exception.ResourceNotFoundException;
 import com.itachallenge.challenge.helper.DocumentToDtoConverter;
 import com.itachallenge.challenge.repository.ResourceRepository;
 import org.slf4j.Logger;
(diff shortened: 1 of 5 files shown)


itachallenge-challenge/src/main/java/com/itachallenge/challenge/service/ResourceServiceImpl.java
```diff
@@ -135,4 +136,20 @@ private Mono<ResourceDto> saveResource(ResourceDto resourceDto) {
                 })
                 .doOnError(error -> log.error("Error occurred when creating resource {}", error.getMessage()));
     }
+
+
+    public Flux<ResourceDto> getResourcesByChallengeId(UUID challengeId) {
+
+        if (challengeId == null) {
```
AlexandraBonetCanela: The Exception should be handled by the global handler. Maybe you can leave it like this and add a task to the backlog to tackle this in another PR later?

itachallenge-challenge/src/main/java/com/itachallenge/challenge/service/ResourceServiceImpl.java
```diff
@@ -135,4 +136,20 @@ private Mono<ResourceDto> saveResource(ResourceDto resourceDto) {
...
+            return Flux.error(new IllegalArgumentException("Challenge ID cannot be null"));
+        }
+
+
+        return resourceRepository.findByChallengeIdsContaining(challengeId)
+                .map(resourceDoc -> resourceConverter.convertDocumentToDto(resourceDoc, ResourceDto.class))
+                .onErrorResume(error -> {
+                    log.error("Error fetching resources: {}", error.getMessage());
```
Almami679: To keep better track of errors, I would create some custom exceptions

Almami679 (changes requested):
Remember to update the changelog file

itachallenge-challenge/src/main/java/com/itachallenge/challenge/service/ResourceServiceImpl.java
AlexandraBonetCanela: Nice one, even though I would add the ChallengeId in the log  something like "Error fetching resources for Challenge {}, for better error tracking

itachallenge-challenge/src/main/java/com/itachallenge/challenge/service/ResourceServiceImpl.java
```diff
@@ -135,4 +136,20 @@ private Mono<ResourceDto> saveResource(ResourceDto resourceDto) {
...
+                    return Flux.error(new RuntimeException("Server error while fetching resources"));
```
AlexandraBonetCanela: Same as before with the Global handler Exception

itachallenge-challenge/src/test/java/com/itachallenge/challenge/repository/ResourceRepositoryTest.java
```diff
@@ -251,4 +251,50 @@ void deleteAllResourcesTest() {
                 .expectNext(0L)
                 .verifyComplete();
     }
+
+    // Revisar + limpiar esto
```
AlexandraBonetCanela: Are the comments still needed?

You can use the //TODO to explicit what is left but not sure if Ivan is keen on doing that

AlexandraBonetCanela: Probably comments should be written in English as well

itachallenge-challenge/src/main/java/com/itachallenge/challenge/service/ResourceServiceImpl.java
Sh1ng0: I will in the next PR, once the endpoint is fully implemented, also the changelog will be updated in the next PR (Once the endpoint is fully created)

itachallenge-challenge/src/main/java/com/itachallenge/challenge/service/ResourceServiceImpl.java
Sh1ng0: I didn't want the PR to be to long because of time constraints, if I have time I will add the personalized exceptions and extend the Global Exception.

itachallenge-challenge/src/main/java/com/itachallenge/challenge/service/ResourceServiceImpl.java
Sh1ng0: Same as above, I will also add it to the backlgo in Taiga

itachallenge-challenge/src/test/java/com/itachallenge/challenge/repository/ResourceRepositoryTest.java
Sh1ng0: Comments are deleted in the  current commit, thanks for pointing it out

AlexandraBonetCanela (approved):

Almami679 (approved):

itachallenge-challenge/src/main/java/com/itachallenge/challenge/service/ResourceServiceImpl.java
ivilarop: I suggest replacing this if (challengeId == null) with a more reactive validation, using Mono.justOrEmpty(...) and switchIfEmpty(...).

In reactive programming, the idea is that everything—including validations—is done within the flow (Mono or Flux). By placing the if statement outside the flow, we're breaking that flow and returning to a more imperative approach (like in traditional Java).

With this change:

-The code will be more consistent with the Spring WebFlux style.
-We gain readability and ease of testing or chaining logic.
-We can keep all the behavior within the same reactive flow.

itachallenge-challenge/src/main/java/com/itachallenge/challenge/service/ResourceServiceImpl.java
ivilarop: Have you added it as technical debt or have you already made the change in this PR? If the change is pending, try to implement it so as not to take on more technical debt than necessary, since it's a small change.

itachallenge-challenge/src/main/java/com/itachallenge/challenge/service/ResourceServiceImpl.java
ivilarop: To avoid further increasing our technical debt, try making the change. Can't you reuse any existing exceptions? The list of custom exceptions is quite extensive, and we don't want to keep adding to it, so if you can reuse one, it's a better option.

ivilarop (changes requested):

itachallenge-challenge/src/main/java/com/itachallenge/challenge/service/ResourceServiceImpl.java
Sh1ng0: Custom exeptions from the project added to the method in the service layer

Sh1ng0: Method changed to follow a proper reactive pipeline.

ivilarop (approved):

PR #853, opened 2025-04-29
feature/270 creacion nuevo challengeStatus en Solution con valor INPROGRESS

Almami679:
Añadido en el enum de challengeStatus el valor de INPROGRESS

### DIFF
--- a/itachallenge-user/src/main/java/com/itachallenge/user/document/enums/ChallengeStatus.java
+++ b/itachallenge-user/src/main/java/com/itachallenge/user/document/enums/ChallengeStatus.java
@@ -13,8 +13,13 @@ public enum ChallengeStatus {ENDED("ENDED"),
     }
 
     public static ChallengeStatus determineChallengeStatus(String status){
-        return status != null  && status.equalsIgnoreCase(ChallengeStatus.ENDED.getValue()) ?
-                ChallengeStatus.ENDED : null;
+        ChallengeStatus output = null;
+        if(status != null  && status.equalsIgnoreCase(ChallengeStatus.ENDED.getValue())){
+            output = ChallengeStatus.ENDED;
+        } else if (status != null && status.equalsIgnoreCase(ChallengeStatus.IN_PROGRESS.getValue())) {
+            output = ChallengeStatus.IN_PROGRESS;
+        }
+        return output;
     }
 
 }
--- a/itachallenge-user/src/main/java/com/itachallenge/user/document/enums/ChallengeStatus.java
+++ b/itachallenge-user/src/main/java/com/itachallenge/user/document/enums/ChallengeStatus.java
@@ -2,6 +2,8 @@ package com.itachallenge.user.document.enums;
 
 import lombok.Getter;
 
+import java.util.Arrays;
+
 @Getter
 public enum ChallengeStatus {ENDED("ENDED"),
     IN_PROGRESS("IN_PROGRESS");
@@ -12,12 +14,13 @@ public enum ChallengeStatus {ENDED("ENDED"),
         this.value = value;
     }
 
-    public static ChallengeStatus determineChallengeStatus(String status){
+    public static ChallengeStatus challengeStatusFromString(String status) {
         ChallengeStatus output = null;
-        if(status != null  && status.equalsIgnoreCase(ChallengeStatus.ENDED.getValue())){
-            output = ChallengeStatus.ENDED;
-        } else if (status != null && status.equalsIgnoreCase(ChallengeStatus.IN_PROGRESS.getValue())) {
-            output = ChallengeStatus.IN_PROGRESS;
+        if (status != null) {
+            output = Arrays.stream(ChallengeStatus.values())
+                    .filter(s -> status.equalsIgnoreCase(s.getValue()))
+                    .findFirst()
+                    .orElse(null);
         }
         return output;
     }
--- a/itachallenge-user/src/main/java/com/itachallenge/user/document/enums/ChallengeStatus.java
+++ b/itachallenge-user/src/main/java/com/itachallenge/user/document/enums/ChallengeStatus.java
@@ -3,7 +3,8 @@ package com.itachallenge.user.document.enums;
 import lombok.Getter;
 
 @Getter
-public enum ChallengeStatus {ENDED("ENDED");
+public enum ChallengeStatus {ENDED("ENDED"),
+    IN_PROGRESS("IN_PROGRESS");
 
     private final String value;
 
--- a/itachallenge-user/src/main/java/com/itachallenge/user/service/UserSolutionServiceImpl.java
+++ b/itachallenge-user/src/main/java/com/itachallenge/user/service/UserSolutionServiceImpl.java
@@ -11,7 +11,6 @@ import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Service;
 import reactor.core.publisher.Mono;
 
-import java.util.List;
 import java.util.UUID;
 
 @Service
@@ -37,7 +36,7 @@ public class UserSolutionServiceImpl implements IUserSolutionService {
                 .solutionText(userSolutionDto.getSolutionText())
                 .build();
 
-        challengeStatus = ChallengeStatus.determineChallengeStatus(status);
+        challengeStatus = ChallengeStatus.challengeStatusFromString(status);
 
         if (challengeStatus == null) {
             log.error("PUT operation failed due to invalid challenge status parameter");


AlexandraBonetCanela (approved):
Perfect!. The name of the PR better not using underscores _ but PR is short and clear and nice. Well done

itachallenge-user/src/main/java/com/itachallenge/user/document/enums/ChallengeStatus.java
```diff
@@ -13,8 +13,13 @@ public enum ChallengeStatus {ENDED("ENDED"),
     }
 
     public static ChallengeStatus determineChallengeStatus(String status){
```
AlexandraBonetCanela: I think determine is not a word commonly used in coding as it does not clearly express what the method does, resolve, parse, map or getChallengeStatusFromString may be options?

itachallenge-user/src/main/java/com/itachallenge/user/document/enums/ChallengeStatus.java
```diff
@@ -13,8 +13,13 @@ public enum ChallengeStatus {ENDED("ENDED"),
...
-        return status != null  && status.equalsIgnoreCase(ChallengeStatus.ENDED.getValue()) ?
-                ChallengeStatus.ENDED : null;
+        ChallengeStatus output = null;
+        if(status != null  && status.equalsIgnoreCase(ChallengeStatus.ENDED.getValue())){
```
AlexandraBonetCanela: No need to test status != null

itachallenge-user/src/main/java/com/itachallenge/user/document/enums/ChallengeStatus.java
```diff
@@ -13,8 +13,13 @@ public enum ChallengeStatus {ENDED("ENDED"),
...
+            output = ChallengeStatus.ENDED;
+        } else if (status != null && status.equalsIgnoreCase(ChallengeStatus.IN_PROGRESS.getValue())) {
```
AlexandraBonetCanela: No need to test status != null

itachallenge-user/src/main/java/com/itachallenge/user/document/enums/ChallengeStatus.java
```diff
@@ -13,8 +13,13 @@ public enum ChallengeStatus {ENDED("ENDED"),
     }
 
     public static ChallengeStatus determineChallengeStatus(String status){
-        return status != null  && status.equalsIgnoreCase(ChallengeStatus.ENDED.getValue()) ?
-                ChallengeStatus.ENDED : null;
+        ChallengeStatus output = null;
+        if(status != null  && status.equalsIgnoreCase(ChallengeStatus.ENDED.getValue())){
+            output = ChallengeStatus.ENDED;
```
AlexandraBonetCanela: You can make a return here inside

itachallenge-user/src/main/java/com/itachallenge/user/document/enums/ChallengeStatus.java
```diff
@@ -13,8 +13,13 @@ public enum ChallengeStatus {ENDED("ENDED"),
...
+        } else if (status != null && status.equalsIgnoreCase(ChallengeStatus.IN_PROGRESS.getValue())) {
+            output = ChallengeStatus.IN_PROGRESS;
```
AlexandraBonetCanela: Same as above, you can meka  return here

AlexandraBonetCanela (changes requested):

itachallenge-user/src/main/java/com/itachallenge/user/document/enums/ChallengeStatus.java
Almami679: The method was already created by Gwen, I just expanded it.

itachallenge-user/src/main/java/com/itachallenge/user/document/enums/ChallengeStatus.java
Almami679: In the service, the value is managed when it is null to throw an exception, that's why I continued controlling it like this

itachallenge-user/src/main/java/com/itachallenge/user/document/enums/ChallengeStatus.java
Almami679: In the service, the value is managed when it is null to throw an exception, that's why I continued controlling it like this

itachallenge-user/src/main/java/com/itachallenge/user/document/enums/ChallengeStatus.java
Almami679: According to good practices, a function should only have one return.

itachallenge-user/src/main/java/com/itachallenge/user/document/enums/ChallengeStatus.java
AlexandraBonetCanela: Oki

AlexandraBonetCanela (approved):

itachallenge-user/src/main/java/com/itachallenge/user/document/enums/ChallengeStatus.java
ivilarop: Avoid repeating status != null in each condition.You can make a guard clause at the beginning

itachallenge-user/src/main/java/com/itachallenge/user/document/enums/ChallengeStatus.java
ivilarop: Okay, Gwen did it, but if we detect that it can be improved, we can apply the change, right?

itachallenge-user/src/main/java/com/itachallenge/user/document/enums/ChallengeStatus.java
```diff
@@ -12,8 +13,13 @@ public enum ChallengeStatus {ENDED("ENDED");
     }
 
     public static ChallengeStatus determineChallengeStatus(String status){
-        return status != null  && status.equalsIgnoreCase(ChallengeStatus.ENDED.getValue()) ?
-                ChallengeStatus.ENDED : null;
+        ChallengeStatus output = null;
+        if(status != null  && status.equalsIgnoreCase(ChallengeStatus.ENDED.getValue())){
```
ivilarop: Could we simplify this by replacing the if/else with a more functional approach, such as using a Map<String, ChallengeStatus> or a stream-based solution? It might improve readability and scalability if more statuses are added in the future.

ivilarop (changes requested):

itachallenge-user/src/main/java/com/itachallenge/user/document/enums/ChallengeStatus.java
Almami679: Yes, precisely I always want to change things that I see wrong, but I focused on a few changes, I'll update it right now

itachallenge-user/src/main/java/com/itachallenge/user/document/enums/ChallengeStatus.java
Almami679: I see the map more for enums with many values, and I think that a solution will not have more than 3 or 4 states, I have used stream, so that it is more readable.

itachallenge-user/src/main/java/com/itachallenge/user/document/enums/ChallengeStatus.java
```diff
@@ -2,18 +2,27 @@
...
     private final String value;
 
     ChallengeStatus(String value) {
         this.value = value;
     }
 
-    public static ChallengeStatus determineChallengeStatus(String status){
-        return status != null  && status.equalsIgnoreCase(ChallengeStatus.ENDED.getValue()) ?
-                ChallengeStatus.ENDED : null;
+    public static ChallengeStatus challengeStatusFromString(String status) {
```
ivilarop, line 17: Good use of streams! To make this even cleaner and safer, I’d suggest returning an Optional<ChallengeStatus> instead of using .orElse(null), since it brings us back to dealing with null.

Here's a proposed change:
public static Optional<ChallengeStatus> fromString(String status) {
    if (status == null) return Optional.empty();
    return Arrays.stream(ChallengeStatus.values())
        .filter(s -> status.equalsIgnoreCase(s.getValue()))
        .findFirst();
}
This way, the method clearly communicates that the result may be absent, and the caller can handle it using .orElse, .orElseThrow, .ifPresent, etc., without risking a NullPointerException.

ivilarop (changes requested):

itachallenge-user/src/main/java/com/itachallenge/user/document/enums/ChallengeStatus.java
Almami679: As we have discussed, it is left as null, since it is controlled in the service layer

ivilarop (approved):

PR #864, opened 2025-05-07
feature/334-Logout-endpoint-response-should-manage-expired-token

trisk910:
1. -Endpoint /logout:

- Modificado para devolver siempre un estado 200 OK con el mensaje "Logout successful", independientemente de si el token JWT es válido, inválido, expirado o ausente.
- Eliminada la validación estricta del token para mejorar la experiencia del usuario.
- Añadidos logs para registrar intentos de logout con tokens inválidos o ausentes.

2. Pruebas actualizadas en AuthControllerTest:

- Se añadieron pruebas para validar el nuevo comportamiento del endpoint /logout.
- Se eliminaron las pruebas que esperaban una respuesta 401 para el endpoint /logout

### DIFF
--- a/itachallenge-auth/src/main/java/com/itachallenge/auth/controller/AuthController.java
+++ b/itachallenge-auth/src/main/java/com/itachallenge/auth/controller/AuthController.java
@@ -145,8 +145,11 @@ public class AuthController {
 
     @PostMapping("/logout")
     public Mono<ResponseEntity<Map<String, String>>> logout(@RequestHeader(value = "Authorization", required = false) String authHeader) {
-        if (authHeader == null || !authHeader.startsWith("Bearer "))
-            log.warn("Logout attempt without token");
+        if (authHeader == null || !authHeader.startsWith("Bearer ")){
+            log.warn("Logged out without token");
+        }else{
+                log.info("Successfully logged out with token
+        }
         
         return Mono.just(ResponseEntity.ok(Map.of(MESSAGE_KEY, "Logout successful")));
     }
--- a/itachallenge-auth/src/main/java/com/itachallenge/auth/controller/AuthController.java
+++ b/itachallenge-auth/src/main/java/com/itachallenge/auth/controller/AuthController.java
@@ -29,6 +29,7 @@ public class AuthController {
     private static final String KEY_USERNAME = "username";
     public static final String X_GITHUB_USERNAME = "X-Github-Username";
     public static final String X_AUTHENTICATION_STATUS = "X-Authentication-Status";
+    private static final String MESSAGE_KEY = "message";
 
     private final IAuthService authService;
 
@@ -99,7 +100,7 @@ public class AuthController {
                 .switchIfEmpty(Mono.defer(() -> {
                     response.put(KEY_USERNAME, null);
                     response.put(KEY_IS_VALID, false);
-                    response.put("message", "User does not exist in the database");
+                    response.put(MESSAGE_KEY, "User does not exist in the database");
                     return Mono.just(ResponseEntity.status(HttpStatus.FORBIDDEN)
                             .header("X-Validation-Status", "Forbidden")
                             .header(X_GITHUB_USERNAME, githubUsername)
@@ -119,7 +120,7 @@ public class AuthController {
                     }
                     response.put(KEY_USERNAME, null);
                     response.put(KEY_IS_VALID, false);
-                    response.put("message", message);
+                    response.put(MESSAGE_KEY, message);
                     return Mono.just(ResponseEntity.status(status)
                             .header("X-Validation-Status", "Forbidden")
                             .header(X_GITHUB_USERNAME, githubUsername)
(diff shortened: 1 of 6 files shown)


shernandez334:
Everything looks solid, and I don’t see a need for major changes. One small improvement I’d recommend is to hash the token before logging it during logout. This way, even if someone were to intercept or access the logs, they wouldn’t be able to view or misuse the actual token. A good place to apply the hash would be here:

log.info("Logout attempt for token: {}", jwt);

shernandez334 (approved):

ivilarop:
@trisk910 The name of the PR does not follow the defined nomenclature

itachallenge-auth/src/main/java/com/itachallenge/auth/controller/AuthController.java
```diff
@@ -143,21 +144,14 @@ public Mono<String> callUserTest() {
...
-        String jwt = authHeader.replace("Bearer ", "");
-
-        if (jwtService.validateToken(jwt)) {
-            log.info("Logout successful for token: {}", jwt);
-            return Mono.just(ResponseEntity.ok(Map.of("message", "Logout successful")));
-        } else {
-            log.warn("Invalid or expired token during logout");
-            return Mono.just(ResponseEntity.status(HttpStatus.UNAUTHORIZED)
-                    .body(Map.of("message", "Invalid or expired token")));
+    public Mono<ResponseEntity<Map<String, String>>> logout(@RequestHeader(value = "Authorization", required = false) String authHeader) {
```
ivilarop: Currently, two very different cases are not being distinguished:

✅ A valid or expired token → returns 200 OK, which makes sense from a UX perspective.

⚠️ A missing or malformed token → also returns 200 OK, which may hide integration issues or incorrect frontend calls.

💡 It would be advisable to consider distinguishing between them:

For missing or malformed tokens, we could return 401 Unauthorized with a message like "Authorization header is missing or malformed".

For expired tokens, we can keep returning 200 OK as currently implemented, to allow the user to log out smoothly.

📌 Regarding the implementation:

There is no need to create a new custom exception. We already have several (e.g., JwtException, BadRequestException, etc.) and could:

Reuse JwtException if it supports custom messages, or

Simply return the 401 directly from the endpoint in this specific case.

A dedicated exception (e.g., MissingTokenException) should only be considered if this behavior starts appearing in multiple endpoints and we want to centralize its handling via the GlobalExceptionHandler. For now, it seems unnecessary.

🧪 In summary:

The current behavior is correct for expired tokens.

We can improve the robustness of the endpoint by differentiating when no token is sent at all.

There is no need to add more classes to the project for this, at least for now.

ivilarop:
@trisk910 ✅ The /logout endpoint should also be added to the postman/ITA-Challenges.postman_collection.json file in order to:

1-Simplify testing: lets you confirm that logout returns 200 OK correctly with valid, expired, or missing tokens.

2-Document the expected behavior for other developers on the team.

3- Detect frontend issues, such as the token not being sent properly.

ivilarop (changes requested):

itachallenge-auth/src/main/java/com/itachallenge/auth/controller/AuthController.java
trisk910: Added header authorization reponse when missing or malformed

ivilarop: Actualmente, el endpoint /logout solo comprueba si el header Authorization existe y empieza por Bearer, pero no valida si el token tiene formato JWT ni si está caducado.
Por tanto, ahora mismo devuelve 200 OK incluso si el token está mal formado o inventado, y eso no es lo que queremos.

Lo que esperamos es que:

Se devuelva 401 Unauthorized si el token no está o está mal formado (por ejemplo, no tiene estructura JWT válida).

Se devuelva 401 Unauthorized si el token está bien formado pero tiene firma inválida (es decir, ha sido manipulado).

Se devuelva 200 OK si el token está bien formado y simplemente está caducado.

También se devuelve 200 OK si el token es totalmente válido.

trisk910: Added method Post - Logout into the postman_collection.json. Also tested and receiving the expected responses.

tonijimenez72 (approved):
Reviewed and approved.

ivilarop (changes requested):

itachallenge-auth/src/main/java/com/itachallenge/auth/controller/AuthController.java
trisk910: - Updates validateToken method to diferenciate between expired and invalid token.
- Updated endpoint logout to validate all cases for mentionate responses.
- Updated endpoint responses tests and removed unnecesary validateToken tests

ivilarop (approved):

PR #872, opened 2025-05-12
Feature/355 tag validation error handling

IsmaPeiro:
Implement error handling for tag validation
User Story: [https://tree.taiga.io/project/luisvi-ita-challenges/us/322?milestone=453813](url)

Built on top of existing existence checks: this new checks complements the previous logic that validated whether a well-formed UUID actually existed in the database

- Detecting malformed UUIDs:

Added a new @ExceptionHandler(InvalidFormatException) in GlobalExceptionHandler that:

Returns a clear, 400-Bad Request when Jackson fails to parse a UUID.

Differentiates between malformed tag UUIDs (tags field) and other UUID fields, with customized error messages.

Why in the GlobalExceptionHandler? Jackson performs UUID conversion during JSON deserialization—before any controller or service method runs—so the only place to catch and transform those parse errors into API-friendly messages is at the global exception-handling layer.

- Preventing duplicate tag IDs before database lookup

Extended TagServiceImpl.getValidatedTags(...) to short-circuit and throw a DuplicateTagUUIDException if the incoming list contains the same UUID more than once.

This avoids unnecessary database calls and ensures clients receive an immediate, descriptive error.

- Centralizing error handling for duplicate IDs

Added a @ExceptionHandler(DuplicateTagUUIDException) in GlobalExceptionHandler to map these errors to 400-Bad Request with a clear message.


tonijimenez72:
Hi Isma,

Well done!

Please review whether comments are needed in the getValidatedTags method and in the unit tests of GlobalExceptionHandler: testHandleInvalidFormat_TagsField, testHandleInvalidFormat_OtherFieldFallback.

A note should also be left regarding the messages in Spanish used in the tests of TagServiceImplTest.

IsmaPeiro:
> Hi Isma,
> 
> Well done!
> 
> Please review whether comments are needed in the getValidatedTags method and in the unit tests of GlobalExceptionHandler: testHandleInvalidFormat_TagsField, testHandleInvalidFormat_OtherFieldFallback.
> 
> A note should also be left regarding the messages in Spanish used in the tests of TagServiceImplTest.

Ooops, thanks, I forgot to delete them; I’ll fix it.

tonijimenez72 (approved):
Reviewed and approved.

itachallenge-challenge/src/main/java/com/itachallenge/challenge/service/TagServiceImpl.java
```diff
@@ -50,9 +51,20 @@ public Set<TagDocument> convertIdTagFromTagDocument(List<UUID> tagsAssigned) {
     @Override
     public Mono<Boolean> getValidatedTags(List<UUID> tagIds) {
```
ivilarop, line 52: 💬 Great work with the logic. It correctly validates both duplicates and the existence of tags in the database. However, the getValidatedTags method has become somewhat dense, and it can be difficult to understand at first glance what each block is doing.

🎯 I encourage you to consider extracting each validation (duplicates, existence) into separate methods with meaningful names. This would help improve the readability and maintainability of the code without changing its behavior.

📚 This is a good practice when a method starts to have multiple responsibilities or nested blocks, as in this case with reactive programming.

itachallenge-challenge/src/main/java/com/itachallenge/challenge/exception/DuplicateTagUUIDException.java
ivilarop, line 1: 💬 Good job defining the DuplicateTagUUIDException exception; it makes sense in the context of tag validation.
That said, I suggest you reflect on this point a little more:

🧠 Do you think this case requires a custom exception, or might a generic exception like BadRequestException or ValidationException suffice?

🎯 I encourage you to investigate:

Is there special handling for this exception in the GlobalExceptionHandler?

Does it add value to have its own class?

Could it complicate maintenance if many custom exceptions proliferate?

If you conclude that it should remain custom, great, but if you see no practical difference, maybe we can simplify by using an existing one.

itachallenge-challenge/src/main/java/com/itachallenge/challenge/service/TagServiceImpl.java
ivilarop: I have created a technical doubt task, review it and we will discuss it to see what you think: https://tree.taiga.io/project/luisvi-ita-challenges/issue/377

itachallenge-challenge/src/main/java/com/itachallenge/challenge/exception/GlobalExceptionHandler.java
```diff
@@ -104,6 +105,27 @@ public ResponseEntity<MessageDto> handleIllegalArgumentException(IllegalArgument
 
     @ExceptionHandler(InvalidFormatException.class)
     public ResponseEntity<MessageDto> handleInvalidFormat(InvalidFormatException ex) {
-        return ResponseEntity.badRequest().body(new MessageDto(ex.getMessage()));
```
ivilarop, line 102: 💬 Good work specifically handling the InvalidFormatException to return a clearer message when the malformed UUID comes from the tags field.

🧠 Although the GlobalExceptionHandler should be as simple as possible, this type of logic is justified because:

It improves error understanding from the frontend.

It's not business logic.

And it can't be caught early, because the exception is thrown by Jackson.

As an additional suggestion, you could extract the condition to a private method (isInvalidUUIDFromTags(ex)) to keep the handler clean and readable.

ivilarop (changes requested):

itachallenge-challenge/src/main/java/com/itachallenge/challenge/service/TagServiceImpl.java
IsmaPeiro: Hi @ivilarop, 
thanks for your feedback! 
As you suggested, I’ve extracted the duplicate-checking and existence-checking logic into two private helper methods. Given the technical debt this refactor creates (which I agree we’ll need to address) and the current need for getValidatedTags to return a Mono\<Boolean>, I opted for both helpers to also return Boolean. That way, if no exceptions are thrown and both return true, getValidatedTags naturally does the same. I did consider having them return Mono\<Void>, but that would’ve forced me to use thenReturn(true) in the reactive chain—a workaround that felt less clean. I believe this strikes the right balance until we resolve the underlying debt.

itachallenge-challenge/src/main/java/com/itachallenge/challenge/exception/DuplicateTagUUIDException.java
IsmaPeiro: Hi @ivilarop,

Thanks for your feedback.
I think simplifying makes perfect sense. In this specific case, because the duplicate-UUID check ultimately results in the same 400 Bad Request response as other validation errors (with no unique payload or downstream behavior), using a generic exception avoids unnecessary classes and keeps our error-handling consistent and maintainable.

Here’s a summary of the pros, cons, and my plan:

#### Advantages of a custom `DuplicateTagUUIDException`
- **Semantic clarity:** the code expressly communicates “this is a UUID-duplication error.”  
- **Extensibility:** we could add dedicated logging, metrics, or handling in the future.  
- **Expressiveness:** `throw new DuplicateTagUUIDException(...)` reads very clearly in context.

#### Disadvantages of a custom exception
- **Maintenance overhead:** another class and handler to maintain.  
- **Boilerplate growth:** too many one-off exceptions can clutter our hierarchy.  
- **Redundancy:** currently it always becomes a plain 400 Bad Request—no unique behavior or payload.

#### Advantages of a generic `BadRequestException` (or shared `ValidationException`)
- **Fewer classes/handlers:** keeps the exception hierarchy lean.  
- **Reuses existing logic:** uses our standard 400 response and DTO for validation errors.  
- **Less boilerplate:** easier for future developers to follow.

#### Disadvantages of a generic exception
- **Less specificity:** logs or metrics won’t immediately distinguish “duplicate UUID” from other bad requests.  
- **Later refactoring if needed:** if we do need special treatment, we’ll introduce a new exception at that time.

---

#### Conclusion & Next Steps

Your proposal to replace `DuplicateTagUUIDException` with the generic `BadRequestException` aligns perfectly with the current needs. I will:

1. Refactor the validation method validateNoDuplicatesUUIDTags(...) to throw `BadRequestException`.  
2. Delete `@ExceptionHandler(DuplicateTagUUIDException.class)` from `GlobalExceptionHandler`.  
3. Update or remove any tests that referred to the old exception.
4. Remove the `DuplicateTagUUIDException` class.

This will simplify the code while preserving clear behavior. If we ever need distinct handling for UUID-duplication, we can introduce a custom exception at that point. Let me know if you see any issues!

itachallenge-challenge/src/main/java/com/itachallenge/challenge/exception/GlobalExceptionHandler.java
IsmaPeiro: Hi @ivilarop,

I agree that pulling out the UUID-from-tags logic into its own method makes the handler much cleaner. Below is a breakdown of **why** we’re doing it, the **pros and cons** of each approach (without tables), and a short **action plan** for the refactor:

#### Why extract the logic?

* **Improved Readability:** The `@ExceptionHandler` method stays focused on deciding between a “tag-UUID” error and the default error, without getting bogged down in the details of how we detect that case.
* **Separation of Concerns:** All the JSON-parsing logic lives in a helper; the handler’s only job is to delegate and return a response.

#### Advantages of Original (in-method) approach
* Zero indirection—you see the full flow in one place.
* No extra private methods to navigate to.
#### Disadvantages of Original (in-method) approach
  * The handler body is larger and more nested, making it harder to scan quickly.
  * Harder to isolate and test just the UUID-tag-detection logic.

#### Advantages of Refactored (private helper) approach
  * Handler method becomes very concise, improving overall readability.
  * Helper method can be unit-tested in isolation for both matching and non-matching paths.
#### Disadvantages of Refactored (private helper) approach
  * Introduces a small level of indirection—you need to jump to the helper to see the detection details.
  * Slightly more boilerplate (an extra method).

---

#### Conclusion

Extracting the tag-UUID detection into its own method aligns with the single-responsibility principle, keeps our exception handler focused, and makes unit testing straightforward. I’ll proceed with the refactor.

ivilarop: Is this exception automatically thrown by Jackson when a malformed UUID is sent in the tags field? 

If it's not automatically triggered, this code wouldn't be necessary, and we could eliminate both the handler and the test.

If it is triggered, then another question arises: does it make sense for the message to focus solely on the tags field?

In that case, we could:

Leave the handler, but make the message more generic (valid for any UUID field)

Or move the UUID format validation directly into the service logic and throw a BadRequestException, as you already do with duplicates.

👉 The important thing is to keep the GlobalExceptionHandler as generic and reusable as possible, and to ensure the code doesn't have "dead" exceptions or unnecessarily coupled cases.

ivilarop: We have reviewed the automatic validation of the UUID list and confirmed that error handling is currently limited to the tags field.

Due to timing issues and given that the validation is working correctly, we are leaving this behavior as is for now.

However, in light of technical debt, we want to investigate how to refactor this logic so that it doesn't depend on a specific field (tags) and can be reused more generically for similar errors in other fields with UUIDs.

This will allow us to have a more robust @ControllerAdvice that is decoupled from specific cases, improving the project's maintainability and future extensibility.

@IsmaPeiro When you have a definitive solution for this refactor, create the technical debt task and record it here for historical reference.

ivilarop (approved):

itachallenge-challenge/src/main/java/com/itachallenge/challenge/exception/GlobalExceptionHandler.java
IsmaPeiro: I have created a technical doubt task, let me know if I’ve phrased it correctly or if I need to make any changes.:
https://tree.taiga.io/project/luisvi-ita-challenges/issue/387

PR #873, opened 2025-05-12
Feature/349 Creation method connect micro.user with micro.challenge

RustyGearBox:
This PR contains the necessary changes to enable communication between the user microservice and the challenge microservice, with the goal of incrementing the field `timesSolved`

Changes:

- **NEW** `ChallengeServiceImpl.java` : This class is responsible for sending a request to an endpoint of the challenge micro, so that the challenge micro handles only the part it is responsible. In this case, incrementing the value of the field `timesSolved`

- **NEW** `IChallengeService.java` : This is the interface for `ChallengeServiceImpl.java`

- **NEW** `ChallengeNotFoundException.java` and `InternalServerErrorException.java` 

- **UPDATED** `UserGlobalHandlerException.java` : Added the previously mentioned exceptions

- **UPDATED** `gradle.build` : Added the library okhttp3 for the testing for `ChallengeServiceImpl.java`

- **NEW** `ChallengeServiceImplTest.java` : This is the test of the `ChallengeServiceImpl.java`, checks for the exceptions

NOTE : The okhttp3 library is already used in the challenge microservice.

### DIFF
--- a/itachallenge-user/build.gradle
+++ b/itachallenge-user/build.gradle
@@ -54,6 +54,7 @@ dependencies {
     testImplementation group: 'io.projectreactor', name: 'reactor-test', version: '3.1.0.RELEASE'
     testImplementation group: 'org.testcontainers', name: 'mongodb', version: '1.17.6'
     testImplementation group: 'org.testcontainers', name: 'junit-jupiter', version: '1.17.6'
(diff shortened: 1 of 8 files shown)


tonijimenez72 (approved):
Reviewed and approved.

PR #882, opened 2025-05-21
feat(auth): allow user role change at runtime (#404)

tonijimenez72:
## Allow user role change at runtime
User Story: #388 [https://tree.taiga.io/project/luisvi-ita-challenges/us/388](url)

### Summary

This update introduces a new feature that allows authenticated users to temporarily switch their role (ADMIN or USER) and receive a new JWT reflecting the new role. All logic is isolated to the auth microservice.

### Changes
SwitchRoleRequest: New User role request DTO.

UserRole: New user role enum (USER, ADMIN). 
-  Validates that the new role:
   - Is not null or blank.
   - Is different from the current role.
   - Is a valid transition between ADMIN and USER.

Since the roles are fixed and limited, using an enum improves cohesion and avoids unnecessary abstraction. Delegating this logic to an external service is not considered necessary for now.
 
AuthController:
- Added POST /switch-role endpoint.
- Tokens generated from a role switch include the claim isTemporaryRole: true.
- Extracted method extractBearerToken(...) improves token extraction clarity.

IJwtService:
- switchRole(String token, String requestedRole).
  - Extracts claims from the token.
 - Returns a new token or throws an appropriate exception based on token state.

- extractAllClaims(String token) 
  - Parses and validates the JWT using the secret signing key.
  - Verifies the token’s signature and integrity.
  - Extracts and returns the payload (claims) if the token is valid.
  - Throws exceptions if the token is:
    - Expired (ExpiredJwtException).
    - Modified or invalid (JwtException).

### Unit Tests Added
UserRoleTest:
- Validates successful role transitions between ADMIN and USER.
- Rejects same-role transitions (e.g., USER to USER)
- Rejects null, blank, or unknown role values.

JwtServiceTest:
- Role switch success case.
- Rejects invalid role switch.

AuthControllerTest:
- Successful switch returns 200.
- Expired token returns 200.
- Bad role requests returns 400.
- Invalid token, malformed or missing headers returns 401.

GlobalExceptionHandlerTest:
- Validates that InvalidRoleChangeRequestException is handled with a 400 response.

### DIFF
--- a/itachallenge-auth/src/main/java/com/itachallenge/auth/controller/AuthController.java
+++ b/itachallenge-auth/src/main/java/com/itachallenge/auth/controller/AuthController.java
@@ -1,6 +1,5 @@
 package com.itachallenge.auth.controller;
 
-
 import com.itachallenge.auth.dto.SwitchRoleRequest;
 import com.itachallenge.auth.exception.CustomBadRequestException;
 import com.itachallenge.auth.exception.CustomInternalServerErrorException;
@@ -18,9 +17,6 @@ import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.validation.annotation.Validated;
 import org.springframework.web.bind.annotation.*;
-import io.jsonwebtoken.Jwts;
-import io.jsonwebtoken.ExpiredJwtException;
-import io.jsonwebtoken.JwtException;
 
 import reactor.core.publisher.Mono;
 
@@ -39,7 +35,6 @@ public class AuthController {
     public static final String X_AUTHENTICATION_STATUS = "X-Authentication-Status";
     private static final String MESSAGE_KEY = "message";
     private static final String LOGOUT_SUCCESS = "Logout successful";
-    private static final String BEARER_KEY = "Bearer ";
 
     private final IAuthService authService;
 
@@ -182,18 +177,9 @@ public class AuthController {
             @RequestHeader(value = "Authorization", required = false) String authHeader,
             @RequestBody SwitchRoleRequest request) {
 
-        String token = extractBearerToken(authHeader);
+        String token = jwtService.extractBearerToken(authHeader);
         String newToken = jwtService.switchRole(token, request.getNewRole());
         log.info("Switch-role successful for token");
         return Mono.just(ResponseEntity.ok(Map.of("token", newToken)));
     }
-
-
-    private String extractBearerToken(String authHeader) {
-        if (authHeader == null || !authHeader.startsWith(BEARER_KEY)) {
-            throw new JwtException("Authorization header is missing or malformed");
-        }
-        return authHeader.replace(BEARER_KEY, "").trim();
-    }
-
 }
--- a/itachallenge-auth/src/main/java/com/itachallenge/auth/controller/AuthController.java
+++ b/itachallenge-auth/src/main/java/com/itachallenge/auth/controller/AuthController.java
@@ -6,6 +6,10 @@ import com.itachallenge.auth.exception.CustomInternalServerErrorException;
 import com.itachallenge.auth.service.IAuthService;
 import com.itachallenge.auth.service.IJwtService;
 import com.itachallenge.auth.service.IUserService;
+import io.swagger.v3.oas.annotations.Operation;
+import io.swagger.v3.oas.annotations.media.Content;
+import io.swagger.v3.oas.annotations.media.Schema;
+import io.swagger.v3.oas.annotations.responses.ApiResponse;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Value;
@@ -17,6 +21,7 @@ import io.jsonwebtoken.Jwts;
 import io.jsonwebtoken.ExpiredJwtException;
 import io.jsonwebtoken.JwtException;
 
+import org.springframework.web.server.ResponseStatusException;
 import reactor.core.publisher.Mono;
 
 import java.util.HashMap;
@@ -34,6 +39,7 @@ public class AuthController {
     public static final String X_AUTHENTICATION_STATUS = "X-Authentication-Status";
     private static final String MESSAGE_KEY = "message";
     private static final String LOGOUT_SUCCESS = "Logout successful";
+    private static final String BEARER_KEY = "Bearer ";
 
     private final IAuthService authService;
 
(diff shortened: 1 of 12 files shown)


tonijimenez72:
I’m closing this PR because it includes changes that go beyond the scope of my user story.

tonijimenez72:
I’m closing this PR because it includes changes that go beyond the scope of my user story.

IsmaPeiro:
Hi, I’ve been reviewing the code and everything looks good. I tested it in Postman, but I could only try it with an expired token and a malformed one, and I’ve confirmed that it returns the appropriate responses: a 200 for the expired token and a 401 for the malformed one.

tonijimenez72:
I've sent to you a recent token on Discord for testing.

IsmaPeiro:
> I've sent to you a recent token on Discord for testing.

Thanks, here is the token you sent me, decoded:
image
I was able to test it. I changed the role of the original token from ADMIN to USER:
image
the token that I received, once decoded, contains the USER role:
image
I then changed the role of this new token back to ADMIN:
image
and the decoded output shows that the change was applied correctly:
image
In my opinion, it works correctly, well done.

IsmaPeiro (approved):

IngaD89:
Overall, the code is well-structured, clean, and easy to understand. Here are a few recommendations to consider:

JwtService:
In the switchRole() method, it would be good practice to add a validation check for the current role value before performing the switch. This helps prevent unexpected behavior in case the token contains an unrecognized or tampered role. You could explicitly verify that the current role is either ADMIN or USER before toggling it, and throw an exception if the value is invalid.

The extractAllClaims() method:
This method is currently handling HTTP responses (e.g., throwing a ResponseStatusException). It would be better to separate these concerns. For example, the claim extraction logic could throw a custom exception, and let the controller decide how to handle the response accordingly.

AuthController:
In the switchRole method, as a best practice, you could extract the logic that validates and retrieves the token from the authHeader into a separate method, for example, extractBearerToken(String authHeader). This would improve code readability and also make the token extraction reusable in other cases.

tonijimenez72:
Thanks for your reviews. 

I'm gonna check your recommendations.

tonijimenez72:
JwtService:
- Added role control. Returns 400 BAD_REQUEST.
** Empty or null: "New role must be provided."
** Not USER or ADMIN: "Requested role change is not allowed."
** Same role: "New role is the same as current role."

- extractAllClaims() updated.

AuthController:
- Refactor applied.

IngaD89 (approved):

tonijimenez72:
> > I've sent to you a recent token on Discord for testing.
> 
> Thanks, here is the token you sent me, decoded: image I was able to test it. I changed the role of the original token from ADMIN to USER: image the token that I received, once decoded, contains the USER role: image I then changed the role of this new token back to ADMIN: image and the decoded output shows that the change was applied correctly: image In my opinion, it works correctly, well done.

itachallenge-auth/src/main/java/com/itachallenge/auth/controller/AuthController.java
```diff
@@ -34,6 +39,7 @@ public class AuthController {
     public static final String X_AUTHENTICATION_STATUS = "X-Authentication-Status";
     private static final String MESSAGE_KEY = "message";
     private static final String LOGOUT_SUCCESS = "Logout successful";
+    private static final String BEARER_KEY = "Bearer ";
```
ivilarop: Does this change apply to this task? If it is changed, shouldn't the other endpoints that use Bearer be refactored?

itachallenge-auth/src/main/java/com/itachallenge/auth/service/JwtService.java
```diff
@@ -64,4 +64,73 @@ public SecretKey getSigningKey() {
...
+        }
+
+        if (requestedRole.equalsIgnoreCase(currentRole)) {
+            log.warn("Role change rejected: requested role '{}' is same as current role '{}'.",
+                    requestedRole, currentRole);
+            throw new InvalidRoleChangeRequestException("New role is the same as current role.");
+        }
+
+        boolean isAllowed =
+                ("ADMIN".equalsIgnoreCase(currentRole) && "USER".equalsIgnoreCase(requestedRole)) ||
```
ivilarop: avoid hardcode

itachallenge-auth/src/main/java/com/itachallenge/auth/service/JwtService.java
```diff
@@ -64,4 +64,73 @@ public SecretKey getSigningKey() {
...
+            throw new JwtException("Invalid or tampered token: " + e.getMessage(), e);
+        }
+    }
+
+    @Override
+    public String switchRole(String token, String requestedRole) {
+        Claims claims = extractAllClaims(token);
+        String currentRole = claims.get("role", String.class);
+
+        validateRoleChange(currentRole, requestedRole);
```
ivilarop: move to enum and study best option for validate role

ivilarop (changes requested):

itachallenge-auth/src/main/java/com/itachallenge/auth/controller/AuthController.java
tonijimenez72: The BEARER_KEY constant is used in my new endpoint and could also be reused in other endpoints such as /logout. However, applying this refactor globally is outside the scope of the current user story. I'll create a technical debt issue to address it separately.

itachallenge-auth/src/main/java/com/itachallenge/auth/service/JwtService.java
tonijimenez72: UserRole control moved to enum.
 
 Returns 400 BAD_REQUEST.
* Empty or null: "New role must be provided."
* Not USER or ADMIN: "Requested role change is not allowed."
* Same role: "New role is the same as current role.

itachallenge-auth/src/main/java/com/itachallenge/auth/controller/AuthController.java
ivilarop: And wouldn't it have been better to use the carrier like the rest of the endpoints and create a technical debt to regularize everything and not leave it inconsistent?

tonijimenez72: I created the constant to fix a Sonar error.

ivilarop: ok

itachallenge-auth/src/main/java/com/itachallenge/auth/controller/AuthController.java
```diff
@@ -160,5 +166,34 @@ public Mono<ResponseEntity<Map<String, String>>> logout(
...
+            @RequestBody SwitchRoleRequest request) {
+
+        String token = extractBearerToken(authHeader);
+        String newToken = jwtService.switchRole(token, request.getNewRole());
+        log.info("Switch-role successful for token");
+        return Mono.just(ResponseEntity.ok(Map.of("token", newToken)));
+    }
+
+
+    private String extractBearerToken(String authHeader) {
```
ivilarop: This method partially breaks the SRP principle because:

It validates and transforms a JWT token.

This is already part of the security/authentication domain, which corresponds to the JWTService, not the controller.

itachallenge-auth/src/main/java/com/itachallenge/auth/enums/UserRole.java
ivilarop, line 1: 💡 Improvement Suggestion – Readability in the UserRole Class

You've made a reasonable decision to centralize the role change validation logic within the UserRole enum, which makes sense if the roles are well-defined (as in this case: USER and ADMIN) and aren't expected to grow in the short term.

🔍 That said, the validateRoleChange(...) method currently contains many consecutive if statements, which can hinder readability and maintainability even in simple scenarios.

✅ Suggestion: You might consider refactoring the conditional blocks into more expressive private helper methods.

This doesn't mean breaking the current pattern or moving the logic out of the enum; it simply improves its clarity and readability.

📌 Regarding the PR description: it correctly explains that the validation logic is in the enum, but you could make a brief note about why you chose to include all the logic directly there (e.g., "since the roles are fixed and limited, delegating to an external service is not considered necessary for now").

ivilarop (changes requested):

itachallenge-auth/src/main/java/com/itachallenge/auth/controller/AuthController.java
tonijimenez72: Refactorized.

Also moved BEARER_KEY  to JwtService. Now the technical debt could change to "Refactor POST /logout to use the method extractBearerToken(...) from JwtService".

itachallenge-auth/src/main/java/com/itachallenge/auth/enums/UserRole.java
tonijimenez72: Refactorized enum.

itachallenge-auth/src/main/java/com/itachallenge/auth/controller/AuthController.java
ivilarop: OK. If the technical debt is aligned with the new requirements, we can proceed to close the conversation. Well done ;)

itachallenge-auth/src/main/java/com/itachallenge/auth/enums/UserRole.java
ivilarop: ok well done

ivilarop (approved):

PR #884, opened 2025-05-22
feat(user): Hardcode getAllSolutionsByUser endpoint response for frontend (#435)

IsmaPeiro:
This change introduces a new controller action that exposes a reactive endpoint at GET /users/{userId}/solutions. Instead of querying the database, it returns a fixed list of two UserSolutionResponseDto objects. The primary goal is to allow Front-end teams to integrate against this contract and iterate on UI workflows while backend persistence is still under construction.

Originally it was implemented by filtering by challengeId, but in a meeting we held with the Frontend team, it was suggested that it would be preferable to return an array of all a user’s solutions, and that’s what was agreed upon.

From the Frontend team, it was requested to do it this way mainly for two reasons:

Fewer backend calls: we avoid making a request for each challengeId.

Flexibility: if tomorrow we want to display an overall list of progress or statistics, we already have all the data available.

That said, if in the future we find that this approach hurts performance (for example, if the payload becomes too large), we can create a dedicated endpoint—/users/{userId}/challenges/{challengeId}/solution—to fetch only a single challenge’s solution. For now, given the expected data volume, we believe this gives us faster development and better maintainability.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * Issue #871  Secured POST endpoint addChallenge for creating a challenge in the Challenge Microservice.
+* PR #884: Added a hardcoded GET endpoint to return all of a user’s challenge solutions.
 * Issue #864: Added JWT authentication to logout endpoint in User microservice.
 * PR #872: Added error handling for malformed tag UUIDs and duplicate UUID detection in TagServiceImpl.getValidatedTags
 * PR #866: Added tag validation in the addChallenge endpoint.
--- a/itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
+++ b/itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
@@ -19,6 +19,7 @@ import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.validation.annotation.Validated;
 import org.springframework.web.bind.annotation.*;
+import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
 import java.util.Set;
@@ -397,5 +398,32 @@ public class UserController {
                     log.info("Retrieved {} bookmarked challenges for user {}", bookmarks.size(), userId);
                     return ResponseEntity.ok().body(bookmarks);
                 });
+    }
+    
+    @Operation(
+            summary = "Retrieve all solutions for a user.",
+            parameters = {
+                    @Parameter(
+                            name = "userId",
+                            in = ParameterIn.PATH,
+                            required = true,
+                            description = "User UUID")
+            },
+            responses = {
+                    @ApiResponse(responseCode = "200", description = "Solutions found"),
+                    @ApiResponse(responseCode = "404", description = "Solutions not found"),
+                    @ApiResponse(responseCode = "400", description = "Invalid UUID"),
+                    @ApiResponse(responseCode = "500", description = "Unexpected error")
+            }
+    )
+    @GetMapping(
+            path = "/users/{userId}/solutions"
+    )
+    public Mono<ResponseEntity<Flux<UserSolutionResponseDto>>> getAllSolutions(
+            @PathVariable String userId
+    ) {
+        return Mono.just(ResponseEntity.ok()
+                .body(userSolutionService.getAllSolutionsByUser(userId))
+        );
     }
 }
--- a/itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
+++ b/itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
@@ -419,7 +419,7 @@ public class UserController {
     @GetMapping(
             path = "/users/{userId}/solutions"
     )
-    public Mono<ResponseEntity<Flux<UserSolutionResponseDto>>> getAllSolutions(
+    public Mono<ResponseEntity<Flux<UserSolutionResponseDto>>> getAllSolutionsByUser(
             @PathVariable String userId
     ) {
         return Mono.just(ResponseEntity.ok()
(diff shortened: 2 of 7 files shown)


tonijimenez72:
The endpoint accepts any String. Does it fall under the category of hardcoded?

IsmaPeiro:
> The endpoint accepts any String. Does it fall under the category of hardcoded?

Yes, the goal is for the frontend to be able to use the endpoint by passing any UUID and simply return the hardcoded solutions, without currently handling errors such as malformed UUIDs

tonijimenez72 (approved):
Reviewed and approved.

itachallenge-user/src/main/java/com/itachallenge/user/service/UserSolutionServiceImpl.java
```diff
@@ -77,7 +78,27 @@ private Mono<UserSolutionDocument> saveValidSolution(UUID userUuid, UUID challen
                     return userSolutionRepository.save(userSolutionDocument);
                 }));
     }
-
+    
+    @Override
+    public Flux<UserSolutionResponseDto> getAllSolutionsByUser(String userId) {
```
ivilarop, line 83: Great work with this unmock, it allows you to move forward to the frontend without blocking. As a best practice suggestion, I recommend adding TODO comments where hardcoded data is currently returned, to demonstrate that this logic is temporary and will be replaced with a real integration (e.g., with MongoDB).

This helps you and any other developers understand that the code is not complete and that a more robust implementation is planned for another stage or branch.

for example: // TODO: Replace this mock logic with an actual query to MongoDB that fetches the user's solutions.

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
```diff
@@ -398,4 +399,31 @@ public Mono<ResponseEntity<Set<UUID>>> getUserBookmarks(@PathVariable String use
...
+                    @ApiResponse(responseCode = "404", description = "Solutions not found"),
+                    @ApiResponse(responseCode = "400", description = "Invalid UUID"),
+                    @ApiResponse(responseCode = "500", description = "Unexpected error")
+            }
+    )
+    @GetMapping(
+            path = "/users/{userId}/solutions"
+    )
+    public Mono<ResponseEntity<Flux<UserSolutionResponseDto>>> getAllSolutions(
+            @PathVariable String userId
```
ivilarop, line 423: Why did you decide to return all of the user's solutions (/users/{userId}/solutions) instead of filtering by a challengeId, for example with an endpoint like /users/{userId}/challenges/{challengeId}/solution?

This isn't necessarily incorrect, but I'd like to understand your reasoning. In many cases, returning all solutions can be useful (such as in a dashboard), but if the frontend only needs one solution per challenge, it might be more efficient to filter directly from the backend.

👉 Is this due to frontend requirements? For simplicity for this PR? Is this planned to be separated later?

If we make it clear when we look at the history, it will be possible to understand why it was decided to do it this way.

ivilarop:
Does the PR title match the wiki guide title?

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+* PR #884: Added a hardcoded GET endpoint to return all of a user’s challenge solutions.
```
ivilarop, line 7: When you have the final implementation, will you leave this change?

ivilarop (changes requested):

CHANGELOG.md
IsmaPeiro: I’ve looked into it, and it seems best to keep it. Then, once the final method is in place, add a line indicating that it has been replaced by the definitive method, so there’s a clear record.

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
IsmaPeiro: Originally it was implemented by filtering on challengeId, but in the meeting we had with Luís and Alberto, Luís suggested it would be preferable to return an array of all a user’s solutions, and that’s what Alberto (on the Frontend) and I agreed to do.

From the Frontend team, it was requested to do it this way mainly for two reasons:

Fewer backend calls: we avoid making a request for each challengeId.

Flexibility: if tomorrow we want to display an overall list of progress or statistics, we already have all the data available.

That said, if in the future we find that this approach hurts performance (for example, if the payload becomes too large), we can create a dedicated endpoint—/users/{userId}/challenges/{challengeId}/solution—to fetch only a single challenge’s solution. For now, given the expected data volume, we believe this gives us faster development and better maintainability.

itachallenge-user/src/main/java/com/itachallenge/user/service/UserSolutionServiceImpl.java
IsmaPeiro: Thank you—that makes sense; I’ll add the comment.

IsmaPeiro:
> Does the PR title match the wiki guide title?

It’s true—the PR’s name might not make its purpose clear, since it only indicates that an endpoint is being hard-coded, but not which one. Would this be better?
Feature/435 hardcode getAllSolutionsByUser endpoint response for frontend
In fact, now that you mention it, the endpoint name in the controller isn’t clear either. It’s currently called getAllSolutions, but it would be better as getAllSolutionsByUser, since the former can be confusing.

CHANGELOG.md
ivilarop: perfect it seems ideal to me

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
ivilarop: our reply gives useful context on who made the decision, but to make the reasoning technically sound and clear for future readers, it would help to also explain:

> What specific frontend requirement motivated this change?
> What problem does returning all the user's solutions solve?
> Why is it better than filtering by challengeId in this particular case?

Additionally, this approach has a potential downside:
While it's true that students will have a small number of solutions now, returning all of them could become a performance bottleneck in the future as the platform grows.

That's why it’s important to document the technical pros and cons, and make it clear whether this is a temporary trade-off or a long-term design choice.

This way, future contributors (or even yourself later on) will easily understand why this was done, and whether it may need to be revisited.

It would be good to add the explanation to the definition of the PR

itachallenge-user/src/main/java/com/itachallenge/user/service/UserSolutionServiceImpl.java
ivilarop: ok perfect

ivilarop (approved):

PR #885, opened 2025-05-22
feat (challenge): Modify method to increase times solved counter (#415)

RustyGearBox:
This pull request contains the changes to be ready to increment the timesSolved counter when called from the user microservice after the challenge has been completed.

### 📄 Deleted unnecesary method addChallengeToSolved from UserServiceImpl

Deleted unnecessary method `addChallengeToSolved()` from `UserServiceImpl`. This method was part of the previous and wrong way of handling the counter increment, but it's no longer needed.
- **UPDATE** - `IUserService`: Deleted method from the inteface `IUserService`.
- **UPDATE** - `UserServiceImpl`: Deleted method from the implementation `UserServiceImpl`.
- **UPDATE** - `UserServiceImpTest`: The test related to the method `addChallengeToSolved()` from `UserServiceImpl` has been deleted.

###  📄 Fixed wrong Index JsonProperty

Fixed the value of the JsonProperty index of `ChallengeDto`. 
- **UPDATE** - `ChallengeDto`: The value of the JsonProperty Index has been changed from 12 to 13. Because bookmarks also was indexed as 12.

### 📄 Adapted ChallengeSolvedController for auto-call

The `ChallengeSolvedController` has been modified to be able to recieve the call of the `user.microservice` and his test too.
- **UPDATE** - `ChallengeSolvedController`: Now doesnt require the userId either the authoritation only the **ChallengeId**.
- **UPDATE** - `ChallengeSolvedControllerTest`: Has been modifiy to match with the modifications of `ChallengeSolvedController`.

### 📄 Updated ChallengeServiceImpl to don't need userId

The `ChallengeServiceImpl` has been updated to no longer require the `userId` and, once the counter is incremented, to send a **SolvedDto**.
- **UPDATE** - `IChallengeSevice`: The field **userId** has been removed.
- **UPDATE** - `ChallengeServiceImpl`: The service doesnt use the userId, and returns a **SolvedDto** onces the **timesSolved** has been increased.
- **UPDATE** - `ChallengeServiceImplTest`: The test related to the method  `addChallengeToSolved()` from `ChallengeServiceImpl` has been updated and other has been deleted, because it doesn't use userId anymore.

### DIFF
--- a/itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeSolvedController.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeSolvedController.java
@@ -1,10 +1,7 @@
 package com.itachallenge.challenge.controller;
 
 import com.itachallenge.challenge.dto.SolvedDto;
-import com.itachallenge.challenge.exception.BadRequestException;
-import com.itachallenge.challenge.exception.JwtException;
 import com.itachallenge.challenge.service.IChallengeService;
-import com.itachallenge.challenge.service.IJwtService;
 import io.swagger.v3.oas.annotations.Operation;
 import io.swagger.v3.oas.annotations.media.Content;
 import io.swagger.v3.oas.annotations.media.Schema;
@@ -13,6 +10,7 @@ import lombok.RequiredArgsConstructor;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.validation.annotation.Validated;
 import org.springframework.web.bind.annotation.*;
@@ -26,10 +24,9 @@ public class ChallengeSolvedController {
 
     private static final Logger log = LoggerFactory.getLogger(ChallengeSolvedController.class);
 
-    private IChallengeService challengeService;
-    private IJwtService jwtService;
+    private final IChallengeService challengeService;
 
-    @PostMapping("/challenges/{challengeId}/solved")
+    @PostMapping("/challenges/{challengeId}/ended")
     @Operation(
             operationId = "Add a challenge to User's solved challenges.",
             summary = "Add a challenge to solved challenges.",
@@ -41,13 +38,17 @@ public class ChallengeSolvedController {
                     @ApiResponse(responseCode = "500", description = "Internal Server Error")
             }
     )
-    public Mono<ResponseEntity<SolvedDto>> addChallengeToSolved(
-            @PathVariable String challengeId,
-            @RequestHeader(name = "Authorization", required = false) String authHeader) {
-        return Mono.fromCallable(() -> jwtService.getUserUuIdFromAuthenticationHeader(authHeader))
-                .onErrorMap(JwtException.class, e -> new BadRequestException(e.getMessage()))
-                .flatMap(userId -> challengeService.addChallengeToSolved(challengeId, userId))
-                .doOnError(error -> log.error("Error adding challenge to solved: {}", error.getMessage()))
-                .map(ResponseEntity::ok);
+    public Mono<ResponseEntity<SolvedDto>> addChallengeToSolved(@PathVariable String challengeId) {
+        return challengeService.addChallengeToSolved(challengeId)
+                .map(solvedDto -> {
+                    if (Boolean.TRUE.equals(solvedDto.isSolved())) {
+                        log.info("Challenge '{}' has increased his value timesSolved", challengeId);
+                        return ResponseEntity.status(HttpStatus.CREATED).body(solvedDto);
+                    } else {
+                        log.info("Challenge '{}' has not increased his value timesSolved", challengeId);
+                        return ResponseEntity.ok(solvedDto);
+                    }
+                });
     }
+
 }
--- a/itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeSolvedController.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeSolvedController.java
@@ -43,10 +43,9 @@ public class ChallengeSolvedController {
                     if (solvedDto.isSolved()) {
                         log.info("Challenge '{}' has increased his value timesSolved", challengeId);
                         return ResponseEntity.status(HttpStatus.CREATED).body(solvedDto);
-                    } else {
-                        log.info("Challenge '{}' could not be found, so the value timesSoved has not been increased", challengeId);
-                        return ResponseEntity.ok(solvedDto);
                     }
+                    log.info("Challenge '{}' could not be found, so the value timesSoved has not been increased", challengeId);
+                    return ResponseEntity.ok(solvedDto);
                 });
     }
 
--- a/itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeSolvedController.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeSolvedController.java
@@ -26,7 +26,7 @@ public class ChallengeSolvedController {
 
     private final IChallengeService challengeService;
 
-    @PostMapping("/challenges/{challengeId}/ended")
+    @PostMapping("/challenges/ended/{challengeId}")
     @Operation(
             operationId = "Add a challenge to User's solved challenges.",
             summary = "Add a challenge to solved challenges.",
--- a/itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeSolvedController.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeSolvedController.java
@@ -33,7 +33,6 @@ public class ChallengeSolvedController {
             description = "The ID Challenge sent through the URI is added to the user's solved challenges. User Id is determined from the headers.",
             responses = {
                     @ApiResponse(responseCode = "200", content = {@Content(schema = @Schema(implementation = SolvedDto.class), mediaType = "application/json")}),
-                    @ApiResponse(responseCode = "400", description = "Missing or invalid authorization header."),
                     @ApiResponse(responseCode = "404", description = "The Challenge with given Id was not found."),
                     @ApiResponse(responseCode = "500", description = "Internal Server Error")
             }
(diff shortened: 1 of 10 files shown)


tonijimenez72:
### Testing endpoint 

- Before: timesSolved = 5
a1d7e044-4a78-445e-9dba-da76a5863473

- After: timesSolved = 6
a1d7e044-4a78-445e-9dba-da76a5863473

- Forcing Bad Request: error 400
3116bc01-ccbe-4156-b768-c5db5520dd8c

- Forcing Not Found Request: error 404
b47d3d12-538a-493e-a932-daf3c9c0283e

tonijimenez72 (approved):
Reviewed and approved.

ivilarop:
changelog?

itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeSolvedController.java
```diff
@@ -41,13 +38,17 @@ public class ChallengeSolvedController {
...
     )
-    public Mono<ResponseEntity<SolvedDto>> addChallengeToSolved(
-            @PathVariable String challengeId,
-            @RequestHeader(name = "Authorization", required = false) String authHeader) {
-        return Mono.fromCallable(() -> jwtService.getUserUuIdFromAuthenticationHeader(authHeader))
-                .onErrorMap(JwtException.class, e -> new BadRequestException(e.getMessage()))
-                .flatMap(userId -> challengeService.addChallengeToSolved(challengeId, userId))
-                .doOnError(error -> log.error("Error adding challenge to solved: {}", error.getMessage()))
-                .map(ResponseEntity::ok);
+    public Mono<ResponseEntity<SolvedDto>> addChallengeToSolved(@PathVariable String challengeId) {
```
ivilarop, line 40: Shouldn't it be secured? Can we add it as technical debt? Can we include it in the PR description for the record? Should we have taken this into account when defining the user story and perhaps made it optional?

ivilarop: That 400 authorization error code can no longer be returned because:

There is no validation of the Authorization header

There is no call to jwtService.getUserUuidFromAuthenticationHeader(...)

➡️ The endpoint is documenting a possible response that can no longer occur.

itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeSolvedController.java
```diff
@@ -41,13 +38,17 @@ public class ChallengeSolvedController {
...
+        return challengeService.addChallengeToSolved(challengeId)
+                .map(solvedDto -> {
+                    if (Boolean.TRUE.equals(solvedDto.isSolved())) {
```
ivilarop: Boolean.TRUE.equals(...) is very old-school Java.
It works, but if isSolved() already returns a boolean (not Boolean), it would be clearer: if (solvedDto.isSolved()) { ... }.

if/else can be separated if you want to make an early return.

itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeSolvedController.java
```diff
@@ -41,13 +38,17 @@ public class ChallengeSolvedController {
...
+                        log.info("Challenge '{}' has increased his value timesSolved", challengeId);
+                        return ResponseEntity.status(HttpStatus.CREATED).body(solvedDto);
+                    } else {
+                        log.info("Challenge '{}' has not increased his value timesSolved", challengeId);
```
ivilarop: The log message "Challenge '{}' has not increased its value timesSolved" is functional, but could be more informative. Would it be possible to add the reason (for example, if it was already solved)? This would help with future tracing and debugging.

itachallenge-challenge/src/main/java/com/itachallenge/challenge/service/ChallengeServiceImpl.java
```diff
@@ -454,32 +454,20 @@ public Mono<BookmarkDto> removeChallengeFromBookmarks(String challengeId, String
...
-                    return challengeRepository.findByUuid(challengeUuid)
-                            .switchIfEmpty(Mono.error(new ChallengeNotFoundException(
-                                    String.format(CHALLENGE_NOT_FOUND_ERROR, challengeUuid))))
-                            .flatMap(challenge -> {
-                                log.info("It should be connected to user service using the fields {} and {}", challengeUuid, userUuid);
-
-                                //The part of the user service is not implemented yet, it should be connected to the user service in the future.
-
-                                if (Optional.ofNullable(challenge.getTimesSolved()).orElse(0) == 0) {
+        return challengeIdMono
```
ivilarop, line 448: I notice that the validation to determine whether the challenge has already been solved (`timesSolved == 0`) has been removed. Currently, the method always increments and returns `SolvedDto(true)` without checking the previous state.

Is this intentional? If the goal is to make the endpoint idempotent, it might be better to check the current state first and return `SolvedDto(false)` if it has already been solved.

Otherwise, we could consider throwing an exception like `ChallengeAlreadySolvedException`.

**This also relates to the log message in the controller** that says `"Challenge '{}' has not increased his value timesSolved"` — that message assumes a conditional check that no longer exists in the service layer.

ivilarop (changes requested):

itachallenge-challenge/src/main/java/com/itachallenge/challenge/service/ChallengeServiceImpl.java
RustyGearBox: If a challenge alredy has been solved, the method in the user.micro in charge of saving the user solution will not call the endpoint in charge to increase the counter timesSolved.

itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeSolvedController.java
RustyGearBox: I think the security must be done in the micro.user, because the flow start there. But it could be added in the future as technical debt. About the responseCode 400 it's true, must be removed until the security is build.

itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeSolvedController.java
RustyGearBox: I have changed the changes asked into: if (solvedDto.isSolved()

itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeSolvedController.java
RustyGearBox: The log message have been modified to especify that the timesSolved counter couln't be increased due the challenge could not be found.

RustyGearBox:
> changelog?

Has been updated in the next PR

itachallenge-challenge/src/main/java/com/itachallenge/challenge/service/ChallengeServiceImpl.java
RustyGearBox: This endpoint only will be called by the backend from the method incharge of saving the user solution. The method that's saves the solution should be the one responsable to check if the challenge alredy has been solved bu the user.
That's why i have removed that part.

itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeSolvedController.java
ivilarop: Yes, it's always good practice to secure it in case direct calls are made to this endpoint externally. Add it as technical debt, and when you have it, we'll resolve this conversation.

itachallenge-challenge/src/main/java/com/itachallenge/challenge/service/ChallengeServiceImpl.java
ivilarop: And what would happen if we called this endpoint directly, for example via Postman?

itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeSolvedController.java
ivilarop: ok thks! you can resolve

itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeSolvedController.java
ivilarop: In order to align:

What should be the goal of this endpoint?

🎯 Endpoint Responsibility addChallengeToSolved(String ChallengeId)
This endpoint, based on its name and expected functionality, should:

1-Check if the challenge exists
2-Check if the user has already solved it
3-Increment timesSolved only if it's the first time
4-Return a SolvedDto(isSolved, timesSolved) indicating whether it has been updated or not

Is this correct?

ivilarop: The current log message:

log.info("Challenge '{}' could not be found, so the timesSolved value has not been incremented", challengeId);

is not accurate in this context. At this point in the code, the challenge has already been successfully retrieved; otherwise, findByUuid(...).switchIfEmpty(...) would have previously thrown a ChallengeNotFoundException.

This log can be misleading during debugging, suggesting that a challenge is missing when it is not.

🔁 Suggestion: Update the log to reflect the current reason why timesSolved was not incremented. For example:

"Challenge '{}' has already been solved. The timesSolved value has not been incremented."

This would improve traceability and better align with the endpoint's current behavior.

itachallenge-challenge/src/main/java/com/itachallenge/challenge/service/ChallengeServiceImpl.java
ivilarop: Although this endpoint is currently only called from user.micro, it is generally recommended to double-check whether the issue has been resolved before increasing timesSolved.

Microservices should protect their own domain logic, even when coordination is handled externally. Adding this check would allow:

Making the endpoint idempotent

Avoiding inconsistencies if it is ever called from another source

Aligning with the existing return model (SolvedDto) and record structure

ivilarop: ✅ Thank you all for the contributions. We confirm that the current design — where the increment of timesSolved is handled exclusively by the challenge microservice, without requiring userId or authentication — is valid in the current context, since:

The endpoint is only expected to be called internally by the user microservice.

The logic to determine whether a challenge has been completed is correctly handled in user.

🔒 However, a security risk has been identified: the endpoint is currently unprotected and could be invoked directly by external clients (e.g., Postman, browser, etc.).

📝 For this reason, we have created a technical debt task aimed at restricting access to this endpoint via APISIX, ensuring that only the user microservice can call it in production.

With this improvement logged and planned, we approve the current implementation, assuming that access control will be addressed as part of the hardening or microservice integration security phase.

https://tree.taiga.io/project/luisvi-ita-challenges/issue/490

ivilarop (approved):

PR #887, opened 2025-05-26
feat(user): Replaced the hardcoded return of a user’s solutions with the actual method (#406)

IsmaPeiro:
### Description:
User Story: [https://tree.taiga.io/project/luisvi-ita-challenges/us/394?milestone=456189](url)
This pull request removes the temporary hardcoded response for the getAllSolutionsByUser endpoint and replaces it with a fully functional implementation that retrieves data from MongoDB. Key updates include:

- **Repository Layer**: Introduced a method to query all user solutions by their UUID, enabling reactive access to stored documents.

- **Service Layer**: Updated the service interface to reflect the new repository-driven contract. In the implementation, the service now transforms retrieved documents into response DTOs, and throws a NotFoundException when no solutions are found.

- **Controller layer** : Enhanced the handler to emit informative logs for each retrieved solution and when no solutions are found. 
  
- **Unit Tests**: Revised existing tests to cover both scenarios—when no solutions exist (asserting the appropriate exception is thrown) and when multiple solutions are returned (verifying correct mapping to the response DTO).

### DIFF
--- a/itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
+++ b/itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
@@ -29,24 +29,24 @@ import java.util.UUID;
 @Validated
 @RequestMapping(value = "/itachallenge/api/v1/user")
 public class UserController {
-    
+
     private static final Logger log = LoggerFactory.getLogger(UserController.class);
     public static final String X_VALIDATION_STATUS = "X-Validation-Status";
     public static final String X_GITHUB_USERNAME = "X-Github-Username";
-    
+
     private final UserService userService;
     private final IUserSolutionService userSolutionService;
-    
+
     public UserController(UserService userService, IUserSolutionService userSolutionService) {
         this.userService = userService;
         this.userSolutionService = userSolutionService;
     }
-    
+
     @GetMapping(value = "/test")
     public String test() {
         return "Hello from ITA Challenge UserController!!!";
     }
-    
+
     @Operation(
             summary = "Retrieve User",
             description = "Retrieves user details for the given GitHub username if it exists in the database.",
@@ -81,7 +81,7 @@ public class UserController {
                     )
             }
     )
-    
+
     @GetMapping("/users/{githubUsername}")
     public Mono<ResponseEntity<UserDocument>> getUser(@PathVariable @ValidGithubUsername String githubUsername) {
         return userService.getUser(githubUsername)
@@ -93,7 +93,7 @@ public class UserController {
                             .body(user);
                 });
     }
-    
+
     @Operation(
             summary = "Add Challenge to User Favorite Challenges",
             description = "Adds challenge to user favorites",
@@ -139,7 +139,7 @@ public class UserController {
                     )
             }
     )
-    
+
     @PostMapping("/users/{userId}/favorites/{challengeId}")
     public Mono<ResponseEntity<Boolean>> addToFavorites(@PathVariable String userId, @PathVariable String challengeId) {
         return userService.addChallengeToFavorites(userId, challengeId)
@@ -153,7 +153,7 @@ public class UserController {
                     }
                 });
     }
-    
+
     @PutMapping(path = "/solution")
     @Operation(
             summary = "perform a solution, adding challenge,language,user, status and the corresponding solution text.",
@@ -168,7 +168,7 @@ public class UserController {
     )
     public Mono<ResponseEntity<UserSolutionResponseDto>> addSolution(
             @Valid @RequestBody UserSolutionRequestDto userSolutionDto) {
-        
+
         return userSolutionService.addSolution(userSolutionDto)
                 .map(savedUserSolutionDto ->
                         ResponseEntity.status(HttpStatus.OK).body(savedUserSolutionDto)
@@ -220,7 +220,7 @@ public class UserController {
                     )
             }
     )
-    
+
     @PostMapping("/users/{userId}/bookmarks/{challengeId}")
     public Mono<ResponseEntity<Boolean>> addToBookmarks(@PathVariable String userId, @PathVariable String challengeId) {
         return userService.addChallengeToBookmarks(userId, challengeId)
@@ -234,8 +234,8 @@ public class UserController {
                     }
                 });
     }
-    
-    
+
+
     @Operation(
             summary = "Delete Challenge from User Favorite Challenges",
             description = "Deletes challenge from user favorites",
@@ -289,7 +289,7 @@ public class UserController {
                     }
                 });
     }
-    
+
     @Operation(
             summary = "Delete Challenge from User Bookmark Challenges",
             description = "Deletes challenge from user bookmarks",
@@ -343,7 +343,7 @@ public class UserController {
                     }
                 });
     }
-    
+
     @Operation(
             summary = "Gets challenges marked as favorites by a user",
             description = "Returns a set of challenge IDs that the specified user has marked as favorites",
@@ -362,7 +362,7 @@ public class UserController {
                     @ApiResponse(responseCode = "500", description = "Unexpected error")
             }
     )
-    
+
     @GetMapping("/users/{userId}/favorites")
     public Mono<ResponseEntity<Set<UUID>>> getUserFavorites(@PathVariable String userId) {
         return userService.getUserFavorites(userId)
@@ -371,7 +371,7 @@ public class UserController {
                     return ResponseEntity.ok().body(favorites);
                 });
     }
-    
+
     @Operation(
             summary = "Gets challenges marked as bookmarks by a user",
             description = "Returns a set of challenge IDs that the specified user has marked as bookmarked",
@@ -390,7 +390,7 @@ public class UserController {
                     @ApiResponse(responseCode = "500", description = "Unexpected error")
             }
     )
-    
+
     @GetMapping("/users/{userId}/bookmarks")
     public Mono<ResponseEntity<Set<UUID>>> getUserBookmarks(@PathVariable String userId) {
         return userService.getUserBookmarks(userId)
--- a/itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
+++ b/itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
@@ -174,7 +174,7 @@ public class UserController {
                         ResponseEntity.status(HttpStatus.OK).body(savedUserSolutionDto)
                 );
     }
-    
+
     @Operation(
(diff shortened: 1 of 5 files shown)


tonijimenez72:
User has solutions: 200 - (Array of solutions)
imatge

User has no solutions: 404 - Solutions not found.
imatge

Invalid user ID: 404 - Invalid UUID string: (string)
imatge

imatge


- The endpoint specifies a 400 API response for "Invalid UUID", but it returns a 404.

- The test getAllSolutions_returns400IfInvalidUUID checks for a 400 response, but it should check for a 404 to align with the expected response.

- Should null or empty User UUID be controlled?

The code is well-structured, clean, and easy to understand. Good job!

IsmaPeiro:
> User has solutions: 200 - (Array of solutions) imatge
> 
> User has no solutions: 404 - Solutions not found. imatge
> 
> Invalid user ID: 404 - Invalid UUID string: (string) imatge
> 
> imatge
> 
> * The endpoint specifies a 400 API response for "Invalid UUID", but it returns a 404.
> * The test getAllSolutions_returns400IfInvalidUUID checks for a 400 response, but it should check for a 404 to align with the expected response.
> * Should null or empty User UUID be controlled?
> 
> The code is well-structured, clean, and easy to understand. Good job!

Hi @tonijimenez72, thanks for the feedback! I’ve made the following changes in response:

I extracted all of the userId checks into a private reactive helper that immediately returns a BadRequestException (400) if the ID is null, empty, or not a valid UUID. This means invalid inputs no longer reach the repository.

I kept the existing behavior of throwing a NotFoundException (404) only when the repository actually returns an empty result, so “no solutions” remains a 404.

I expanded the unit tests to cover every edge case in the public API:

Null, empty, and malformed userId inputs now each trigger a 400 without calling the repository.

I’ve changed the API so that an invalid UUID now returns a 400 Bad Request, and therefore the test getAllSolutions_returns400IfInvalidUUID is now asserting the correct, intended behavior.

image

With these updates, the API now consistently returns 400 for bad IDs and 404 only for genuinely missing data, and our test suite fully documents that behavior. Let me know if you’d like any further tweaks!

tonijimenez72:
Null or empty returns 400.
imatge
Bad request returns 400.
imatge

tonijimenez72 (approved):
Reviewed and approved.

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
```diff
@@ -423,7 +423,14 @@ public Mono<ResponseEntity<Flux<UserSolutionResponseDto>>> getAllSolutionsByUser
             @PathVariable String userId
```
ivilarop, line 422: 👀 I see that the GET /users/{userId}/solutions endpoint exposes a specific user's solutions.
Have you considered whether this endpoint should be secured?

Although it's a GET, it returns potentially sensitive information (the user's solutions), so it may make sense to require authentication and validate that the authenticated user has permission to access that data (e.g., they're the user themselves or have a mentor/admin role).

🔐 In other endpoints like POST /logout, you're already applying validations to the JWT token, so you could follow similar logic here if you see fit.

What was your approach?

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
```diff
@@ -423,7 +423,14 @@ public Mono<ResponseEntity<Flux<UserSolutionResponseDto>>> getAllSolutionsByUser
...
     ) {
         return Mono.just(ResponseEntity.ok()
-                .body(userSolutionService.getAllSolutionsByUser(userId))
+                .body(userSolutionService.getAllSolutionsByUser(userId)
+                        .doOnNext(dto ->
+                                log.info("→ Solution retrieved for user {}: challengeId={}", userId, dto.getChallengeId()))
```
ivilarop: 👀 In the ChallengeController, I see you're using .doOnNext(...) to execute a log.info(...) with the userId and challengeId.

🔁 This log can be useful for traceability, but conceptually it would be more appropriate to move it to the service, which is where you're already building the DTO and applying the business logic.

📉 Additionally, since it's in the controller (and if multiple solutions are issued), this log.info will be executed for each element of the Flux, which can generate a lot of noise in the logs if the user has many solutions or if the request is repeated frequently (for example, during a UI refresh).

✅ Moving this log to the service's map() allows you to centralize traceability and avoid overloading the controller with responsibilities that don't correspond to it.

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
```diff
@@ -423,7 +423,14 @@ public Mono<ResponseEntity<Flux<UserSolutionResponseDto>>> getAllSolutionsByUser
...
+                        .switchIfEmpty(Flux.defer(() -> {
```
ivilarop: In the controller, you're also using .switchIfEmpty(...) to log a log.warn() if no solutions are found.

However, you're already handling the empty case within the userSolutionService itself.

👉 Therefore, this switchIfEmpty in the controller is not only redundant, but it can cause confusion by having two different logics for the same case (one throws an exception, the other returns empty).

💡 I would suggest keeping that responsibility solely in the service and removing it from the controller to avoid duplication and ambiguity.

itachallenge-user/src/main/java/com/itachallenge/user/service/UserSolutionServiceImpl.java
```diff
@@ -81,25 +83,30 @@ private Mono<UserSolutionDocument> saveValidSolution(UUID userUuid, UUID challen
...
-                .languageId("09fabe32-7362-4bfb-ac05-b7bf854c6e0f")
-                .solutionText("This is the submitted solution")
-                .build();
-        
-        return Flux.just(sol1, sol2);
+        return validateAndParseUuid(userId)
+                .flatMapMany(uuid ->
+                userSolutionRepository
+                        .findAllByUserId(UUID.fromString(userId))
+                        .switchIfEmpty(Mono.error(new NotFoundException("Solutions not found."))))
```
ivilarop: 👀 Currently, in userSolutionService.getAllSolutionsByUser(...), you are throwing a NotFoundException when no solutions are found.

I'd like you to consider whether this behavior is the most appropriate:

Does it make sense to consider the fact that a user hasn't yet submitted solutions as an "error"?

Or would it be more correct (and consistent with REST APIs) to return a 200 OK with an empty list?

You might also consider whether this decision (returning 404 vs. 200) should be made from the controller or from the service.
I encourage you to do some research and justify your choice.

ivilarop (changes requested):

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
IsmaPeiro: I’ve removed the .doOnNext(…) from the controller and moved the log.info("→ Solution retrieved…") into the service’s .map(…) when building each DTO. This keeps the controller focused on HTTP handling and centralizes traceability with the business logic.

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
IsmaPeiro: I’ve eliminated the switchIfEmpty(...) (and its warning) from the controller. All empty-case behavior is now solely in the service layer.

itachallenge-user/src/main/java/com/itachallenge/user/service/UserSolutionServiceImpl.java
IsmaPeiro: I was just debating this with our frontend colleague this morning.
By REST convention, a user with zero solutions should get 200 OK + an empty list, not a 404. I’ve updated the service to remove the switchIfEmpty(Mono.error(...)), so findAllByUserId now returns a Flux.empty(). The controller naturally serializes that to [].

Test update: The unit test now uses expectNextCount(0) and verifyComplete() instead of expecting a NotFoundException.

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
IsmaPeiro: I followed the existing pattern in UserController by leaving its endpoints unsecured, but I agree it would be wise to secure them all to avoid exposing sensitive user data. Since this wasn’t part of the original user-story requirements and given our time constraints to meet the sprint deadline, I think it’s more practical to log this as technical debt and tackle it in a future iteration. If you consider it critical, I can try to reprioritize and secure it right away.

Action: Log it in the backlog as technical debt to secure UserController endpoints.

ivilarop: No, don't worry, it's fine to audit it as technical debt, and we'll address it later. But keep this in mind for future developments, since we should have initially detected it in the story and marked it as optional. We can close the conversation. Thank you.

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
ivilarop: ok thks!

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
ivilarop: ok thks!

itachallenge-user/src/main/java/com/itachallenge/user/service/UserSolutionServiceImpl.java
ivilarop: ok thks!

ivilarop (approved):

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
IsmaPeiro: I’ll keep it in mind, thank you very much.

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
IsmaPeiro: Thanks to you.

itachallenge-user/src/main/java/com/itachallenge/user/service/UserSolutionServiceImpl.java
IsmaPeiro: Thanks to you.

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
IsmaPeiro: Thanks to you.

PR #922, opened 2025-06-18
feat(user): add status field to response DTOs (#522)

FerGimenezRoglia:
### Purpose

Add the `status` field (`IN_PROGRESS` or `ENDED`) to the response DTOs in the `user` microservice, allowing the frontend to determine whether a solution is in progress or completed.

Related to task [#522].

---

### Changes made

- Added `status` field to `UserSolutionResponseDto`.
- Added `status` field to `SubmitSolutionResponseDto`.
- Created serialization tests to ensure the `status` field is correctly included in both `UserSolutionResponseDto` and `SubmitSolutionResponseDto`.
- Adjusted related unit tests to reflect the new status field.

---

### Tests performed

- Tested using Postman.  
- Verified that the `status` field appears correctly in the responses of the `PUT /solution` and `GET /users/{userId}/solutions` endpoints.  

---

## PR Author Self-check

- [x] The PR is focused and does only one thing.  
- [x] Branch name follows the convention: `feat(user): add status field to response DTOs (#522)`  
- [x] All related business logic is implemented and tested locally.  
- [x] The code is compiling and passing tests.  
- [x] No console.log or debugging artifacts were left behind.  
- [x] DTOs and endpoints are updated and consistent.  
- [x] I verified that only one solution per user/challenge/language is allowed in IN_PROGRESS.  
- [x] I verified that ENDED solutions are no longer editable.  
- [x] The PR description explains the purpose and context clearly.  
- [x] This PR is part of Issue #922 and follows the task breakdown logic described in the wiki #522.

### DIFF
--- a/itachallenge-user/src/main/java/com/itachallenge/user/dto/SubmitSolutionResponseDto.java
+++ b/itachallenge-user/src/main/java/com/itachallenge/user/dto/SubmitSolutionResponseDto.java
@@ -20,4 +20,7 @@ public class SubmitSolutionResponseDto {
 
     @JsonProperty("timesSolved")
     private Integer timesSolved;
+
+    @JsonProperty("status")
+    private String status;
 }
--- a/itachallenge-user/src/test/java/com/itachallenge/user/dto/UserSolutionResponseDtoTest.java
+++ b/itachallenge-user/src/test/java/com/itachallenge/user/dto/UserSolutionResponseDtoTest.java
@@ -1,6 +1,7 @@
 package com.itachallenge.user.dto;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.itachallenge.user.document.enums.ChallengeStatus;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
@@ -16,6 +17,7 @@ class UserSolutionResponseDtoTest {
     String challengeId = UUID.randomUUID().toString();
     String languageId = UUID.randomUUID().toString();
     String solutionText = "This is my solution";
+    String status = ChallengeStatus.IN_PROGRESS.name();
     UserSolutionResponseDto solutionScoreDto = new UserSolutionResponseDto();
     UserSolutionResponseDto dto1 = UserSolutionResponseDto.builder().build();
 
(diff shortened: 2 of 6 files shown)


ivilarop:
@FerGimenezRoglia  Where is the PR Author Self-check (to include in every pull request description) that appears in the wiki PR checklist?

itachallenge-user/src/test/java/com/itachallenge/user/dto/UserSolutionResponseDtoTest.java
ivilarop, line 1: 👀 Question to ponder regarding test coverage:
I see you've added a test to verify that the status field works correctly with the "IN_PROGRESS" value, which is great 👏

Do you think this would be enough to properly cover the new behavior?
Or would it make sense to add additional tests that cover more status cases (e.g., ENDED) or validate how serialization behaves when there are different functional values?

itachallenge-user/src/main/java/com/itachallenge/user/dto/SubmitSolutionResponseDto.java
ivilarop, line 1: 📌 Improvement suggestion or potential technical debt:
I noticed you've also added the status field to SubmitSolutionResponseDto, but this class doesn't have a test yet.

Do you think it would be a good idea to add at least one serialization test to ensure the field is correctly included in the JSON response?

If you feel it's beyond the scope of this task, we could create a technical debt to cover the tests for this class later.

ivilarop (changes requested):

itachallenge-user/src/main/java/com/itachallenge/user/dto/SubmitSolutionResponseDto.java
FerGimenezRoglia: Thanks for pointing it out! You’re absolutely right — this DTO didn’t have any tests yet, and I agree it’s worth adding at least one serialization test to make sure the status field is correctly included in the JSON output.

I’ll go ahead and include that now so we leave it properly covered and avoid dragging a tech debt here. Should be a quick addition.

itachallenge-user/src/test/java/com/itachallenge/user/dto/UserSolutionResponseDtoTest.java
FerGimenezRoglia: Good point
The current test was focused on validating the baseline scenario (IN_PROGRESS) to ensure the integration point was working. I agree it makes sense to include at least one more functional case like ENDED, both for completeness and to avoid regressions in future changes.

I’ll add that right away so we leave the behavior properly covered across possible states.

itachallenge-user/src/main/java/com/itachallenge/user/dto/SubmitSolutionResponseDto.java
ivilarop: Well done!

itachallenge-user/src/test/java/com/itachallenge/user/dto/UserSolutionResponseDtoTest.java
ivilarop: Well done!

ivilarop (approved):

PR #923, opened 2025-06-19
feat(user): map status in service methods (#523)

FerGimenezRoglia:
### Purpose

Map the `status` field (IN_PROGRESS or ENDED) in the service logic of the `user` microservice, so that solutions saved by users reflect the correct status in the response.

Related to task [#523]

---

### Changes made

- Mapped the `status` field in `UserSolutionServiceImpl.addSolution()` when building the `SubmitSolutionResponseDto`.
- Mapped the `status` field in `getAllSolutionsByUser()` when building each `UserSolutionResponseDto`.
- Included status when mocking `UserSolutionDocument` in `UserSolutionServiceImplTest`.
- Created unit test to verify that a new solution with IN_PROGRESS status is accepted and properly returned.
- Added example requests to `ITA-Challenges.postman_collection.json` to test `IN_PROGRESS` and `ENDED` statuses in solution creation.

---

### Tests performed

- Verified manually via Postman.
- Confirmed that status is returned correctly in both `PUT /solution` and `GET /users/{userId}/solutions`.

## PR Author Self-check

- [x] The PR is focused and does only one thing.
- [x] Branch name follows the convention: `feat(user): map status in service methods (#523)`
- [x] All related business logic is implemented and tested locally.
- [x] The code is compiling and passing tests.
- [x] No console.log or debugging artifacts were left behind.
- [x] The service layer correctly maps `IN_PROGRESS` and `ENDED` status values.
- [x] I verified that updates respect the business rules for editing solutions.
- [x] I verified that ENDED solutions are not modifiable through the service.
- [x] The PR description explains the purpose and context clearly.
- [x] This PR is part of Issue #923 and follows the task breakdown logic described in the wiki (#523).

### DIFF
--- a/itachallenge-user/src/test/java/com/itachallenge/user/service/UserSolutionServiceImplTest.java
+++ b/itachallenge-user/src/test/java/com/itachallenge/user/service/UserSolutionServiceImplTest.java
@@ -217,4 +217,36 @@ class UserSolutionServiceImplTest {
                 })
                 .verifyComplete();
     }
+
+    @Test
+    @DisplayName("addSolution creates new IN_PROGRESS solution and returns response")
+    void addSolutionNewInProgressSolution() {
+        UserSolutionRequestDto request = UserSolutionRequestDto.builder()
+                .userId(userUuid.toString())
+                .challengeId(challengeUuid.toString())
+                .languageId(languageUuid.toString())
+                .status("IN_PROGRESS")  // ⬅️ acá cambiamos ENDED por IN_PROGRESS
+                .solutionText(solutionText)
+                .build();
+
+        when(userSolutionRepository.findByUserIdAndChallengeIdAndLanguageId(userUuid, challengeUuid, languageUuid))
+                .thenReturn(Mono.empty());
+
+        when(userSolutionRepository.save(any(UserSolutionDocument.class)))
+                .thenAnswer(invocation -> Mono.just(invocation.getArgument(0)));
+
+        Mono<SubmitSolutionResponseDto> result = userSolutionService.addSolution(request);
+
+        StepVerifier.create(result)
+                .assertNext(dto -> {
+                    assertEquals(solutionText, dto.getSolutionText());
+                    assertFalse(dto.getIsSolved());
+                    assertEquals("IN_PROGRESS", dto.getStatus());
+                })
+                .verifyComplete();
+
+        verify(userSolutionRepository).findByUserIdAndChallengeIdAndLanguageId(userUuid, challengeUuid, languageUuid);
+        verify(userSolutionRepository).save(any(UserSolutionDocument.class));
+        verifyNoInteractions(challengeService);
+    }
 }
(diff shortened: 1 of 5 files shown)


ivilarop:
@FerGimenezRoglia  

To complete the change and ensure everyone can easily test the behavior from Postman, it would be a good idea to also update the ITA-Challenges.postman_collection.json file:

If two possible values ​​are allowed (IN_PROGRESS, ENDED), consider including both examples to facilitate validation.

This will help us keep the test collections in sync with the implemented functional changes 🚀

ivilarop:
> To complete the change and ensure everyone can easily test the behavior from Postman, it would be a good idea to also update the ITA-Challenges.postman_collection.json file:
> 
> If two possible values ​​are allowed (IN_PROGRESS, ENDED), consider including both examples to facilitate validation.
> 
> This will help us keep the test collections in sync with the implemented functional changes 🚀

FerGimenezRoglia:
@ivilarop 

Thanks for the suggestion
I’ve already updated the ITA-Challenges.postman_collection.json file accordingly.

You'll now find two separate requests under the User folder:

User/solution - IN_PROGRESS

User/solution - ENDED

This should make it easier to validate the behavior directly from Postman.

itachallenge-user/src/test/java/com/itachallenge/user/service/UserSolutionServiceImplTest.java
```diff
@@ -204,6 +204,7 @@ void getAllSolutionsByUser_returnsSolutions() {
                 .challengeId(challengeUuid)
                 .languageId(languageUuid)
                 .solutionAttemptDocument(SolutionAttemptDocument.builder().solutionText("My solution").build())
+                .status(com.itachallenge.user.document.enums.ChallengeStatus.IN_PROGRESS)
```
ivilarop, line 207: 🎯 Since the /users/{userId}/solutions endpoint always returns at most one solution (the most recent one based on its status), it's recommended to cover both possible cases in your unit tests:

✅ When the user has an IN_PROGRESS solution, return that one.

✅ When the user only has one ENDED solution, return that one.

This ensures that the expected behavior is maintained even if the internal logic changes in the future.

itachallenge-user/src/test/java/com/itachallenge/user/service/UserSolutionServiceImplTest.java
ivilarop, line 1: You can add a third test that validates that an empty Flux is returned if there are no solutions.

ivilarop (changes requested):

itachallenge-user/src/test/java/com/itachallenge/user/service/UserSolutionServiceImplTest.java
FerGimenezRoglia: Thanks for the feedback.
The case where the user has a solution with an IN_PROGRESS status is currently covered.
I'm going to add a new test to explicitly cover the case where the user only has a solution with an ENDED status, as you suggested.

itachallenge-user/src/test/java/com/itachallenge/user/service/UserSolutionServiceImplTest.java
FerGimenezRoglia: Completely agree.
I'm also going to add an additional test to verify that a Flux.empty() is returned when the user has no associated solutions. Thanks for the suggestion.

ivilarop (approved):

PR #946, opened 2025-07-04
feat(user): add POST /users/create endpoint with ADMIN validation [2/3] (#557)

acocinas:
# ✅ Summary
This Pull Request implements the functionality required in Task #557, which allows a user with the ADMIN role to create new users in the system.

The POST /itachallenge/api/v1/admin/users/create endpoint has been implemented, along with all the necessary business logic and tests to ensure its correct functionality.

## 🛠️ Implemented Changes
New Controller: Created AdminCreateUserController to handle the specific logic for this endpoint, keeping it isolated.

New Service: Implemented AdminCreateUserService to encapsulate the business logic, including the validation that the username does not already exist.

Specific DTOs: Created AdminCreateUserRequestDto and AdminCreateUserResponseDto to provide clarity and a specific context for this feature.

JWT Service: Introduced IJwtService and its implementation JwtServiceImpl to handle JWT role extraction locally.

Exception Handling: Added the UnauthorizedException and updated the UserGlobalExceptionHandler to manage the new error cases (401, 403, and 409).

Test Coverage: Added robust unit tests for AdminCreateUserController, AdminCreateUserService, and JwtServiceImpl, covering success, error, and authorization cases to meet SonarQube requirements.

## justifies the technical approach
During development, the initial approach of testing the controller with integration tests (@WebFluxTest) proved to be unstable. Persistent errors (IllegalStateException) related to the creation of the Spring Test Context were encountered, where the configuration of one test case interfered with another.

To resolve these issues and ensure reliability, the following testing strategy has been adopted:

Unit Tests for the Controller: A pure unit test (AdminCreateUserControllerTest) has been chosen to test the controller's internal logic (service calls, role handling, etc.) in an isolated and stable manner, without loading the complex Spring web layer.

Unit Tests for the Service: A specific test (AdminCreateUserServiceTest) has been created to validate the business logic (interaction with the repository, user creation, etc.) in isolation.

This approach ensures that each component is tested correctly and eliminates the instability we encountered, resulting in a more robust and maintainable codebase.

## ⚠️ Technical Debt & Justification
Justification
The decision to separate tests was made after encountering recurrent and unstable IllegalStateException errors with the initial WebTest approach, ensuring the delivery of a reliable feature.

Technical Debt
Current Implementation: The validation of the JWT role is performed locally by JwtServiceImpl within the itachallenge-user microservice itself.

Future Action: The responsibility for validating tokens should belong exclusively to the itachallenge-auth microservice. A new endpoint should be created in auth for this purpose, and JwtServiceImpl should be refactored to call that new endpoint instead of decoding the token locally.

Reasoning: This temporary solution was adopted to deliver the core functionality of this PR without being blocked by the absence of said endpoint in the auth service. It is recommended to create a new backlog task to address this technical debt.

## 📋 PR Author Self-check
✅ The logic is isolated in its own Controller and Service.
✅ All tests related to this PR are now passing correctly.
✅ The changes are organized into logical commits.
✅ The technical debt is clearly documented.
✅ This is PR 2/3 for user story #553.


itachallenge-user/src/main/java/com/itachallenge/user/controller/UserCreationController.java
```diff
@@ -0,0 +1,46 @@
...
+            @RequestHeader("Authorization") String token,
+            @RequestBody CreateUserRequest request) {
+
+        return jwtServiceProxy.extractRole(token)
+                .flatMap(role -> {
+                    if ("ADMIN".equals(role)) {
+                        return userService.createUser(request)
+                                .map(userResponse -> new ResponseEntity<>(userResponse, HttpStatus.CREATED));
+                    } else {
+                        // FIX: Explicitly specify the generic type <UserResponse> for ResponseEntity
```
isaac-diez: According to the PR checklist there shouldn't be any commented code.

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserCreationController.java
```diff
@@ -0,0 +1,46 @@
...
+                        return Mono.just(new ResponseEntity<UserResponse>(HttpStatus.FORBIDDEN));
+                    }
+                })
+                // FIX: Also explicitly specify the generic type <UserResponse> for ResponseEntity
```
isaac-diez: Same thing about commented code.

itachallenge-user/src/test/java/com/itachallenge/user/controller/UserCreationControllerTest.java
```diff
@@ -0,0 +1,89 @@
...
+                .uuid(UUID.randomUUID().toString())
+                .username(username)
+                .role("USER")
+                .build();
+
+        when(jwtServiceProxy.extractRole(token)).thenReturn(Mono.just(role));
+        when(userService.createUser(any(CreateUserRequest.class))).thenReturn(Mono.just(expectedResponse));
+
+        webTestClient.post()
+                .uri("/itachallenge/api/v1/user/users/create") // URI remains the same if UserCreationController uses it
```
isaac-diez: Commented code

itachallenge-user/src/test/java/com/itachallenge/user/service/UserServiceCreateUserEmptyTest.java
```diff
@@ -0,0 +1,21 @@
...
+    private final UserServiceImpl userService = new UserServiceImpl(null);
+
+    @Test
+    @DisplayName("createUser() returns empty Mono when not implemented")
+    void createUser_returnsEmptyMono() {
+        CreateUserRequest request = new CreateUserRequest();
+        request.setUsername("dummyUser");
+
+        StepVerifier.create(userService.createUser(request))
+                .verifyComplete(); // Verifica que no emite nada y se completa correctamente
```
isaac-diez: Commented code

isaac-diez (commented):
I find everything ok. The tests run without errors. What I've only seen is some commented code and according to the wiki guide in Taiga there shouldn't be any.

acocinas:
@isaac-diez Thank you for the feedback! I hadn’t realized it before, as I was using the comments as notes to keep track of the changes I was making. I’ve now removed them to keep everything clean and correct.

isaac-diez (approved):
Thanks for removing the commented code! All OK!!

itachallenge-user/src/test/java/com/itachallenge/user/controller/UserCreationControllerTest.java
acocinas: Done

itachallenge-user/src/test/java/com/itachallenge/user/service/UserServiceCreateUserEmptyTest.java
acocinas: Done

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserCreationController.java
acocinas: Done

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserCreationController.java
acocinas: Done

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
ivilarop, line 1: Minor formatting change only — consider removing from the PR to avoid noise in history.

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserCreationController.java
ivilarop, line 1: ❗ Small inconsistency: the PR description says logic was added to UserController, but this endpoint is implemented in a new class: UserCreationController. Please clarify this in the "Implemented changes" section, and ideally justify why a new controller was introduced (e.g., to isolate ADMIN-only functionality). This helps reviewers understand the architectural decision and keeps the changelog accurate.

Can we give it another name that better reflects the objective of this class?

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserCreationController.java
```diff
@@ -0,0 +1,44 @@
...
+    public Mono<ResponseEntity<UserResponse>> createUser(
+            @RequestHeader("Authorization") String token,
+            @RequestBody CreateUserRequest request) {
+
+        return jwtServiceProxy.extractRole(token)
+                .flatMap(role -> {
+                    if ("ADMIN".equals(role)) {
+                        return userService.createUser(request)
+                                .map(userResponse -> new ResponseEntity<>(userResponse, HttpStatus.CREATED));
+                    } else {
```
ivilarop: The else here is redundant. Since the if statement has a return, the following block could be left without the else statement and would be cleaner and more readable

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserCreationController.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    @PostMapping("/users/create")
+    public Mono<ResponseEntity<UserResponse>> createUser(
+            @RequestHeader("Authorization") String token,
+            @RequestBody CreateUserRequest request) {
+
+        return jwtServiceProxy.extractRole(token)
+                .flatMap(role -> {
+                    if ("ADMIN".equals(role)) {
```
ivilarop: Avoid magic strings like "ADMIN", "Authorization". Define constants

itachallenge-user/src/main/java/com/itachallenge/user/proxy/IJwtServiceProxy.java
ivilarop, line 1: ❓ Could you please justify the use of the IJwtServiceProxy interface?
It would be helpful to clarify:

What is the objective of this interface?

How is it expected to work?

Which mechanism implements it (e.g., Feign client, WebClient proxy, manual implementation)?

Without this context, it's hard to understand how the token role extraction is being resolved at runtime, and it’s unclear whether the interface is actually wired.

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserCreationController.java
```diff
@@ -0,0 +1,44 @@
...
+                        return userService.createUser(request)
+                                .map(userResponse -> new ResponseEntity<>(userResponse, HttpStatus.CREATED));
+                    } else {
+                        return Mono.just(new ResponseEntity<UserResponse>(HttpStatus.FORBIDDEN));
+                    }
+                })
+                .defaultIfEmpty(new ResponseEntity<>(HttpStatus.UNAUTHORIZED));
```
ivilarop: 💡 Instead of handling an empty response with .defaultIfEmpty(...), consider having extractRole() return Mono.error(...) when the token is invalid, so that the GlobalExceptionHandler can return the proper 401 Unauthorized response. This would simplify the controller and centralize error handling where it belongs.

itachallenge-user/src/main/java/com/itachallenge/user/dto/CreateUserRequest.java
ivilarop, line 1: 💡 Naming suggestion: since this creation endpoint is for ADMIN use only, consider renaming CreateUserRequest to AdminCreateUserRequest for clarity.

itachallenge-user/src/main/java/com/itachallenge/user/dto/UserResponse.java
ivilarop, line 1: ❗ The current UserResponse class name is too generic and may cause ambiguity as the system evolves (especially if we introduce other user views or DTOs).
Since this response is specific to the admin-only creation flow, consider renaming it to something more explicit like AdminUserCreatedResponse or AdminCreateUserResponse.

itachallenge-user/src/main/java/com/itachallenge/user/exception/UsernameAlreadyExistsException.java
ivilarop, line 1: ❗ The UsernameAlreadyExistsException looks like a reasonable custom business exception, but it's not mentioned or justified in the PR description.
Could you briefly explain why a new exception was introduced instead of reusing an existing one (like ConflictException or BadRequestException)? That helps reviewers understand the decision and keep the exception landscape consistent.

itachallenge-user/src/main/java/com/itachallenge/user/service/UserServiceImpl.java
```diff
@@ -16,6 +18,8 @@
 public class UserServiceImpl implements UserService {
 
     private final UserRepository userRepository;
+    private static final String USER_NOT_FOUND = "User not found";
```
ivilarop: ℹ️ Se observó que se introdujo y aplicó la constante USER_NOT_FOUND para reemplazar los mensajes codificados en UserServiceImpl. Si bien es una buena práctica, no se menciona en la descripción de la solicitud de modificación.
Sería útil mencionar brevemente estas pequeñas refactorizaciones para que sean visibles en el registro de cambios y en el contexto de las revisiones.

itachallenge-user/src/main/java/com/itachallenge/user/service/UserServiceImpl.java
```diff
@@ -164,15 +168,19 @@ public Mono<Set<UUID>> getUserBookmarks(String userId) {
...
     @Override
     public Mono<UserDocument> getUserById(String userId) {
         return userRepository.findById(UUID.fromString(userId))
-                .switchIfEmpty(Mono.error(new NotFoundException("User not found")));
+                .switchIfEmpty(Mono.error(new NotFoundException(USER_NOT_FOUND)));
+    }
+
+    @Override
+    public Mono<UserResponse> createUser(CreateUserRequest request) {
+        return Mono.empty();
```
ivilarop: The implementation of createUser() in UserServiceImpl currently returns Mono.empty() with no logic.
This contradicts the description of the create request, which indicates that user creation logic has been implemented. If this is a placeholder, explicitly indicate this with a comment or a TODO statement, and clarify in the create request description that this logic will be completed in a later step.

itachallenge-user/src/main/java/com/itachallenge/user/service/UserServiceImpl.java
ivilarop, line 1: 💡 Since a new controller (UserCreationController) was introduced to isolate admin-only logic, should the user creation logic also be placed in a dedicated service (e.g., AdminUserService)?
Right now, the existing UserServiceImpl mixes logic for both the user themself (favorites/bookmarks) and admin-only user creation. Separating these responsibilities could improve cohesion and long-term maintainability.

ivilarop (changes requested):

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserController.java
acocinas: Formatting fix reverted — only one constant kept to avoid noise in PR history. Thanks!

itachallenge-user/src/main/java/com/itachallenge/user/service/UserServiceImpl.java
acocinas: Good point. We'll clarify this in the PR description to avoid adding in-code comments. The logic will be implemented in the next step.

itachallenge-user/src/main/java/com/itachallenge/user/service/UserServiceImpl.java
acocinas: Thanks for pointing it out. The USER_NOT_FOUND constant was introduced as a small refactor to remove magic strings. We'll clarify it directly in the PR description for visibility. I will explain in the PR

itachallenge-user/src/main/java/com/itachallenge/user/service/UserServiceImpl.java
acocinas: Thanks for the suggestion regarding extracting an AdminUserService.
While we agree with the idea in principle — isolating admin-only logic improves cohesion — in this case, the method added to UserServiceImpl is minimal, non-invasive, and doesn't violate SRP at this stage.

I added only one method (createUser) that is self-contained, doesn't interfere with existing logic, and follows the same contract already defined by the interface.

Introducing a new service would increase complexity and indirection without clear immediate benefit. If this method grows in logic or side effects in the future, we’ll definitely refactor it accordingly.

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserCreationController.java
acocinas: Good catch! We'll update the PR description to reflect that the new endpoint was added in AdminUserCreationController, not UserController.

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserCreationController.java
acocinas: You're right — the else was removed to improve readability, as the preceding if block already returns.

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserCreationController.java
acocinas: Good catch — constants were added to replace the magic strings.

itachallenge-user/src/main/java/com/itachallenge/user/proxy/IJwtServiceProxy.java
acocinas: Thanks for the great question!

`IJwtServiceProxy` is a declarative HTTP client used to interact with the authentication microservice. It defines a POST request to the `/extract-role` endpoint to retrieve the user’s role from the JWT.

Although there’s no concrete implementation class, this interface is wired at runtime — likely using Spring Cloud OpenFeign or a similar mechanism — which autogenerates the implementation based on the annotations.

This approach simplifies inter-service communication, improves testability, and adheres to separation of concerns.

We'll clarify this in the PR description.

itachallenge-user/src/main/java/com/itachallenge/user/dto/UserResponse.java
acocinas: Thanks for the suggestion — the DTO was renamed to AdminUserCreatedResponseDto to better reflect its purpose within the admin-only user creation flow.

itachallenge-user/src/main/java/com/itachallenge/user/exception/UsernameAlreadyExistsException.java
acocinas: You're right — we’ll clarify this in the PR description. We introduced UsernameAlreadyExistsException as a custom business exception to make the conflict case more explicit and semantically clear, rather than overloading a generic exception like ConflictException. Thanks for pointing it out!

itachallenge-user/src/main/java/com/itachallenge/user/controller/UserCreationController.java
acocinas: Done

itachallenge-user/src/main/java/com/itachallenge/user/dto/CreateUserRequest.java
acocinas: Done

isaac-diez:
Ivan raised some good points that Alfonso addressed accordingly. Now the PR description and the code are clearer and easier to understand and maintain. Good work!

itachallenge-user/src/main/java/com/itachallenge/user/dto/AdminUserCreatedResponseDto.java
ivilarop, line 1: better AdminCreateUserResponseDto no?

itachallenge-user/src/main/java/com/itachallenge/user/service/UserServiceImpl.java
ivilarop: Perfect, I think it's a very thoughtful answer. It's fine to keep it that way if the method is small, and to mark it as a point of focus for future developments. Thanks for taking this into account.

itachallenge-user/src/main/java/com/itachallenge/user/service/UserServiceImpl.java
ivilarop: ok thks

itachallenge-user/src/main/java/com/itachallenge/user/service/UserServiceImpl.java
ivilarop: ok thks

itachallenge-user/src/main/java/com/itachallenge/user/controller/AdminUserCreationController.java
```diff
@@ -0,0 +1,48 @@
...
+        this.jwtServiceProxy = jwtServiceProxy;
+    }
+
+    @PostMapping("/users/create")
+    public Mono<ResponseEntity<AdminUserCreatedResponseDto>> createUser(
+            @RequestHeader(AUTHORIZATION_HEADER) String token,
+            @RequestBody AdminCreateUserRequestDto request) {
+
+        return jwtServiceProxy.extractRole(token)
+                .switchIfEmpty(Mono.error(new UnauthorizedException("Token is invalid or missing")))
```
ivilarop: 👀 Good job using .switchIfEmpty(...) to cover the case where extractRole(token) returns a Mono.empty().

🔄 However, for complete error flow management, you should also add an .onErrorResume(...), as:

.switchIfEmpty(...) covers the case where the auth microservice responds with a 200 OK response but doesn't return the role (an empty body).

.onErrorResume(...) catches when the auth microservice throws a real error (e.g., 401 Unauthorized or 403 Forbidden), which arrives as a WebClientResponseException.

This will ensure you cover both scenarios correctly and prevent unhandled exceptions from propagating to the frontend.

itachallenge-user/src/main/java/com/itachallenge/user/proxy/IJwtServiceProxy.java
```diff
@@ -0,0 +1,13 @@
...
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestHeader;
+import org.springframework.web.bind.annotation.RequestMapping;
+import reactor.core.publisher.Mono;
+
+@RequestMapping("/itachallenge/api/v1/auth/internal/jwt")
+public interface IJwtServiceProxy {
+
+    @PostMapping("/extract-role")
+    Mono<String> extractRole(@RequestHeader("Authorization") String authHeader);
```
ivilarop: 👓 The solution you propose using IJwtServiceProxy with a declarative interface is completely valid and modern, especially in environments with Spring WebFlux.

🔄 However, we are currently using a manually configured WebClient in the challenge microservice and other aspects of the project. For architectural consistency, it would be advisable to comprehensively assess which approach we want to standardize across the project.

✅ If after analyzing it, we consider IJwtServiceProxy (or declarative proxies in general) to be the approach we want to follow, it would be advisable to record this decision as technical debt so that in future sprints we can apply this pattern uniformly across all microservices.

Thank you for providing a clean and well-structured alternative — these types of decisions are what improve the quality of the project in the medium term.

itachallenge-user/src/main/java/com/itachallenge/user/controller/AdminUserCreationController.java
```diff
@@ -0,0 +1,48 @@
...
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestHeader;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+import reactor.core.publisher.Mono;
+
+@RestController
+@RequestMapping(value = "/itachallenge/api/v1/user")
```
ivilarop: For a more consistent and semantic API, you could adjust the controller's base path to /itachallenge/api/v1/admin, since all endpoints in this class are intended for administrator use only. This would leave the final path as /admin/users/create, which is clearer and more extensible for future administrative functionality.

ivilarop (changes requested):

itachallenge-user/src/main/java/com/itachallenge/user/dto/AdminUserCreatedResponseDto.java
acocinas: That's a great point. I had the same thought and already refactored it to AdminCreateUserResponseDto to make its purpose clearer. Thanks for the feedback!

itachallenge-user/src/main/java/com/itachallenge/user/controller/AdminUserCreationController.java
acocinas: Thanks for the detailed feedback! That's a great point regarding onErrorResume.

In a previous iteration of this code, I was indeed using a WebClient proxy, and your suggestion would have been spot-on. However, that approach was causing some deep-seated instability, so as discussed in one of our daily meetings, we decided to change the approach and register the corresponding technical debt.

The current implementation (JwtServiceImpl) now handles token validation locally. This new service was created following the existing pattern from the itachallenge-challenge microservice. It's wrapped in a try-catch block that returns a Mono.error(new UnauthorizedException(...)) for any parsing or validation failure. This way, all possible error scenarios are centralized into a single exception type, which is then handled by the UserGlobalExceptionHandler.

As you mentioned, and as noted in the PR description, this local validation is considered technical debt. The long-term plan is to create a proper validation endpoint in the auth microservice and refactor this service to call it.

Thanks again for the sharp observation!

itachallenge-user/src/main/java/com/itachallenge/user/controller/AdminUserCreationController.java
acocinas: That's an excellent suggestion for API consistency. I've implemented this change and updated the @RequestMapping to /itachallenge/api/v1/admin. The frontend team has been notified to ensure alignment. Thanks for the feedback!

itachallenge-user/src/main/java/com/itachallenge/user/proxy/IJwtServiceProxy.java
acocinas: Thanks for this excellent architectural feedback. I completely agree that standardizing our approach to inter-service communication is crucial for the project's health.

You are correct, the initial implementation did use a declarative proxy (IJwtServiceProxy). However, during development and testing, this approach led to critical instability issues that were blocking the PR.

As a pragmatic solution to move forward, we pivoted to the current local JwtServiceImpl, following the pattern already present in the itachallenge-challenge microservice. We have documented this decision and the local validation as technical debt in the PR's description.

The long-term plan is completely aligned with your thinking. Once the itachallenge-auth microservice exposes a dedicated endpoint for token validation, this implementation will be refactored to make a standard client call to it. The discussion about standardizing on a declarative vs. manual WebClient will be very valuable for the team when we tackle that task.

Thanks again for the insightful comment!

PR #950, opened 2025-07-16
feat(jwt-core): create jwt-core module to unify token management 1/3 (#592)

isaac-diez:
### Summary

This PR introduces the centralized JWT logic shared by auth, challenge, and user microservices into a new reusable module jwt-core to eliminate:

- Token validation
- UUID extraction from the Authorization header
- Claims extraction
- Exception handling for expired or tampered tokens

This PR is the first part of 3 which are derived from the following tasks:
- [#592](https://tree.taiga.io/project/luisvi-ita-challenges/task/592) 1/3 Create a centralized JWT module to unify token management across all services
- [#593](https://tree.taiga.io/project/luisvi-ita-challenges/task/593) 2/3 Refactor Auth Service to use new jwt-core module
- [#594](https://tree.taiga.io/project/luisvi-ita-challenges/task/594) 3/3 Refactor Challenge Service to use new jwt-core module

### Changes
Created a new module jwt-core that contains: 

**build.gradle & settings.gradle**
Needed settings for module to load plugins, dependencies, etc. and interact with the other modules.

**IJwtService & JwtService**
Created empty into the new module. Next task [#593](https://tree.taiga.io/project/luisvi-ita-challenges/task/593) will take care of the refactoring of Auth and implementation of IJwtService in jwt-core accordingly.

**JwtServiceTest & JwtServiceUuidExtractionTest**
Created empty for sonarquebe to pass. Next task [#593](https://tree.taiga.io/project/luisvi-ita-challenges/task/593) will take care of new tests following the new module structure for the Challenge microservice needs.

**CHANGELOG.md**
Added version 1.0.0 of jwt-core module

🚨 No .env.CI,dev changed since it is not a runnable microservice.

📌 Documentation (README.md) for the jwt-core module will be included in PR 2/3 (#593) to avoid blocking this one.

## 📋 PR Author Self-check
✅ PR is focused and isolated
✅ Branch follows naming conventions: feat(jwt-core): create jwt-core module to unify token management 1/3 (#592).
✅ No debug logs left.
✅ Exception mapped correctly.
✅ PR clearly documents scope and purpose.

Issue #593, opened 2024-09-26
User micro refactoring


jonatanvicente:
At User Micro refactoring Service Layer must contains only UserService (maybe with Interface).

PR #952, opened 2025-07-17
feat(jwt-core & auth): refactor auth microservice to use jwt-core module 2/3 (#593)

isaac-diez:
### Summary

This PR continues the work started in PR #950 (task [#592](https://tree.taiga.io/project/luisvi-ita-challenges/task/592) 1/3 Create a centralized JWT module to unify token management across all services) but now refactores (takes out) IJwtService and JwtService in Auth microservice to make use of the new jwt-core module created to such effect.

This PR is the second part of 3 which are derived from the following tasks:
- [#592](https://tree.taiga.io/project/luisvi-ita-challenges/task/592) 1/3 Create a centralized JWT module to unify token management across all services
- [#593](https://tree.taiga.io/project/luisvi-ita-challenges/task/593) 2/3 Refactor Auth Service to use new jwt-core module
- [#594](https://tree.taiga.io/project/luisvi-ita-challenges/task/594) 3/3 Refactor Challenge Service to use new jwt-core module

### Changes

**App** in auth
A ComponentScan annotation is needed in microservices to detect module beans, since jwt-core is not a Spring Boot app.

**IJwtService, JwtService and JwtServiceTest** in auth
Deleted. They are not longer needed. Now the logic goes through jwt-core.

**IJwtService and JwtService** in jwt-core
Moved all aforementioned logic into jwt-core except for the role switching.

**AuthJwtFacade()** in auth
A new "fake" service layer that acts as a facade for the real service in jwt-core. This decouples the controller completely from whichever library is used for the logic of the token. 

**AuthController** in auth
Modified to use the new service JwtRoleSwitchService()

**JwtRoleSwitchService()** in auth
Created the service because it was inside (coupled) JwtService in auth and should stay there. This logic doesn't fit in jwt-core that only takes care of token creation, validation, UUID and Claims extraction and exception handling.

**buil.gradle** in auth
added jwt-core dependency

**JwtServiceTest & JwtServiceUuidExtractionTest** in jwt-core
Created all tests to check the right implementation of the logic.

**AuthControllerTest** in auth
Modified to check the logic with the new JwtRoleSwitchService() and the AuthJwtFacade() "fake" auth service layer.

**JwtRoleSwitchServiceTest** in auth
Created from scratch to test the service

### Postman test:
Successfully retrieved and checked the token:

image

image

🚨 Note:
Initially, the plan was to have the structure of the new module jwt-core in the PR 1/3 but sonarqube wouldn't green light the PR due to duplication of code. That's why, in the end, in the PR 1/3, the jwt-core module only has the class structure but empty of code.
That's why this PR has more files modified because I had to move the service layer from Auth to jwt-core in the same PR.

## 📋 PR Author Self-check
✅ PR is focused and isolated
✅ Branch follows naming conventions: feat(jwt-core & auth): refactor auth microservice to use jwt-core module 2/3 (#593).
✅ All logic tested locally.
✅ No debug logs left.
✅ Exception mapped correctly.
✅ Tests included and passing.


VaninaVega (approved):
The PR is solid and well executed. It successfully addresses the refactorization of the Auth microservice as described in subtask #593 and contributes significantly to the main task of modularization (#589).

CHANGELOG.md
```diff
@@ -3,6 +3,16 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### [itachallenge-auth-2.0.5-RELEASE] - 2025-07-22
+
+### Refactored
+- Moved all JWT logic to new jwt-core module. (NOTE: Role Switch Service remains in auth)  (Taiga [#593], PR [#952])
+
+### [itachallenge-jwtcore-1.0.0-RELEASE] - 2025-07-22
```
ivilarop: We had already incorporated this in the previous PR, right?

itachallenge-auth/src/main/java/com/itachallenge/auth/App.java
```diff
@@ -3,9 +3,11 @@
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.cloud.client.discovery.EnableDiscoveryClient;
+import org.springframework.context.annotation.ComponentScan;
 
 @SpringBootApplication
 @EnableDiscoveryClient
+@ComponentScan(basePackages = {"com.itachallenge.auth", "com.itachallenge.jwtcore"})
```
ivilarop: @ComponentScan is well-suited to include beans from the jwt-core module, which is outside the microservice's root package. However, com.itachallenge.auth is already implicitly included because it's in the same package as the App class. The basePackages could be simplified to just "com.itachallenge.jwtcore."

itachallenge-auth/src/main/java/com/itachallenge/auth/controller/AuthController.java
```diff
@@ -46,16 +47,19 @@ public class AuthController {
 
     private final String appName;
 
+    private final JwtRoleSwitchService jwtRoleSwitchService;
```
ivilarop, line 50: Great job extracting the switchRole logic into JwtRoleSwitchService — that’s definitely the right direction for isolating domain logic from technical details.

I noticed that other JWT-related methods (like generateToken, validateToken, etc.) are still being called directly from IJwtService inside the controller.

If we wanted to keep the controller decoupled from the underlying implementation, how could we make that access more consistent across all endpoints?

(Bonus question: what would you do if we ever needed to replace jwt-core with another library?)

jwt-core/README.md
ivilarop, line 1: Currently, there is no mention that a @ComponentScan is needed in microservices to detect module beans, since jwt-core is not a Spring Boot app.

Could you please include a minimal code example? It would help any new developers see how it's used in practice without having to explore the classes.

ivilarop (changes requested):

CHANGELOG.md
isaac-diez: No, in the previous PR we created the (empty) structure for the new jwt-core module. Sonarqube wouldn't let me have the service interface and the implementation in the new module and in auth due to duplicity of code.

itachallenge-auth/src/main/java/com/itachallenge/auth/App.java
isaac-diez: Thanks for the comment. My bad. I thought it was necessary to state all microservices in detail. I'm removing "com.itachallenge.auth" from the anotation.

isaac-diez: Thanks for the comment. After looking into this, if I take out "com.itachallenge.auth" the AuthController tests won't pass since they are a @WebFluxTest and it has a limite scope. What I suggest is creating a config class in /test that add this @ComponentScan(basePackages = "com.itachallenge.auth") and the tests pass, the build is successful and the microservice can be run.

jwt-core/README.md
isaac-diez: Sorry about that, I thought I had it in. It is now.

acocinas (approved):
✅ Overall Review
This new version of the PR is excellent and a significant improvement over the previous one. The main goal of refactoring the auth microservice to use jwt-core has been met, but more importantly, it has been done in a much more robust and decoupled manner.

The introduction of the Facade pattern (AuthJwtFacade) is a strong architectural choice. It effectively decouples the controller from the jwt-core implementation, resulting in a more modular and maintainable design.

👍 Highlights (Positive Points)
Facade Pattern (AuthJwtFacade): This is the key improvement. By creating AuthJwtFacade, you have completely decoupled the AuthController from the specific implementation of jwt-core. The controller now only knows about the facade, which in turn delegates to the module's service. This perfectly answers the senior's question: "(what would you do if we had to replace jwt-core?)". With this design, you would only need to change the implementation inside AuthJwtFacade without touching the controller. 👏

Separation of Concerns: The logic remains well-separated. JwtRoleSwitchService handles the business rules for role changes, while AuthJwtFacade handles the technical interaction with tokens. It's a very clean design.

Adapted Tests: The tests (AuthControllerTest, JwtRoleSwitchServiceTest) have been correctly updated to use the new facade and the corresponding mocks. This ensures the new abstraction layer is well-tested.

Correct Configuration: The @ComponentScan annotation has been added to the App class to ensure Spring detects the beans from the jwt-core module, which is a crucial and correct technical detail.

⚠️ Acknowledged Security Fix
We are aware that the previous version of this PR included a critical security issue regarding the manual parsing of JWT claims (extractAllClaimsMap method).

Thank you for acknowledging this feedback. We understand that this security fix has already been implemented in the next PR (3/3) of this series. This is a great example of iterative improvement and ensures the final solution is secure.

🏁 Conclusion
Approve. 👍

This is a high-quality PR with an excellent architecture. The introduction of the Facade pattern is a standout improvement. Well done.

itachallenge-auth/src/main/java/com/itachallenge/auth/controller/AuthController.java
isaac-diez: I see what you mean and yes, it's good to have the controlle completely detached from the implementation. For this, i have created a "fake" or facade service layer in auth called AuthJwtFacade(). The controller and JwtRoleSwitchService call this fake service layer.
I had to adapt both tests as well.

And for your bonus question, now simply is a matter of changing the implementation inside AuthJwtFacade and the rest of the microservice remains unchanged. It is more convenient since it is a single point of coupling with JWT.

CHANGELOG.md
isaac-diez: No, in the previous PR we created the (empty) structure for the new jwt-core module. Sonarqube wouldn't let me have the service interface and the implementation in the new module and in auth due to duplicity of code.

itachallenge-auth/src/main/java/com/itachallenge/auth/service/AuthJwtFacade.java
ivilarop, line 1: Excellent work introducing AuthJwtFacade as a single entry point for JWT-related operations in the authentication microservice. This creates a clear separation from the jwt-core module.

A possible improvement: Would it make sense to extract an interface for this facade (e.g., IAuthJwtFacade) and have the controller depend on it? This way, the controller remains completely decoupled from the implementation and also makes it easier to simulate in testing or replace in the future.

For example, if we ever decide to replace jwt-core with another library or third-party provider, only the facade implementation would need to be changed, not the controller or other consumers.

You're very close to achieving a truly clean and testable design. 👏

itachallenge-auth/src/main/java/com/itachallenge/auth/controller/AuthController.java
ivilarop: Ok thks! I need to review the comment I made to you on the new service AuthJwtFacade

itachallenge-auth/src/main/java/com/itachallenge/auth/service/JwtRoleSwitchService.java
ivilarop, line 1: there's no strong need to extract an interface for this class at this point.

Since it doesn't depend directly on jwt-core but instead uses AuthJwtFacade, it's already isolated from external implementation details. The controller depends on it as a use case class, not as a low-level adapter.

Creating an interface would only be necessary if:

You expect multiple implementations of this service.

You want to mock it in controller tests.

Or you're following a strict hexagonal architecture where every use case is behind an interface.

For now, keeping it as a concrete class is perfectly fine 👌

CHANGELOG.md
ivilarop: Leave a comment on how the final solution turned out please

itachallenge-auth/src/main/java/com/itachallenge/auth/App.java
ivilarop: reat insight and solution — you're absolutely right about the limited scope of @WebFluxTest.

Extracting the component scan for "com.itachallenge.auth" into a test-specific config class is a clean and modular approach. Thanks for checking that in detail!

jwt-core/README.md
ivilarop: ok thks!

ivilarop (changes requested):

itachallenge-auth/src/main/java/com/itachallenge/auth/service/AuthJwtFacade.java
isaac-diez: Thanks for the comment. It makes total sense to extract an interface for the AuthJwtFacade and fully decouple the controller. I've created and pushed it.

isaac-diez (commented):
Thanks for the comments. I've resolved the OK  ones and added the IAuthJwtFacade interface as mentioned.

ivilarop (approved):

PR #953, opened 2025-07-21
Feat(user): add POST /users/create endpoint without validation (#618)

Heyvkram:
This pull request modifies the previous one (https://github.com/IT-Academy-BCN/ita-challenges-backend/pull/946), by removing the ADMIN role requirement. Previously, the POST /itachallenge/api/v1/admin/users/create route required an authenticated token with admin privileges. This validation has been removed to allow any client to register users.

Changes Made
AdminCreateUserController.java:

- Removed the dependency on IJwtService.
 
- Removed all logic related to reading the Authorization header and validating the user's role.
 
- The controller method now directly invokes AdminCreateUserService without any permission checks.

- The class names (Admin...) and endpoint path (.../admin/...) are deliberately kept to minimize the impact of this temporary patch and to facilitate future reversion or refactoring.

✅ Manual Testing (Postman)
The endpoint has been tested manually and is working correctly.

Method: POST

URL: http://localhost:8764/itachallenge/api/v1/admin/users/create

Body:
Select raw and JSON.

JSON
{
    "username": "testOne"
}

Expected Result:

Status: 201 Created

Body: A JSON object containing the new user's uuid, username, and role.

JSON

{
    "uuid": "some-unique-uuid-string",
    "username": "testuser_en",
}

PR #618, opened 2024-10-28
Feature#533version3

MarineAvramova:
Implemented a new GET endpoint over HTTP protocol:
/itachallenge/api/v1/user/{idUser}/challenge/{idChallenge}/solution/{idSolution}/score

PR #956, opened 2025-07-28
fix(user) add post users/create without validation (#618)

Heyvkram:
added back "roll' to the AdminCrateUserService and to the AdminCreateUserResponseDto due to create the user with USER roll assigned.

Fix of PR #953


acocinas (approved):
I think is better with a changelog X.X.+1 and .env.CI

ivilarop (approved):

PR #651, opened 2024-11-11
Feature#642

auguNa:
This PR introduces a complete Docker setup for the itachallenge-score microservice, enabling containerized execution and deployment. It includes the creation of a custom Docker image, which has been successfully tested and pushed to DockerHub for easy accessibility.
**Key Changes:**
Added a Dockerfile to build the container for the itachallenge-score microservice.
Created a docker-compose.dev.yml file for local development and testing.
Configured the container to run a sandbox environment for secure code execution.
Implemented a shell script (compile_and_run.sh) to compile and run Java user code within the container.
Uploaded the Docker image to DockerHub under the repository: augustin19/itachallenge-score-sandbox:latest.

PR #959, opened 2025-09-11
feat(github-core): Create GitHub Username Validation in Shared Module (#649)

isaac-diez:
### Summary

This PR introduces a centralized Github Username Validation that will be shared with User and Auth microservices. 

Main tasks:
Create new github-core module that will centralize all github user validation logic:
    - service
    - configuration
    - exceptions
    - test

This PR is the first part of 3 which are derived from the following tasks:
- [#649](https://tree.taiga.io/project/luisvi-ita-challenges/task/649) 1/3 Create GitHub Username Validation in Shared Module
- [#650](https://tree.taiga.io/project/luisvi-ita-challenges/task/650) 2/3 Integrate Validation in User Creation Endpoint
- [#651](https://tree.taiga.io/project/luisvi-ita-challenges/task/651) 3/3 Refactor & Reuse in Auth-Service

### Changes
Created a new module github-core that contains: 

**GithubServiceConfig**
This class tells Spring how to create a GithubApiService bean that other parts of the app can use. It reads the GitHub API URL from the configuration (application.yml) and passes it, along with a WebClient builder, to the service so it can make requests to GitHub.

**GithubApiService & GithubApiServiceImpl**
Interface and implementation that take care of the mail logic of validating the username with Github. Returns true if server returns a 200 cods or false if the server returns a 4xx or 5xx code.

**GithubUnavailableException**
This exceptions are thrown when the API itself is down (code 5xx) and when the user is not found (code 4xx), respectively.

**GithubApiServiceTest**
Created to test the logic. It tests three situations: the user is not in the repo and is a github user, the user is already in the repo and the user is not a github user.

**settings.gradle & build.gradle**
Needed settings for module to load plugins, dependencies, etc. and interact with the other modules.

**CHANGELOG.md**
Added version 1.0.0 of github-core module

🚨 No .env.CI,dev changed since it is not a runnable microservice.

📌 Documentation (README.md) for the github-core module will be included in PR 2/3 (#593) to avoid blocking this one.

**Results on Postman:**
Screenshot-2025-09-16-17:38:27

## 📋 PR Author Self-check
✅ PR is focused and isolated
✅ Branch follows naming conventions.
✅ No debug logs left.
✅ Exception mapped correctly.

### DIFF
--- a/github-core/src/main/java/com/itachallenge/githubcore/exception/GithubApiException.java
+++ b/github-core/src/main/java/com/itachallenge/githubcore/exception/GithubApiException.java
@@ -1,0 +1,7 @@
+package com.itachallenge.githubcore.exception;
+
+public class GithubApiException extends RuntimeException {
+    public GithubApiException(String message) {
+        super(message);
+    }
+}
--- a/github-core/src/main/java/com/itachallenge/githubcore/exception/GithubApiException.java
+++ b/github-core/src/main/java/com/itachallenge/githubcore/exception/GithubApiException.java
@@ -1,7 +1,0 @@
-package com.itachallenge.githubcore.exception;
-
-public class GithubApiException extends RuntimeException {
-    public GithubApiException(String message) {
-        super(message);
-    }
-}
--- a/github-core/src/main/java/com/itachallenge/githubcore/service/GithubApiService.java
+++ b/github-core/src/main/java/com/itachallenge/githubcore/service/GithubApiService.java
@@ -1,5 +1,7 @@
 package com.itachallenge.githubcore.service;
 
+import com.itachallenge.githubcore.exception.GithubApiException;
+import com.itachallenge.githubcore.exception.GithubUserNotFoundException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.http.HttpStatusCode;
@@ -23,11 +25,11 @@ public class GithubApiService implements IGithubApiService{
                 .retrieve()
                 .onStatus(HttpStatusCode::is4xxClientError, r -> {
                     log.info("GitHub user not found: {}", username);
-                    return Mono.error(new RuntimeException("User not found"));
+                    return Mono.error(new GithubUserNotFoundException(username));
                 })
                 .onStatus(HttpStatusCode::is5xxServerError, r -> {
                     log.info("GitHub API error: {}", username);
-                    return Mono.error(new RuntimeException("GitHub API error"));
+                    return Mono.error(new GithubApiException("GitHub API error"));
                 }
                 )
                 .toBodilessEntity()
--- a/github-core/src/main/java/com/itachallenge/githubcore/service/GithubApiService.java
+++ b/github-core/src/main/java/com/itachallenge/githubcore/service/GithubApiService.java
@@ -1,0 +1,37 @@
+package com.itachallenge.githubcore.service;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.http.HttpStatusCode;
+import org.springframework.web.reactive.function.client.WebClient;
+import reactor.core.publisher.Mono;
+
+
+public class GithubApiService implements IGithubApiService{
+
+    private static final Logger log = LoggerFactory.getLogger(GithubApiService.class);
+    private final WebClient webClient;
+
+    public GithubApiService(WebClient.Builder builder, String baseUrl) {
+        this.webClient = builder.baseUrl(baseUrl).build();
+    }
+
+    public Mono<Boolean> userExists(String username) {
+        return webClient.get()
+                .uri("/users/{username}", username)
+                .retrieve()
+                .onStatus(HttpStatusCode::is4xxClientError, r -> {
+                    log.info("GitHub user not found: {}", username);
+                    return Mono.error(new RuntimeException("User not found"));
+                })
+                .onStatus(HttpStatusCode::is5xxServerError, r -> {
+                    log.info("GitHub API error: {}", username);
+                    return Mono.error(new RuntimeException("GitHub API error"));
+                }
+                )
+                .toBodilessEntity()
+                .map(response -> true)
+                .defaultIfEmpty(false);
+    }
+
+}
--- a/github-core/src/main/java/com/itachallenge/githubcore/service/GithubApiService.java
+++ b/github-core/src/main/java/com/itachallenge/githubcore/service/GithubApiService.java
@@ -16,6 +16,7 @@ public class GithubApiService implements IGithubApiService{
         this.webClient = builder.baseUrl(baseUrl).build();
     }
 
+    @Override
     public Mono<Boolean> userExists(String username) {
         return webClient.get()
                 .uri("/users/{username}", username)
--- a/github-core/src/main/java/com/itachallenge/githubcore/service/GithubApiService.java
+++ b/github-core/src/main/java/com/itachallenge/githubcore/service/GithubApiService.java
@@ -1,40 +1,8 @@
 package com.itachallenge.githubcore.service;
 
-import com.itachallenge.githubcore.exception.GithubApiException;
-import com.itachallenge.githubcore.exception.GithubUserNotFoundException;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-import org.springframework.http.HttpStatusCode;
-import org.springframework.web.reactive.function.client.WebClient;
+import com.itachallenge.githubcore.document.enums.GithubUserStatus;
 import reactor.core.publisher.Mono;
 
-
-public class GithubApiService implements IGithubApiService{
-
-    private static final Logger log = LoggerFactory.getLogger(GithubApiService.class);
-    private final WebClient webClient;
-
-    public GithubApiService(WebClient.Builder builder, String baseUrl) {
-        this.webClient = builder.baseUrl(baseUrl).build();
-    }
-
-    @Override
-    public Mono<Boolean> userExists(String username) {
-        return webClient.get()
-                .uri("/users/{username}", username)
-                .retrieve()
-                .onStatus(HttpStatusCode::is4xxClientError, r -> {
-                    log.info("GitHub user not found: {}", username);
-                    return Mono.error(new GithubUserNotFoundException(username));
-                })
-                .onStatus(HttpStatusCode::is5xxServerError, r -> {
-                    log.info("GitHub API error: {}", username);
-                    return Mono.error(new GithubApiException("GitHub API error"));
-                }
-                )
-                .toBodilessEntity()
-                .map(response -> true)
-                .defaultIfEmpty(false);
-    }
-
+public interface GithubApiService {
+    Mono<GithubUserStatus> userExists(String username);
 }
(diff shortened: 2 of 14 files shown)


acocinas (approved):
The overall quality of this PR is very high. The core logic and design are excellent, and the author has followed several best practices. This is solid work. I'm providing a few minor suggestions to make the new module even more robust and maintainable. Please consider them as optional improvements for a future refactoring.

Suggestions for Improvement
Dependency Redundancy: The build.gradle file has some redundant test dependencies and specific version declarations that could lead to version conflicts in other microservices. It's a good practice to simplify this list and rely on the versions managed by the spring-boot-starter-test dependency. This would make the module's dependencies cleaner and more consistent.

Module Naming: The name of the new module, github-core, is clear, but the configuration class GithubConfig could be more specific, like GithubServiceConfig. This helps other developers understand its purpose more precisely.

These are minor points. I don't see any issues that would block this PR from being merged. The code is well-designed, well-tested, and serves its intended purpose. Great work!

Heyvkram (approved):
I agree with the feedback above. The implementation is solid and the logic is easy to follow.

The suggestions for improving dependency management and configuration naming are great points to consider for future refactoring.

No blockers from my side.

isaac-diez:
> The overall quality of this PR is very high. The core logic and design are excellent, and the author has followed several best practices. This is solid work. I'm providing a few minor suggestions to make the new module even more robust and maintainable. Please consider them as optional improvements for a future refactoring.
> 
> Suggestions for Improvement Dependency Redundancy: The build.gradle file has some redundant test dependencies and specific version declarations that could lead to version conflicts in other microservices. It's a good practice to simplify this list and rely on the versions managed by the spring-boot-starter-test dependency. This would make the module's dependencies cleaner and more consistent.
> 
> Module Naming: The name of the new module, github-core, is clear, but the configuration class GithubConfig could be more specific, like GithubServiceConfig. This helps other developers understand its purpose more precisely.
> 
> These are minor points. I don't see any issues that would block this PR from being merged. The code is well-designed, well-tested, and serves its intended purpose. Great work!

OK Alfonso, thanks for the review and for the tips. I have changed the name of the GithubConfig to GithubServiceConfig for the sake of better understanding.

ivilarop:
@isaac-diez 

⚠️ According to the PR guidelines, the title should use feat (not feature)

ivilarop:
@isaac-diez @Heyvkram @acocinas 

The PR description is confusing to me. 👉 This shouldn't be part of this module.

- Token validation
- UUID extraction from the Authorization header
- Claims extraction
- Exception handling for expired or tampered tokens

All the JWT token logic is already in jwt-core and should remain centralized there.
This PR should be limited to GitHub integration only (validating usernames and, in the case of auth-service, sharing code and OAuth profiles).

isaac-diez:
> @isaac-diez @Heyvkram @acocinas
> 
> The PR description is confusing to me. 👉 This shouldn't be part of this module.
> 
>     * Token validation
> 
>     * UUID extraction from the Authorization header
> 
>     * Claims extraction
> 
>     * Exception handling for expired or tampered tokens
> 
> 
> All the JWT token logic is already in jwt-core and should remain centralized there. This PR should be limited to GitHub integration only (validating usernames and, in the case of auth-service, sharing code and OAuth profiles).

You are right. I took and old PR summary to modify for this one and didn't do it thoroughly. It is fixed now.

ivilarop:
@isaac-diez @Heyvkram @acocinas 
I see that in the PR description you only mention that this module is only responsible for centralized GitHub Username Validation. However, what about the OAuth flow (code→access_token exchange and /user profile retrieval), which is currently in the authentication microservice? Won't this module also be responsible for it? What is the purpose of this? Can we mention it?

github-core/src/main/java/com/itachallenge/githubcore/service/GithubApiService.java
```diff
@@ -0,0 +1,40 @@
...
+    public GithubApiService(WebClient.Builder builder, String baseUrl) {
+        this.webClient = builder.baseUrl(baseUrl).build();
+    }
+
+    @Override
+    public Mono<Boolean> userExists(String username) {
+        return webClient.get()
+                .uri("/users/{username}", username)
+                .retrieve()
+                .onStatus(HttpStatusCode::is4xxClientError, r -> {
```
ivilarop: Regarding GitHub's 404 handling:

- In GET https://api.github.com/users/{username}, a 404 doesn't mean that GitHub has failed, but rather that the user doesn't exist.

👉 This is an expected result for the domain, not a technical error.

- Therefore, it's not recommended to throw a GithubUserNotFoundException in this case.

Using exceptions for a normal flow (user doesn't exist) clutters logs, makes flow control more expensive, and conflates "doesn't exist" with "something went wrong."

Instead of an exception, the GitHub client should return a value that clearly indicates the result, and the microservice will decide how to translate it.

- In user-service, this value should be transformed into a validation error for the front end (for example, returning a 400 Bad Request with the message "githubUsername: does not exist on GitHub").

That is: the library integrates with GitHub → returns "not found" as normal.

The microprocessor applies the business logic → converts the "not found" into a validation error for the client.

- Conclusion: it's not necessary to create a custom GithubUserNotFoundException exception.

 Reserve custom exceptions for real technical errors (timeouts, 5xx, invalid auth, rate limit).

github-core/src/main/java/com/itachallenge/githubcore/service/GithubApiService.java
```diff
@@ -0,0 +1,40 @@
...
+                    log.info("GitHub user not found: {}", username);
+                    return Mono.error(new GithubUserNotFoundException(username));
+                })
+                .onStatus(HttpStatusCode::is5xxServerError, r -> {
```
ivilarop: Here, you do need to throw an exception because a GitHub 5xx (or timeout) is a real technical error.
The only thing I suggest is not using a generic name like GithubApiException, but something more precise:

GithubUnavailableException → for 5xx/timeout (external service down/not responding).

github-core/src/main/java/com/itachallenge/githubcore/service/GithubApiService.java
```diff
@@ -0,0 +1,40 @@
...
+
+    private static final Logger log = LoggerFactory.getLogger(GithubApiService.class);
+    private final WebClient webClient;
+
+    public GithubApiService(WebClient.Builder builder, String baseUrl) {
+        this.webClient = builder.baseUrl(baseUrl).build();
+    }
+
+    @Override
+    public Mono<Boolean> userExists(String username) {
```
ivilarop: Right now, the method returns Mono<Boolean>, but in reality it only returns true (when GitHub responds with a 200) or throws an exception (for 404, 5xx, timeout, etc.). False is never returned, so the contract is misleading.

To be correct, the 404 shouldn't be treated as an exception, but rather as a normal result: "user not found on GitHub." That case should be returned along with the "user found" case (200).

👉 Ideally, the method shouldn't return a Boolean, but rather an expressive type (e.g., an enum with two values: FOUND and NOT_FOUND). This way, we clearly differentiate the expected results from the domain, while real technical errors (5xx, timeout, etc.) will still be propagated as exceptions.

github-core/src/main/java/com/itachallenge/githubcore/service/IGithubApiService.java
ivilarop, line 1: It's better to rename GithubApiService to follow standard nomenclature.

github-core/src/main/java/com/itachallenge/githubcore/service/GithubApiService.java
ivilarop, line 1: It's better to rename GithubApiServiceImpl to follow standard nomenclature.

github-core/src/main/java/com/itachallenge/githubcore/exception/GithubApiException.java
ivilarop, line 1: I suggest is not using a generic name like GithubApiException, but something more precise:

GithubUnavailableException → for 5xx/timeout (external service down/not responding).

ivilarop (changes requested):

github-core/src/main/java/com/itachallenge/githubcore/exception/GithubApiException.java
isaac-diez: Thanks for the suggestion. Yes, i also agree that is clearer and more precise.

github-core/src/main/java/com/itachallenge/githubcore/service/IGithubApiService.java
isaac-diez: OK, i've renamed the interface to GithubApiService.

github-core/src/main/java/com/itachallenge/githubcore/service/GithubApiService.java
isaac-diez: I have eliminated the exception altogether. It is not needed since it is normal behavior of the program and not a technical problem.

isaac-diez:
> @isaac-diez @Heyvkram @acocinas
> 
> The PR description is confusing to me. 👉 This shouldn't be part of this module.
> 
>     * Token validation
> 
>     * UUID extraction from the Authorization header
> 
>     * Claims extraction
> 
>     * Exception handling for expired or tampered tokens
> 
> 
> All the JWT token logic is already in jwt-core and should remain centralized there. This PR should be limited to GitHub integration only (validating usernames and, in the case of auth-service, sharing code and OAuth profiles).

I completely misunderstood the use of Github on the Auth microservice. I thought that it just checked if the user existed, but rather it manages the OAuth flow with GitHub (exchange code → access_token) and validates that a token corresponds to a real GitHub user (calling user-info-uri and parsing login).

In order for this PR to keep it rather small, i suggest creating another task and PR to create the extra two services in the github-core module.

github-core/src/main/java/com/itachallenge/githubcore/service/GithubApiService.java
isaac-diez: Yes, I have changed it to GithubUnavailableException,

github-core/src/main/java/com/itachallenge/githubcore/service/GithubApiService.java
isaac-diez: OK. It is renamed now to GithubApiServiceImpl

acocinas:
Hey there!

I've reviewed the changes you've implemented. The code looks great, and you've addressed all the suggestions very effectively.

Great work!

github-core/src/main/java/com/itachallenge/githubcore/service/GithubApiService.java
isaac-diez: After in depth analyzing your approach, I have come to the same conclusion. It is clearer going with the enum FOUND/NOT_FOUND. 
You are right, it never returned a false. I totally misused the exception thrown. Now the exception is gone and the service looks and behaves in a much clearer way. Thanks for the tip.

github-core/src/main/java/com/itachallenge/githubcore/service/GithubApiService.java
ivilarop: Great thks!

github-core/src/main/java/com/itachallenge/githubcore/service/GithubApiService.java
ivilarop: Great! thks

github-core/src/main/java/com/itachallenge/githubcore/service/GithubApiService.java
ivilarop: thks!

github-core/src/main/java/com/itachallenge/githubcore/exception/GithubApiException.java
ivilarop: Thks!

ivilarop (approved):

PR #960, opened 2025-09-12
update(user): Integrate validation in user creation endpoint (#650)

isaac-diez:
### Summary

This PR updated the behaviour of AdminCreateUserService() so the username is checked against the  a centralized Github Username Validation that will be shared with User and Auth microservices. 

This PR is the second part of 3 which are derived from the following tasks:
- [#649](https://tree.taiga.io/project/luisvi-ita-challenges/task/649) 1/3 Create GitHub Username Validation in Shared Module
- [#650](https://tree.taiga.io/project/luisvi-ita-challenges/task/650) 2/3 Integrate Validation in User Creation Endpoint
- [#651](https://tree.taiga.io/project/luisvi-ita-challenges/task/651) 3/3 Refactor & Reuse in Auth-Service

**Context**

When creating a new user in the user microservice, the GitHub username provided needed to be validated against GitHub’s user base to ensure it actually exists. Previously, this validation was missing.

**Before**

The user microservice allowed creating users with any GitHub username, even if it did not exist. No external check against GitHub’s API was performed.

**After**

The user microservice now validates GitHub usernames by calling the GitHub API via the new github-core module. 
On user creation, the service checks if the GitHub user exists and rejects invalid usernames.

**Impact**

- Breaking change: Clients must now provide valid GitHub usernames when creating users (invalid ones will be rejected with a validation error).
- Functional case (GithubUserStatus.NOT_FOUND) is now translated in user-service into a NotFoundException, which results in 404 Not Found to clients.
- Provider failures (GithubUnavailableException, timeouts, 5xx) are surfaced from github-core and consistently mapped in GlobalExceptionHandler to 503 Service Unavailable/504 Gateway Timeout.

IMPORTANT: 
-
**Since this task depends on the #649 all its changes and new contributions also appear in this PR, so refrain from checking everything that falls under the github-core folder.**

### Main tasks:
- Refactor AdminCreateUserService to implement Github username validation for which the creation of a new user is depends on the username being a Github user.
- Update all dependent classes.
- Keep all tests green in case the new implementation needs some refactoring.

### Changes 

**App.java**
Added annotation to import the GithubServiceConfig() class to get the Github API URL to validate the username.
It is important because, this class needs the github user info uri to be in the application.yml or as environment variables of every micro that is using the Github-core module until it is moved into a central config server. Here is, for reference, the GithubServiceConfig():

    @Configuration
    public class GithubServiceConfig {

    @Value("${github.user-info-uri}")
    private String githubApiUrl;

    @Bean
    public GithubApiService githubApiService(WebClient.Builder webClientBuilder) {
        return new GithubApiServiceImpl(webClientBuilder, githubApiUrl);
        }
    }

and the application.yml:

    github:
      user-info-uri: https://api.github.com

**ExternalGithubService  & ExternalGithubServiceImpl**
In order to fully decouple the main service from the github-core module, this interface and its implementation take care of passing true/false on github username validation to the main service.

**AdminCreateUserService**
Passed the interface ExternalGithubService as an argument and run the username by its userExists() method right after checking if the user is alredy in the database.

**APIErrorResponse**
Created to conveniently inform when the Github API fails.
--> TECHNICAL DEBT: Right now the handler is misaligned so the APIResponse should be adapted  to the rest of the exceptions so that they follow the same format.

**UserGlobalExceptionHandler**
Added the GithubUnavailableException handler for when the Github API is unavailable (code 503) or times out (504), returning an APIErrorResponse DTO.
--> TECHNICAL DEBT: a more robust solution to throw codes 503 or 504 and informs more securely.

**AdminCreateUserServiceTest**
Added the following tests:
- for the situation when the user in not on the database but neither is a Github user.
- for when the Github API fails and should propagate error
- to check that the new user should always have default USER role

**ExternalGithubServiceTest & APIErrorResponseTest**
Created to test the new middleware service and DTO.

**AdminCreateUserControllerTest**
The tests were not returning the right server codes because they needed injecting of @ContextConfiguration and @Import of the UserGlobalExceptionHandler.class to throw the right exceptions and Http responses. 

**UserGlobalExceptionHandlerTest**
Tested the new GithybUnavailableException handler that checks whether a 503 or 504 code is thrown.

**UserControllerSpringTest & UserSolucitonServiceImplTest**
The new ExternalGithubService bean is needed to be injected for the test to work.

**settings.gradle**
Include github-core for the build to use the new module and its services.

**CHANGELOG.md & .env.CI,dev**
Moved to version 1.5.0 of user microservice.

**README.md** for Github-core module
This was left from the previous PR where i created the module. It explains what the new module Github-core does, structure and usage.

**POSTMAN TESTS**
Screenshot-2025-09-16-17:38:27

## 📋 PR Author Self-check
✅ PR is focused and isolated
✅ Branch follows naming conventions.
✅ No debug logs left.
✅ Exception mapped correctly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### [itachallenge-user-1.5.0-RELEASE] - 2025-09-12
+
+### Refactored
+- Added github username validation to user microservice upon creation of a new user  (Taiga [#650], PR [#960])
+
 ### [itachallenge-githubcore-1.0.0-RELEASE] - 2025-09-11
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,10 +3,13 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-### [itachallenge-user-1.5.0-RELEASE] - 2025-09-12
+### [itachallenge-user-2.0.0-RELEASE] - 2025-09-12
 
 ### Changed
-- Added github username validation to user microservice upon creation of a new user  (Taiga [#650], PR [#960])
+- Added github username validation to user microservice upon creation of a new user (Taiga [#650], PR [#960])
+  - Before: POST /users/create accepted any githubUsername value and returned success (201/ok) even if that wasn't a GitHub username. 
+  - After: POST /users/create now calls the GitHub API and rejects the request when the GitHub username does not exist. 
+  - Clients that previously relied on creating users with invalid GitHub usernames will now get errors for some requests that used to succeed.
 
 ### [itachallenge-githubcore-1.0.0-RELEASE] - 2025-09-11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### [itachallenge-user-1.5.0-RELEASE] - 2025-09-12
 
-### Refactored
+### Changed
 - Added github username validation to user microservice upon creation of a new user  (Taiga [#650], PR [#960])
 
 ### [itachallenge-githubcore-1.0.0-RELEASE] - 2025-09-11
--- a/conf/.env.CI.dev
+++ b/conf/.env.CI.dev
@@ -2,7 +2,7 @@ MICROSERVICE_DEPLOY_itachallenge-auth=itachallenge-auth
 MICROSERVICE_VERSION_itachallenge-auth=2.0.5-RELEASE
 
 MICROSERVICE_DEPLOY_itachallenge-user=itachallenge-user
-MICROSERVICE_VERSION_itachallenge-user=1.4.2-RELEASE
+MICROSERVICE_VERSION_itachallenge-user=1.5.0-RELEASE
 
 MICROSERVICE_DEPLOY_itachallenge-challenge=itachallenge-challenge
 MICROSERVICE_VERSION_itachallenge-challenge=2.4.2-RELEASE
--- a/conf/.env.CI.dev
+++ b/conf/.env.CI.dev
@@ -2,7 +2,7 @@ MICROSERVICE_DEPLOY_itachallenge-auth=itachallenge-auth
 MICROSERVICE_VERSION_itachallenge-auth=2.0.5-RELEASE
 
 MICROSERVICE_DEPLOY_itachallenge-user=itachallenge-user
-MICROSERVICE_VERSION_itachallenge-user=1.5.0-RELEASE
+MICROSERVICE_VERSION_itachallenge-user=2.0.0-RELEASE
 
 MICROSERVICE_DEPLOY_itachallenge-challenge=itachallenge-challenge
 MICROSERVICE_VERSION_itachallenge-challenge=2.4.2-RELEASE
--- a/github-core/README.md
+++ b/github-core/README.md
@@ -1,0 +1,38 @@
+# ITA GITHUB-CORE MODULE
+
+`github-core` is a shared library module designed to centralize all Github checks logic across ITA Challenge microservices.
+
+## Features
+
+- 👤 **Username Validation**: Checks whether the username is a valid Github username.
+- ⚙️ **Configuration**: easy change of API Github URL should there be the need.
+- ⚠️ **Exceptions**: for when the Username is not found in Github and for when the API is not working.
+- 🧪 **Test Coverage**: Comes with unit tests to ensure reliability and ease of maintenance.
+- 🚫 **No BootJar**: This module is a pure library and not a Spring Boot application.
+
+## Usage
+
+Add the dependency to your Spring Boot microservice in `build.gradle`:
+
+```groovy
+implementation(project(":github-core"))
+
+testImplementation project(':github-core')
+```
+
+## Project structure
+```markdown
+github-core/
+├── src/main/java/com/itachallenge/githubcore/
+│   ├── service/      # IGithubApiService & GithubApiService
+    ├── exception/    # GithubApiException & GithubUserNotFoundException
+    ├── config/       # GithubServiceConfig
+├── src/test/         # Unit tests
+├── build.gradle      # Gradle module configuration
+└── readme.md         # This text file           
+```
+## Development notes
+
+- The module disables bootJar since it’s not a standalone Spring Boot app.
+- SonarQube quality gate compliance with minimum test coverage is enforced.
+- This library is versioned and updated in sync with other ITA Challenge services.
--- a/itachallenge-user/src/main/java/com/itachallenge/user/dto/APIErrorResponse.java
+++ b/itachallenge-user/src/main/java/com/itachallenge/user/dto/APIErrorResponse.java
@@ -1,0 +1,25 @@
+package com.itachallenge.user.dto;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+import java.time.Instant;
+
+@AllArgsConstructor
+@NoArgsConstructor
+@Data
+public class APIErrorResponse {
+
+    @JsonProperty("error")
+    String error;
+
+    @JsonProperty("message")
+    String message;
+
+    @JsonProperty("timestamp")
+    Instant timestamp;
+
+}
+
--- a/itachallenge-user/src/main/java/com/itachallenge/user/dto/APIErrorResponse.java
+++ b/itachallenge-user/src/main/java/com/itachallenge/user/dto/APIErrorResponse.java
@@ -4,12 +4,14 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
+import lombok.Setter;
 
 import java.time.Instant;
 
 @AllArgsConstructor
 @NoArgsConstructor
 @Getter
+@Setter
 public class APIErrorResponse {
 
     @JsonProperty("error")
--- a/itachallenge-user/src/main/java/com/itachallenge/user/dto/APIErrorResponse.java
+++ b/itachallenge-user/src/main/java/com/itachallenge/user/dto/APIErrorResponse.java
@@ -2,14 +2,14 @@ package com.itachallenge.user.dto;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 import lombok.AllArgsConstructor;
-import lombok.Data;
+import lombok.Getter;
 import lombok.NoArgsConstructor;
 
 import java.time.Instant;
 
 @AllArgsConstructor
 @NoArgsConstructor
-@Data
+@Getter
 public class APIErrorResponse {
 
     @JsonProperty("error")
--- a/itachallenge-user/src/main/java/com/itachallenge/user/exception/GithubUserNotFoundException.java
+++ b/itachallenge-user/src/main/java/com/itachallenge/user/exception/GithubUserNotFoundException.java
@@ -1,7 +1,0 @@
-package com.itachallenge.user.exception;
-
-public class GithubUserNotFoundException extends RuntimeException {
-    public GithubUserNotFoundException(String username) {
-        super("The username '" + username + "' is not a Github username.");
-    }
-}
--- a/itachallenge-user/src/main/java/com/itachallenge/user/exception/GithubUserNotFoundException.java
+++ b/itachallenge-user/src/main/java/com/itachallenge/user/exception/GithubUserNotFoundException.java
@@ -1,0 +1,6 @@
+package com.itachallenge.user.exception;
+
+public class GithubUserNotFoundException extends RuntimeException {
+  public GithubUserNotFoundException(String message) {super(message);
+  }
+}
--- a/itachallenge-user/src/main/java/com/itachallenge/user/exception/GithubUserNotFoundException.java
+++ b/itachallenge-user/src/main/java/com/itachallenge/user/exception/GithubUserNotFoundException.java
@@ -1,0 +1,7 @@
+package com.itachallenge.user.exception;
+
+public class GithubUserNotFoundException extends RuntimeException {
+    public GithubUserNotFoundException(String username) {
+        super("The username '" + username + "' is not a Github username.");
+    }
+}
--- a/itachallenge-user/src/main/java/com/itachallenge/user/exception/GithubUserNotFoundException.java
+++ b/itachallenge-user/src/main/java/com/itachallenge/user/exception/GithubUserNotFoundException.java
@@ -1,6 +1,0 @@
-package com.itachallenge.user.exception;
-
-public class GithubUserNotFoundException extends RuntimeException {
-  public GithubUserNotFoundException(String message) {super(message);
-  }
-}
--- a/itachallenge-user/src/main/java/com/itachallenge/user/exception/UserGlobalExceptionHandler.java
+++ b/itachallenge-user/src/main/java/com/itachallenge/user/exception/UserGlobalExceptionHandler.java
@@ -7,7 +7,6 @@ import org.springframework.web.bind.MethodArgumentNotValidException;
 import org.springframework.web.bind.annotation.ExceptionHandler;
 import org.springframework.web.bind.annotation.RestControllerAdvice;
 import org.springframework.web.method.annotation.MethodArgumentTypeMismatchException;
-import com.itachallenge.githubcore.exception.GithubUserNotFoundException;
 
 import jakarta.validation.ConstraintViolationException;
 
@@ -81,12 +80,6 @@ public class UserGlobalExceptionHandler {
     @ExceptionHandler(UsernameAlreadyExistsException.class)
     public ResponseEntity<String> handleUsernameAlreadyExistsException(UsernameAlreadyExistsException e) {
         return ResponseEntity.status(HttpStatus.CONFLICT).body(e.getMessage());
-    }
-
-    @ExceptionHandler(GithubUserNotFoundException.class)
-    public ResponseEntity<String> handleGithubUserNotFound(Exception ex) {
-        return ResponseEntity.status(HttpStatus.NOT_FOUND)
-                .body(ex.getMessage());
     }
 
 }
--- a/itachallenge-user/src/main/java/com/itachallenge/user/exception/UserGlobalExceptionHandler.java
+++ b/itachallenge-user/src/main/java/com/itachallenge/user/exception/UserGlobalExceptionHandler.java
@@ -5,7 +5,6 @@ import com.itachallenge.user.dto.APIErrorResponse;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
-import org.springframework.web.ErrorResponse;
 import org.springframework.web.bind.MethodArgumentNotValidException;
 import org.springframework.web.bind.annotation.ExceptionHandler;
 import org.springframework.web.bind.annotation.RestControllerAdvice;
--- a/itachallenge-user/src/main/java/com/itachallenge/user/exception/UserGlobalExceptionHandler.java
+++ b/itachallenge-user/src/main/java/com/itachallenge/user/exception/UserGlobalExceptionHandler.java
@@ -7,6 +7,7 @@ import org.springframework.web.bind.MethodArgumentNotValidException;
 import org.springframework.web.bind.annotation.ExceptionHandler;
 import org.springframework.web.bind.annotation.RestControllerAdvice;
 import org.springframework.web.method.annotation.MethodArgumentTypeMismatchException;
+import com.itachallenge.githubcore.exception.GithubUserNotFoundException;
 
 import jakarta.validation.ConstraintViolationException;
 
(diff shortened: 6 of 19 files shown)


acocinas (approved):
Hey Isaac,

Great work on this PR! The implementation is solid and the integration with the github-core module is very clean. The test coverage is excellent, and it's clear you've considered all the main scenarios.

The code is in a good state and ready to be merged. I'm leaving a minor comment on the CHANGELOG for clarification, but it doesn't block the PR.


Great job!

CHANGELOG.md
```diff
@@ -3,6 +3,11 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### [itachallenge-user-1.5.0-RELEASE] - 2025-09-12
+
+### Refactored
```
acocinas: Hi. Just a quick question on the CHANGELOG: would ### Added or ### Changed be a more accurate label for this entry? The task implemented a new validation feature rather than just an internal refactor. Just a thought for consistency.

Heyvkram (approved):
Echoing the previous comment, this is a fantastic PR, Isaac. The implementation is very solid, and the decision to create a separate github-core module is a great architectural choice that's been cleanly integrated.

The test coverage is comprehensive, and it's clear all edge cases were considered. 

Take a look at the comment @acocinas made about the CHANGELOG!

Excellent work.

CHANGELOG.md
isaac-diez: OK, I see that now. Yes, it is more accurate. I change that.

acocinas: ok buddy, thank you

ivilarop:
> ### Summary
> This PR updated the behaviour of AdminCreateUserService() so the username is checked against the a centralized Github Username Validation that will be shared with User and Auth microservices.
> 
> This PR is the second part of 3 which are derived from the following tasks:
> 
> * [#649](https://tree.taiga.io/project/luisvi-ita-challenges/task/649) 1/3 Create GitHub Username Validation in Shared Module
> * [#650](https://tree.taiga.io/project/luisvi-ita-challenges/task/650) 2/3 Integrate Validation in User Creation Endpoint
> * [#651](https://tree.taiga.io/project/luisvi-ita-challenges/task/651) 3/3 Refactor & Reuse in Auth-Service
> 
> ## IMPORTANT:
> **Since this task depends on the #649 all its changes and new contributions also appear in this PR, so refrain from checking everything that falls under the github-core folder.**
> 
> ### Main tasks:
> * Refactor AdminCreateUserService to implement Github username validation for which the creation of a new user is depends on the username being a Github user.
> * Update all dependent classes: App.java and application.yml
> * Keep all tests green in case the new implementation needs some refactoring.
> 
> ### Changes
> **App.java** Added annotation to use the GithubServiceConfig class to get the Github API URL to validate the username
> 
> **AdminCreateUserService** Passed the interface IGithubApiService as an argument and run the username by its userExists() method right after checking if the user is alredy in the database.
> 
> **UserGlobalExceptionHandler** Added the GithubUserNotFoundException handler for when the username is not a Github user the exception is thrown.
> 
> **AdminCreateUserServiceTest** Added the following tests:
> 
> * for the situation when the user in not on the database but neither is a Github user.
> * for when the Github API fails and should propagate error
> * to check that the new user should always have default USER role
> 
> **AdminCreateUserControllerTest** The tests were not returning the right server codes because they needed injecting of @ContextConfiguration and @import of the UserGlobalExceptionHandler.class to throw the right exceptions and Http responses.
> 
> **settings.gradle** Include github-core for the build to use the new module and its services.
> 
> **IGithubApiService & GithubApiService** Interface and implementation that take care of the mail logic of validating the username with Github. Returns true if server returns a 200 cods or false if the server returns a 4xx or 5xx code.
> 
> **CHANGELOG.md & .env.CI,dev** Moved to version 1.5.0 of user microservice.
> 
> **README.md** Explaining what this module does, structure and usage.
> 
> **POSTMAN TESTS** Screenshot-2025-09-16-17:38:27
> 
> ## 📋 PR Author Self-check
> ✅ PR is focused and isolated ✅ Branch follows naming conventions. ✅ No debug logs left. ✅ Exception mapped correctly. ✅ PR clearly documents scope and purpose.

@isaac-diez @Heyvkram 

One thing you could reflect on: the PR description is very detailed about the what (classes, methods, validation), but it’s missing some higher-level context.

Ask yourself:

1. Which microservice does this PR affect?
2. How does it fit into the current login flow?
3. What was the behavior before? What is the behavior after?
4. What is the impact for API consumers (SemVer)?

Adding a short “Context / Before / After / Impact” section would make the PR description easier to understand for reviewers who are not deep into the code.

What do you think — could you try to rewrite the PR description with this extra context?

CHANGELOG.md
```diff
@@ -3,6 +3,11 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### [itachallenge-user-1.5.0-RELEASE] - 2025-09-12
```
ivilarop: One thing you should reflect on is the versioning choice.
Ask yourself:

1. Before this change, what requests did the user microservice accept?
2. After this change, what requests are now rejected?
3. Does this break compatibility for any existing clients?
4. Based on Semantic Versioning, what kind of release would that imply?

Try to reason it out and update the changelog accordingly.

conf/.env.CI.dev
```diff
@@ -2,7 +2,7 @@ MICROSERVICE_DEPLOY_itachallenge-auth=itachallenge-auth
 MICROSERVICE_VERSION_itachallenge-auth=2.0.5-RELEASE
 
 MICROSERVICE_DEPLOY_itachallenge-user=itachallenge-user
-MICROSERVICE_VERSION_itachallenge-user=1.4.2-RELEASE
+MICROSERVICE_VERSION_itachallenge-user=1.5.0-RELEASE
```
ivilarop: see previous comment

github-core/README.md
```diff
@@ -0,0 +1,38 @@
+# ITA GITHUB-CORE MODULE
```
ivilarop, line 1: Why did this change happen? Does it have to do with the PR? I know you said, "Since this task depends on #649, all its changes and new contributions also appear in this PR, so refrain from checking everything that falls under the github-core folder." But I don't see this change in #649.
You also added a description of this change to the PR.

itachallenge-user/src/main/java/com/itachallenge/user/service/AdminCreateUserService.java
```diff
@@ -18,9 +21,11 @@ public class AdminCreateUserService implements IAdminCreateUserService {
     private static final Logger log = LoggerFactory.getLogger(AdminCreateUserService.class);
 
     private final UserRepository userRepository;
+    private final GithubApiService githubApiService;
```
ivilarop: But take a moment to think about this:

1. How coupled is the user microservice to the details of github-core right now?
2. What would break if github-core changes its API or internals?
3. Is this the same level of isolation you achieved when integrating jwt-core?

Please review with these questions in mind and adjust so the micro is less dependent on the internals of github-core.

itachallenge-user/src/main/resources/application.yml
```diff
@@ -32,6 +32,9 @@ server:
   tomcat:
     max-http-form-post-size: 2097152
 
+github:
```
ivilarop, line 35: The GitHub base URL shouldn’t live in the user micro’s application.yml.
That config belongs to github-core

itachallenge-user/src/main/java/com/itachallenge/user/service/AdminCreateUserService.java
```diff
@@ -29,34 +34,36 @@ public Mono<AdminCreateUserResponseDto> createUser(AdminCreateUserRequestDto req
...
+                                    return userRepository.save(newUser)
+                                            .map(savedUser -> AdminCreateUserResponseDto.builder()
+                                                    .userId(savedUser.getUuid().toString())
+                                                    .username(savedUser.getUsername())
+                                                    .role(savedUser.getRole().toString())
+                                                    .build())
+                                            .doOnSuccess(responseDto ->
+                                                    log.info("Successfully created user '{}'", responseDto.getUsername()));
+                                })
+                                .doOnError(e -> !(e instanceof UsernameAlreadyExistsException || e instanceof GithubUserNotFoundException), e ->
```
ivilarop: **Current behavior**

1. In AdminCreateUserService you use doOnError(...) to log unexpected errors.
2. doOnError only performs a side-effect (logging). It does not change the reactive flow and does not translate provider errors into an API response.

**Why this is a problem**

1. When github-core fails (5xx/timeout/unexpected), the service just logs and lets it bubble up, which typically ends as a generic 500.
2. This duplicates logging and hides the intent; API consumers don’t get a clear, consistent status (e.g., 503).

**What we should do instead**

**1.** Remove the doOnError used to “handle” provider failures (it doesn’t handle them).
**2.** Handle outcomes centrally in the @ControllerAdvice:

- Provider failures: do not create a new exception in user — import com.itachallenge.githubcore.exception.GithubUnavailableException (from github-core) in the handler and return 503 Service Unavailable (or 504 on timeouts).
- Functional “user not found”: keep it as normal flow from github-core (FOUND/NOT_FOUND).
 In the user micro, translate NOT_FOUND to a domain not-found exception (you can reuse your existing NotFoundException) so the handler returns 404.

**3.** If you need to transform exceptions in the service layer, use onErrorMap/onErrorResume (which affect the flow), not doOnError.
**4.** (Nice to have) Return a structured JSON error payload instead of raw strings for consistency.

Please update the error handling accordingly and add a short note in the PR description explaining:

- how provider failures are mapped (503/504) and
- how the NOT_FOUND case is translated to 404.

itachallenge-user/src/main/java/com/itachallenge/user/exception/GithubUserNotFoundException.java
ivilarop, line 1: Don't create a new exception. Reuse the one you already have:

When github-core returns NOT_FOUND (enum/empty), throw new NotFoundException("GitHub user not found: " + username) in your service.

You already have a NotFoundException handler → it responds with a 404. ✔️

itachallenge-user/src/main/java/com/itachallenge/user/exception/UserGlobalExceptionHandler.java
```diff
@@ -82,4 +82,9 @@ public ResponseEntity<String> handleUsernameAlreadyExistsException(UsernameAlrea
         return ResponseEntity.status(HttpStatus.CONFLICT).body(e.getMessage());
     }
 
+    @ExceptionHandler(GithubUserNotFoundException.class)
```
ivilarop: see previous comment

ivilarop (changes requested):

CHANGELOG.md
isaac-diez: After some analysis, the changes affect the API contract  because some previously valid requests will now be rejected. According to SemVer that is a breaking change so a major version bump is appropriate. I, then, moved the version to 2.0.

conf/.env.CI.dev
isaac-diez: As per my previous comment's response, I have moved the version to 2.0.0

github-core/README.md
isaac-diez: I wanted to push as quickly as possible task #649 so I decided to postpone the Github-core README document to this task's (#650) PR.

itachallenge-user/src/main/java/com/itachallenge/user/exception/GithubUserNotFoundException.java
isaac-diez: I created a new exception with clarity and single responsibility in mind. But it might reduce maintainability.
I have deleted the GithubUserNotFoundException and used NotFoundException instead.

itachallenge-user/src/main/java/com/itachallenge/user/exception/UserGlobalExceptionHandler.java
isaac-diez: It's gone now.

itachallenge-user/src/main/java/com/itachallenge/user/service/AdminCreateUserService.java
isaac-diez: After deeper analysis, I can see that the service is highly coupled with the new Github-core module because it directly depends on:

- GithubApiService (the service interface/class from github-core).
- GithubUserStatus (an enum defined inside github-core).

The user micro is making branching logic (if (status == GithubUserStatus.NOT_FOUND)) based on an internal detail of how github-core treats GitHub responses.

This means if github-core changes its enum values, or switches to a different mechanism (e.g., returning a DTO or throwing exceptions instead of GithubUserStatus), the user micro will break.

For this, i have refactored the service into using an abstraction layer with a new ExternalGithubService that takes care of the boolean response of userExists(username) so the main services only depends on a boolean answer and is decoupled from the new Github-core module.

itachallenge-user/src/main/java/com/itachallenge/user/service/AdminCreateUserService.java
isaac-diez: You are right. I was only throwing the log.error without specific exception. I have refactored the service to throw GithubUnavailableException to return 503 or 504 when needed.

Functional case (GithubUserStatus.NOT_FOUND) is now translated in user-service into a NotFoundException, which results in 404 Not Found to clients.

Provider failures (GithubUnavailableException, timeouts, 5xx) are surfaced from github-core and consistently mapped in GlobalExceptionHandler to 503 Service Unavailable/504 Gateway Timeout.

Removed doOnError in AdminCreateUserService to avoid duplicated logging and misleading error handling.

All error responses now follow a structured JSON format for consistency with a newly created APIErrorResponse DTO.

itachallenge-user/src/main/resources/application.yml
isaac-diez: Since github-core is a library and not a Spring Boot app, it cannot own application.yml (Spring won’t even load it). The correct approach is:

- github-core defines what configuration keys it expects (github.user-info-uri).

- Each runnable microservice (user-micro, auth-micro, …) provides values in its application.yml (or better, in Spring Cloud Config if you want centralization).

So unless everything is moved to a central config server, the application.yml entry must stay in the runnable microservice(s).

acocinas (approved):
I have reviewed all the changes and the evolution of this PR. The work has done is, quite simply, outstanding. You not only solved the initial task, but also took very technical and complex feedback and implemented it flawlessly, significantly improving the microservice's architecture.

Final Analysis
What started as a good PR for validation has evolved into an exceptional piece of code thanks to the review process. Significant improvements have been made at every layer:

Reduced Coupling: The creation of the ExternalGithubService interface is a brilliant solution. It completely decouples the user microservice from the internal details of github-core, such as the GithubUserStatus enum. The user service now only depends on a simple Mono<Boolean>, which makes the code far more flexible and maintainable.

Robust Error Handling: The centralization of error handling in UserGlobalExceptionHandler, the use of onErrorMap, and the structured APIErrorResponse demonstrate an advanced understanding of Spring Boot and reactive best practices.

Adherence to Standards: The versioning correction to 2.0.0 and the justification for the breaking change are a clear example of a mature approach to Semantic Versioning.

This PR has gone through a rigorous review process and has been greatly improved as a result.

Heyvkram (approved):
Approved.

This is an excellent contribution. The preceding review has already highlighted the significant technical achievements in this PR, and I am in full agreement. The evolution of the code through the review process has resulted in a robust and maintainable solution. 
Good job!

CHANGELOG.md
ivilarop: Ok perfect! its ok!

conf/.env.CI.dev
ivilarop: Ok perfect! its ok!

github-core/README.md
ivilarop: If you include a change that might seem out of scope (such as the GitHub-core README), it's important to also make it explicit in the PR description. This way, any reviewer immediately understands why it's there, and clear traceability between tasks and PRs is maintained.

Please leave it noted in the PR description.

itachallenge-user/src/main/java/com/itachallenge/user/service/AdminCreateUserService.java
ivilarop: Great reasoning and solid refactor to reduce coupling 👏.
One minor suggestion for consistency: in the auth micro we talk to jwt-core through a facade (IAuthJwtFacade / AuthJwtFacade).
To keep naming coherent across services, could we mirror that here for github-core?

This way any dev recognizes the pattern immediately and we avoid leaking provider details into the domain.

itachallenge-user/src/main/resources/application.yml
ivilarop: Ok, I understand now: the configuration value (e.g. github.user-info-uri) must indeed live in the user micro (or any other runnable microservice) and then be provided to the github-core library at runtime. That part makes sense 👍.

What I’m missing, however, is that this is not explained anywhere in the PR description. Right now it’s not clear for another developer why the property has to be in the micro’s application.yml instead of inside github-core.

I suggest adding a short Configuration section in the PR description to clarify:

- github-core defines the property key(s) it expects.
- Each runnable microservice must provide the value in its own application.yml or via env vars.
- Example snippet of how to set it up.

This way, anyone reading the PR understands both the technical change and the configuration requirement.

itachallenge-user/src/main/java/com/itachallenge/user/service/AdminCreateUserService.java
ivilarop: OK! The current Boolean mapping (status != NOT_FOUND) is fine as an MVP ✅, but it hides some domain intent.
As technical debt, it would be worth replacing the Boolean with a more explicit result type in the future. That way the meaning is clearer (e.g. FOUND vs NOT_FOUND) and the compiler will force us to handle any new cases if github-core evolves.

If a new value (e.g., RATE_LIMITED) appears in the future, the compiler will force the service to decide how to handle it. It won't accidentally "sneak" in as true.

Not blocking for now, but good to keep in mind for a follow-up improvement.

itachallenge-user/src/main/java/com/itachallenge/user/exception/GithubUserNotFoundException.java
ivilarop: Ok thks! its ok!

itachallenge-user/src/main/java/com/itachallenge/user/exception/UserGlobalExceptionHandler.java
ivilarop: Ok thks! its ok!

itachallenge-user/src/main/java/com/itachallenge/user/dto/APIErrorResponse.java
ivilarop, line 1: It's good that it was created, and that's what I told you, but one clarification: Currently, it's only being used in the GithubUnavailableException handler.

It would be good to extend it to all handlers (404, 409, etc.) so that the client always receives the same format.

APIErrorResponse uses error + message, but there's no stable code/key, status, path, traceId, etc. It's difficult to parse from the front end and correlate in logs. (Optional: browse Problem Details: type, title, status, detail, instance.)

Have you considered this as technical debt?

Please note that we should modify the PR description to consider that right now the handler is misaligned and that we are going to create a technical debt to adapt the APIResponse to the rest of the exceptions so that they follow the same format.

itachallenge-user/src/main/java/com/itachallenge/user/exception/UserGlobalExceptionHandler.java
```diff
@@ -82,4 +85,19 @@ public ResponseEntity<String> handleUsernameAlreadyExistsException(UsernameAlrea
         return ResponseEntity.status(HttpStatus.CONFLICT).body(e.getMessage());
     }
 
+    @ExceptionHandler(GithubUnavailableException.class)
+    public ResponseEntity<APIErrorResponse> handleGithubUnavailable(GithubUnavailableException ex) {
+        HttpStatus status;
+
+        if ("timeout".equalsIgnoreCase(ex.getMessage())) {
```
ivilarop, line 92: It works, but string matching in getMessage() is fragile.

Deciding 503 vs. 504 with ex.getMessage().equalsIgnoreCase("timeout") is fragile: if the message changes, if someone internationalizes it or adjusts it, you'll inadvertently break the mapping.

itachallenge-user/src/main/java/com/itachallenge/user/exception/UserGlobalExceptionHandler.java
```diff
@@ -82,4 +85,19 @@ public ResponseEntity<String> handleUsernameAlreadyExistsException(UsernameAlrea
...
+            status = HttpStatus.GATEWAY_TIMEOUT; // 504
+        } else {
+            status = HttpStatus.SERVICE_UNAVAILABLE; // 503
+        }
+
+        return ResponseEntity.status(status).body(
+                new APIErrorResponse("GitHub API error", ex.getMessage(), Instant.now())
```
ivilarop, line 99: ex.getMessage() goes to the client as is. It may leak internal details or vary between environments. A "secure" message and detailed internal log are preferred.

ivilarop (changes requested):

github-core/README.md
isaac-diez: Ok, noted. The summary is updated now.

itachallenge-user/src/main/java/com/itachallenge/user/dto/APIErrorResponse.java
isaac-diez: Yes, I see your point and it makes total sense. I'll note it as technical debt.

itachallenge-user/src/main/java/com/itachallenge/user/exception/UserGlobalExceptionHandler.java
isaac-diez: Noted as technical debt as well.

itachallenge-user/src/main/java/com/itachallenge/user/exception/UserGlobalExceptionHandler.java
isaac-diez: I think this is best left as technical debt. I have checked and it seems that I might be needing to modify the constructor of the exception created in the Github-core module GithubUnavailableException() in order to pass a cause.

ivilarop (approved):

PR #654, opened 2024-11-14
Solution.Challenge.out feature#468v5

Asousa12:
Adaptation of the getSolution to obtain the required information and output from Json.

Solution challenge out

PR #961, opened 2025-09-15
feat(challenge): require tags on challenge creation (#654)

Heyvkram:
Currently, the POST /challenges endpoint allows a challenge to be created without any associated tags. This PR introduces validation to make it mandatory for every new challenge to be created with at least one tag.

This change ensures data integrity and improves the searchability and categorization of challenges from the moment they are created.

### **Changes Made 💻**

-  DTO Validation: Added the @NotEmpty annotation to the tags field in the ChallengeCreateDto.java DTO.

 - Unit Tests: Added a new test (addChallenge_emptyTags_statusBadRequest) in ChallengeControllerTest.java to verify that the API returns a 400 Bad Request when the tags list is empty.

 - Documentation: Updated CHANGELOG.md to reflect the new feature.

 - Microservice Version: Bumped the microservice version to 3.0.0-RELEASE in the .env.CI.dev file to reflect the breaking change introduced in the API.

### **Technical Implementation Details ⚙️**

The validation is implemented using the Spring Boot and Jakarta Bean Validation ecosystem. It's important to understand the full mechanism to demystify how the 400 error is generated:

@NotEmpty Annotation: This annotation on the tags field in the DTO defines the business rule: the list cannot be null or empty.

Validation Trigger: In the Controller, the request body parameter is annotated with @Valid. This tells Spring to trigger the validation process on the incoming DTO object before executing the controller method.

Validation Engine: Spring Boot uses Hibernate Validator under the hood as its default implementation of the Jakarta Bean Validation standard.

Error Handling:

If Hibernate Validator finds a violation (in this case, an empty tags list), it internally throws an exception (MethodArgumentNotValidException for Spring MVC or WebExchangeBindException for WebFlux).

Spring Boot's default exception handler (DefaultExceptionHandler) catches this specific exception.

This handler automatically translates the exception into an HTTP 400 Bad Request response. The response body includes the error message we defined in the annotation: "The tags list cannot be empty".

This approach is clean and declarative, as the validation logic resides in the data model (DTO) and requires no additional code in the service or controller layers.

### **How to manually test it? 🧪**

To verify the functionality, you can use Postman to make a POST request to the /itachallenge/api/v1/challenge/challenges endpoint.

Failure Case ❌
Send a request with the tags field as an empty array, null, or missing entirely.

Request Body:

JSON

{
  "challengeTitle": "Test challenge without tags",
  "description": "This is the description",
  "level": "EASY",
  "language": "Java",
  "solution": "This is the solution",
  "topic": "ALL",
  "tags": []
}
Expected Result:
The API must return a 400 Bad Request status code with a response body containing the error message.

Captura de pantalla 2025-09-28 090526

Success Case ✅
Send the same request, but this time with at least one valid UUID in the tags list.

Expected Result:
The API must return a 201 Created status code with the newly created challenge in the response body.

**A change in the Challenge Postman Collection is needed in due to test the change via postman.** 
This change is done in a new task: https://tree.taiga.io/project/luisvi-ita-challenges/task/665
And in a separate PR: (#964) 

The error message is actually not user-friendly, we might consider handling this in GlobalExceptionHandler to return a structured JSON with field + message. This is addressed as technical debt in the issue https://tree.taiga.io/project/luisvi-ita-challenges/issue/701.

### **PR Author Self-check 📋**

✅ I tested my changes locally and verified they work as expected
✅ The PR has a clear and focused purpose (only one feature, fix or refactor)
✅ Title, description, and changelog (if needed) are filled in correctly
✅ There are no leftover merge conflicts or unrelated changes from previous branches
✅ All automated checks (like SonarCloud) have passed
✅ I responded to all previous review comments and only resolved those I truly addressed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-### [itachallenge-challenge-2.4.3-RELEASE] - 2025-09-15
+### [itachallenge-challenge-2.5.0-RELEASE] - 2025-09-15
 
 ### Added
 - Added the @NotEmpty annotation to the tags field in the ChallengeCreateDto.java DTO to make it a mandatory field. (Taiga [#654], PR [#961])
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### [itachallenge-challenge-2.4.3-RELEASE] - 2025-09-15
+
+### Added
+- Added the @NotEmpty annotation to the tags field in the ChallengeCreateDto.java DTO to make it a mandatory field. (Taiga [#654], PR [#961])
+
 ### [itachallenge-githubcore-1.0.0-RELEASE] - 2025-09-11
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-### [itachallenge-challenge-2.5.0-RELEASE] - 2025-09-15
+### [itachallenge-challenge-3.0.0-RELEASE] - 2025-09-15
 
 ### Added
 - Added the @NotEmpty annotation to the tags field in the ChallengeCreateDto.java DTO to make it a mandatory field. (Taiga [#654], PR [#961])
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-### [itachallenge-challenge-3.0.0-RELEASE] - 2025-09-15
+### [itachallenge-challenge-3.0.0-RELEASE] - 2025-09-22
 
 ### Added
 - Added the @NotEmpty annotation to the tags field in the ChallengeCreateDto.java DTO to make it a mandatory field. (Taiga [#654], PR [#961])
--- a/conf/.env.CI.dev
+++ b/conf/.env.CI.dev
@@ -5,7 +5,7 @@ MICROSERVICE_DEPLOY_itachallenge-user=itachallenge-user
 MICROSERVICE_VERSION_itachallenge-user=1.4.2-RELEASE
 
 MICROSERVICE_DEPLOY_itachallenge-challenge=itachallenge-challenge
-MICROSERVICE_VERSION_itachallenge-challenge=2.4.3-RELEASE
+MICROSERVICE_VERSION_itachallenge-challenge=2.5.0-RELEASE
 
 MICROSERVICE_DEPLOY_itachallenge-document=itachallenge-document
 MICROSERVICE_VERSION_itachallenge-document=1.0.2-RELEASE
--- a/conf/.env.CI.dev
+++ b/conf/.env.CI.dev
@@ -5,7 +5,7 @@ MICROSERVICE_DEPLOY_itachallenge-user=itachallenge-user
 MICROSERVICE_VERSION_itachallenge-user=1.4.2-RELEASE
 
 MICROSERVICE_DEPLOY_itachallenge-challenge=itachallenge-challenge
-MICROSERVICE_VERSION_itachallenge-challenge=2.4.2-RELEASE
+MICROSERVICE_VERSION_itachallenge-challenge=2.4.3-RELEASE
 
 MICROSERVICE_DEPLOY_itachallenge-document=itachallenge-document
 MICROSERVICE_VERSION_itachallenge-document=1.0.2-RELEASE
--- a/conf/.env.CI.dev
+++ b/conf/.env.CI.dev
@@ -5,7 +5,7 @@ MICROSERVICE_DEPLOY_itachallenge-user=itachallenge-user
 MICROSERVICE_VERSION_itachallenge-user=1.4.2-RELEASE
 
 MICROSERVICE_DEPLOY_itachallenge-challenge=itachallenge-challenge
-MICROSERVICE_VERSION_itachallenge-challenge=2.5.0-RELEASE
+MICROSERVICE_VERSION_itachallenge-challenge=3.0.0-RELEASE
 
 MICROSERVICE_DEPLOY_itachallenge-document=itachallenge-document
 MICROSERVICE_VERSION_itachallenge-document=1.0.2-RELEASE
--- a/itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/ChallengeCreateDto.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/ChallengeCreateDto.java
@@ -34,5 +34,6 @@ public class ChallengeCreateDto {
     @NotNull(message = "cannot be empty")
     private Topic topic;
 
+    @NotEmpty(message = "The tags list cannot be empty")
     private List<UUID> tags;
 }
--- a/itachallenge-challenge/src/test/java/com/itachallenge/challenge/controller/ChallengeControllerTest.java
+++ b/itachallenge-challenge/src/test/java/com/itachallenge/challenge/controller/ChallengeControllerTest.java
@@ -935,4 +935,26 @@ class ChallengeControllerTest {
         verify(challengeService, times(0)).removeChallengeFromBookmarks(anyString(), anyString());
 
     }
+
+    @Test
+    void addChallenge_emptyTags_statusBadRequest() {
+        String challengeWithEmptyTags = """
+                {
+                     "challengeTitle": "title",
+                     "description": "description",
+                     "level": "EASY",
+                     "language": "Java",
+                     "solution": "solution",
+                     "topic": "ALL",
+                     "tags": []
+                 }
+            """;
+
+        webTestClient.post()
+                .uri("/itachallenge/api/v1/challenge/challenges")
+                .contentType(MediaType.APPLICATION_JSON)
+                .bodyValue(challengeWithEmptyTags)
+                .exchange()
+                .expectStatus().isBadRequest();
+    }
 }


acocinas (approved):
This PR is well-scoped and directly addresses a specific validation issue. The author's use of @NotEmpty for list validation and the clear manual test instructions are excellent.

Suggestions for Improvement
Missing Changelog: The PR description correctly mentions this. It's crucial for tracking changes and should be added before merging.

This PR is ready  once the changelog is updated. The core implementation is sound, and it correctly adds the required validation.

Heyvkram:
Thanks Alfonso!
I have already added the information to CHANGELOG.md and .env.CI.dev!

isaac-diez (approved):
Your solution is very elegant and straightforward. Only thing I see is that when i tried to replicate the error in postman, your instructions missed fields "solution" and "description". But otherwise, it works as expected.

Screenshot from 2025-09-15 13 36 44

ivilarop:
@Heyvkram @isaac-diez @acocinas 

⚠️ According to the PR guidelines, the title should use feat (not feature) and the module should be lowercase:
feat(challenge): require tags on challenge creation (#654)

Take a look and update it to maintain consistency with the rest of the repo.

CHANGELOG.md
```diff
@@ -3,6 +3,11 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### [itachallenge-challenge-2.4.3-RELEASE] - 2025-09-15
```
ivilarop: According to Semantic Versioning, is this a patch change? Let me remind you what each case means:

MAJOR → changes that are incompatible with previous versions.

MINOR → new functionality that is backward compatible.

PATCH → bugfixes or internal changes that don't affect the public API.

It would be good if you could justify the impact on SemVer versioning in the PR description.

conf/.env.CI.dev
```diff
@@ -5,7 +5,7 @@ MICROSERVICE_DEPLOY_itachallenge-user=itachallenge-user
 MICROSERVICE_VERSION_itachallenge-user=1.4.1-RELEASE
 
 MICROSERVICE_DEPLOY_itachallenge-challenge=itachallenge-challenge
-MICROSERVICE_VERSION_itachallenge-challenge=2.4.2-RELEASE
+MICROSERVICE_VERSION_itachallenge-challenge=2.4.3-RELEASE
```
ivilarop: ??? see previous comment

Heyvkram:
@ivilarop 
Good point! Changed!

itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/ChallengeCreateDto.java
```diff
@@ -34,5 +34,6 @@ public class ChallengeCreateDto {
     @NotNull(message = "cannot be empty")
     private Topic topic;
 
+    @NotEmpty(message = "The tags list cannot be empty")
```
ivilarop, line 37: Does the @notempty annotation alone throw the exception? Why does it throw a 400 Bad Request error? I'm asking because I'd like you to explain why it throws the 400. At first glance, and for someone who doesn't have much knowledge, it seems that just adding the @notempty creates the magic.

ivilarop:
@Heyvkram @acocinas @isaac-diez

ivilarop:
@Heyvkram  @acocinas @isaac-diez 

Should the wiki endpoint description be modified? Or the Postman collection?

ivilarop (changes requested):

ivilarop:
@Heyvkram @isaac-diez @acocinas 

By the way, wouldn't it be necessary to add tests?

Heyvkram:
> @Heyvkram @acocinas @isaac-diez
> 
> Should the wiki endpoint description be modified? Or the Postman collection?

@ivilarop 
I don't think it's necessary to touch the wiki part, since the endpoint's functionality hasn't changed. On the other hand, the Postman collection will have to be modified because the raw request body is incorrect. As for the Postman part, I created a new task in Taiga for it, and it will be a different PR.

itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/ChallengeCreateDto.java
Heyvkram: You make a fair point, it can definitely look a bit like magic if you're not familiar with the framework! I chose this approach because, from my perspective, it was the cleanest solution with the least impact on the code.

The @NotEmpty annotation doesn't throw the 400 error by itself; Spring Boot is what handles that process for us. To put it simply, you can think of @NotEmpty as just a written rule. In this case, the rule states that the request cannot proceed if the tags field is empty. When a request comes in, Spring Boot uses Hibernate Validator to check for these rules. If it finds that the tags field is empty, it enforces the rule and automatically generates the 400 Bad Request error, using the message we provided in the annotation.

Heyvkram:
> @Heyvkram @isaac-diez @acocinas
> 
> By the way, wouldn't it be necessary to add tests?

You are absolutely right, I know it's good practice, but I saw there was already one created for that endpoint which checked that the title was mandatory and couldn't be created with that empty field.

To avoid repeating code, and since I already tested it directly with Postman, I decided not to add another test for this case.

Should I still add a test in this case?

CHANGELOG.md
Heyvkram: My bad! it was a minor change, it's already corrected!

conf/.env.CI.dev
Heyvkram: Also corrected!

CHANGELOG.md
ivilarop: I have an observation about the versioning strategy (SemVer):

- By adding @NotEmpty to the tags field, the API now rejects requests it previously accepted (e.g., tags: []).
- This means that clients consuming the API and sending empty tags will now receive a 400, whereas they were previously valid.
- According to Semantic Versioning, this is a breaking change → it should be considered a MAJOR release.

👉 The difference with PATCH or MINOR:

PATCH: only bugfixes, with no impact on the public contract.

MINOR: new features, but backwards compatible.

MAJOR: breaks compatibility with existing clients.

In this case, although the change is small in code, the impact on the API contract is large → which is why it corresponds to a MAJOR release.

⚖️ Learning from: what determines the version is not the technical complexity of the change, but the impact on API consumers.

conf/.env.CI.dev
ivilarop: see previous comment

itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/ChallengeCreateDto.java
ivilarop: The explanation is half-baked because it doesn't detail exactly why Spring returns the 400. It's worth mentioning:

The annotation itself doesn't return anything.

Spring performs automatic validation thanks to @Valid in the DTO.

The engine (Hibernate Validator) returns violations.

Spring throws an exception (WebExchangeBindException or MethodArgumentNotValidException).

The exception handler translates this into an HTTP 400.

ivilarop:
> > @Heyvkram @acocinas @isaac-diez
> > Should the wiki endpoint description be modified? Or the Postman collection?
> 
> @ivilarop I don't think it's necessary to touch the wiki part, since the endpoint's functionality hasn't changed. On the other hand, the Postman collection will have to be modified because the raw request body is incorrect. As for the Postman part, I created a new task in Taiga for it, and it will be a different PR.

As for the Postman part, I created a new task in Taiga for it, and it will be a different PR. I understand you're referring to 665? Can we note that in the PR description?

ivilarop:
> Should I still add a test in this case?

Yes, you should add an empty tag test even if a similar one already exists for title.

The reason is that each field validation must be explicitly covered. Having a test for title doesn't guarantee that tags will behave the same, because validations or business rules could be changed independently in the future.

Manually testing in Postman is fine for exploration, but it's no substitute for automated tests, which ensure the contract remains stable over time.

If you're worried about duplicating code, you can factor utilities into the test suite (helpers for invalid requests, etc.), but the important thing is that there is a specific test for tags.

CHANGELOG.md
Heyvkram: You're totally right. I didn't realize the change was so important. The way you explain it, it's much clearer to me! It's changed now!

conf/.env.CI.dev
Heyvkram: Changed!

itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/ChallengeCreateDto.java
Heyvkram: You're right. I was trying to give a quick explanation and I forgot to mention the @Valid part of the DTO. The way you explain it is more complete. Noted!

Heyvkram:
> > Should I still add a test in this case?
> 
> Yes, you should add an empty tag test even if a similar one already exists for title.
> 
> The reason is that each field validation must be explicitly covered. Having a test for title doesn't guarantee that tags will behave the same, because validations or business rules could be changed independently in the future.
> 
> Manually testing in Postman is fine for exploration, but it's no substitute for automated tests, which ensure the contract remains stable over time.
> 
> If you're worried about duplicating code, you can factor utilities into the test suite (helpers for invalid requests, etc.), but the important thing is that there is a specific test for tags.

Yes, you're right! I was worried about repeating code, but you're right that it's good practice and necessary to have a test here.

The test is created now!

@Heyvkram 

leave it noted in the PR description

Heyvkram:
> > > @Heyvkram @acocinas @isaac-diez
> > > Should the wiki endpoint description be modified? Or the Postman collection?
> > 
> > 
> > @ivilarop I don't think it's necessary to touch the wiki part, since the endpoint's functionality hasn't changed. On the other hand, the Postman collection will have to be modified because the raw request body is incorrect. As for the Postman part, I created a new task in Taiga for it, and it will be a different PR.
> 
> As for the Postman part, I created a new task in Taiga for it, and it will be a different PR. I understand you're referring to 665? Can we note that in the PR description?

Yes! It's task 665, i added the new PR to the description of this one!

acocinas (approved):
Hey there! 

I've reviewed the latest changes in the PR and everything looks good. All the suggestions from the review round have been addressed, making the code more consistent and easier to maintain.

The implementation is solid, and the tests pass without issues.

isaac-diez (approved):
All the suggested revisions make perfect sense in order to have a clearer and more robust implementation and the issues have been addressed. I see no problem with the tests as well.

CHANGELOG.md
ivilarop: Ok now it seems more accurate to consider it as major.Well done. You can resolve conversation ;)

conf/.env.CI.dev
ivilarop: Well done. You can resolve conversation ;)

itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/ChallengeCreateDto.java
ivilarop: @Heyvkram 

Ok, but all of this should be in the PR description, not buried in the comments.

The PR description is the "official source" that reviewers and peers read later to understand:

- What change was made. 
- How to test it. 
- Why it was made (technical motivation). 
- Impact on the API/consumers (SemVer). 
- (Optional) Relevant technical details → in this case, how validation works with @NotEmpty + @Valid + Hibernate Validator → 400 Bad Request:

One last reflection: the 400 Bad Request is correct and expected.
However, is the error message returned by Spring descriptive enough for API consumers?
For example, when sending tags: [], do we get a clear message like "The tags list cannot be empty" or just a generic "Validation failed" response?

Could you add a Postman screenshot showing the current response?

If the message is not user-friendly, we might consider handling this in GlobalExceptionHandler to return a structured JSON with field + message. This could be addressed as technical debt in a future PR to keep consistency across all validations.

Please note down anything that might be helpful in the PR description. Link technical debt, explain the genetic message, etc.

ivilarop:
> > @Heyvkram @acocinas @isaac-diez
> > Should the wiki endpoint description be modified? Or the Postman collection?
> 
> @ivilarop I don't think it's necessary to touch the wiki part, since the endpoint's functionality hasn't changed. On the other hand, the Postman collection will have to be modified because the raw request body is incorrect. As for the Postman part, I created a new task in Taiga for it, and it will be a different PR.

@Heyvkram  Ok, but add the link to the new task so it's easier to track. In de PRs description

ivilarop:
@Heyvkram  

missing autocheck right?

Heyvkram:
> > > @Heyvkram @acocinas @isaac-diez
> > > Should the wiki endpoint description be modified? Or the Postman collection?
> > 
> > 
> > @ivilarop I don't think it's necessary to touch the wiki part, since the endpoint's functionality hasn't changed. On the other hand, the Postman collection will have to be modified because the raw request body is incorrect. As for the Postman part, I created a new task in Taiga for it, and it will be a different PR.
> 
> @Heyvkram Ok, but add the link to the new task so it's easier to track. In de PRs description

done!

Heyvkram:
> @Heyvkram
> 
> missing autocheck right?

@ivilarop 
Done! I also redo the PR description to be more complete! adding the autocheck and all the previous advices!

itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/ChallengeCreateDto.java
Heyvkram: Done! I updated the PR description with all the info and the screenshot with the postman response!

ivilarop: @Heyvkram 

The message returned by Spring with @NotEmpty is long, technical and confusing (the one you show in Postman with Validation failed for argument ... codes [...] default message [...])

image

what I told you in "If the message is not user-friendly, we might consider handling this in GlobalExceptionHandler to return a structured JSON with field + message. This could be addressed as technical debt in a future PR to maintain consistency across all validations." It is to avoid precisely this. Have we created technical debt to do it later?

Heyvkram: I completely agree! I have already created the technical debt ticket (https://tree.taiga.io/project/luisvi-ita-challenges/issue/701) and have modified the PR description to include it.

ivilarop (approved):

PR #977, opened 2025-10-01
feat(user): add field points to user document(#747)

ToniR90:
### Summary
This PR modifies the *itachallenge-user microservice, preparing the release of version *2.1.0-RELEASE**.  
It introduces a new points field in the UserDocument entity and ensures it is properly integrated and tested.

---

### Changes
- *Model Update:*  
  - Added a new Integer points field to UserDocument with a default value of 0.  
- *Test Fixes:*  
  - Corrected usages of the UserDocument all-args constructor in tests to account for the new field.  
- *New Tests:*  
  - Added dedicated test cases to verify the correct behavior of the points field.  

---

### Motivation
The points field is required to support upcoming features involving scoring, gamification, and user progress tracking. This PR integrates the field safely while ensuring backward compatibility and full test coverage.

---

### Fields changed
In this PR the following files have been changed:
- __CHANGELOG.md__: updated itachallenge-user to 2.1.0 version
- __.env.CI.dev__: - Updated version to match the one specified in CHANGELOG.md
- __UserDocument.java__: added the new 'points' file with appropriate annotations
- __UserControllerTest.java__: - Modified the constructor in getUser_WhenUserExists_Returns200 test to include the new field with value 0
- __UserDocumentTest.java__: Updated several tests to include the new 'points' field. Added a new test to validate that the builder handles only the 'points' field addition.
- __UserServiceImplTest.java__: modified several tests to include the new field 'points'

Other fields were not modified as part of this pull request.

---

### Notes
- Affects only the *itachallenge-user* microservice.  
- No external API changes introduced.  
- All tests updated and passing.

---

### Changelog
Version bumped to **2.1.0-RELEASE**  
`CHANGELOG.md` updated to reflect:
- Model enhancement (`points` field)  
- Test adjustments and additions  

---

### How to Test
Run unit tests in `UserDocumentTest`. Verify:
- ✅ `points` field defaults to `0` when not explicitly set  
- ✅ Field persists and retrieves correctly   

---

### PR Author Self-check
- [x] I tested my changes locally and verified expected behavior  
- [x] The PR has a clear, focused purpose (model evolution for gamification)  
- [x] Title, description, and changelog are complete and accurate  
- [x] No merge conflicts or unrelated changes  
- [x] All automated checks have passed

### DIFF
--- a/itachallenge-user/src/main/java/com/itachallenge/user/document/UserDocument.java
+++ b/itachallenge-user/src/main/java/com/itachallenge/user/document/UserDocument.java
@@ -34,6 +34,10 @@ public class UserDocument {
     @Field("bookmark_challenges")
     private Set<UUID> bookmarkChallenges;
 
+    @Builder.Default
+    @Field
+    private Integer points = 0;
+
     @Override
     public String toString() {
         StringJoiner joiner = new StringJoiner(", ", "UserDocument{", "}");
--- a/itachallenge-user/src/test/java/com/itachallenge/user/controller/UserControllerTest.java
+++ b/itachallenge-user/src/test/java/com/itachallenge/user/controller/UserControllerTest.java
@@ -65,7 +65,7 @@ class UserControllerTest {
     @Test
     void getUser_WhenUserExists_Returns200() {
         String githubUsername = "existingUser";
-        UserDocument expectedUser = new UserDocument(UUID.randomUUID(), githubUsername, Role.ADMIN, null, null, null);
+        UserDocument expectedUser = new UserDocument(UUID.randomUUID(), githubUsername, Role.ADMIN, null, null, 0);
         when(userService.getUser(githubUsername)).thenReturn(Mono.just(expectedUser));
 
         webTestClient.get()
--- a/itachallenge-user/src/test/java/com/itachallenge/user/controller/UserControllerTest.java
+++ b/itachallenge-user/src/test/java/com/itachallenge/user/controller/UserControllerTest.java
@@ -65,7 +65,7 @@ class UserControllerTest {
     @Test
     void getUser_WhenUserExists_Returns200() {
         String githubUsername = "existingUser";
-        UserDocument expectedUser = new UserDocument(UUID.randomUUID(), githubUsername, Role.ADMIN, null, null);
+        UserDocument expectedUser = new UserDocument(UUID.randomUUID(), githubUsername, Role.ADMIN, null, null, null);
         when(userService.getUser(githubUsername)).thenReturn(Mono.just(expectedUser));
 
         webTestClient.get()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,13 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### [itachallenge-user-2.1.0-RELEASE] - 2025-09-12
+
+### Added
+- Added a new Integer points field to UserDocument with a default value of 0.
+- Corrected usages of the UserDocument all-args constructor in tests to account for the new field.
+- Added dedicated test cases to verify the correct behavior of the points field, including default initialization.
+
 ### [itachallenge-challenge-3.0.0-RELEASE] - 2025-09-22
 
 ### Added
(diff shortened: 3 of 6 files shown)


MarTorrijos (changes requested):
Good work! Just a couple of notes:

- For future branches, do not use a # before the task number. According to the [GUIDELINES.md](https://github.com/IT-Academy-BCN/ita-challenges-backend/blob/develop/GUIDELINES.md) this is the proper naming:

	  git checkout -b feature/123-AddUserAuthentication

- The changelog is okay, but since you are adding a field it would be more appropriate to use the tag '### Added' instead of '### Change'. More info [here ](https://tree.taiga.io/project/luisvi-ita-challenges/wiki/guidehow-to-update-changelog-version)

Aside from that I think the field you added and also the tests are alright

ToniR90:
Applied the requested changes from the review from @MarTorrijos

MarTorrijos (approved):

zohra-b (approved):
PR approved. 

It would have been better to make one commit per each changed file : the third commit (
commit b4158a243ff1784297b9544adc7f0a249edea8c3) is quite long and concerns 3 different files.

MarTorrijos:
@ToniR90 
I just realized that this PR doesn't have a PR author self-check and all PRs should have them
https://tree.taiga.io/project/luisvi-ita-challenges/wiki/pr-review-checklist

Include it on the description, please

ToniR90:
> @ToniR90 I just realized that this PR doesn't have a PR author self-check and all PRs should have them https://tree.taiga.io/project/luisvi-ita-challenges/wiki/pr-review-checklist
> 
> Include it on the description, please

@MarTorrijos , done! Thanks

ivilarop:
@ToniR90  @zohra-b  @MarTorrijos  Please add an entry in the CHANGELOG.md reflecting this change.
Since we are introducing a new field (points) in the UserDocument, this counts as a functional addition


.env.CI.dev is it necessary?review other PRs to see if necessary

itachallenge-user/src/test/java/com/itachallenge/user/controller/UserControllerTest.java
```diff
@@ -65,7 +65,7 @@ void testEndpoint_ShouldReturnHelloMessage() {
     @Test
     void getUser_WhenUserExists_Returns200() {
         String githubUsername = "existingUser";
-        UserDocument expectedUser = new UserDocument(UUID.randomUUID(), githubUsername, Role.ADMIN, null, null);
+        UserDocument expectedUser = new UserDocument(UUID.randomUUID(), githubUsername, Role.ADMIN, null, null, null);
```
ivilarop: There are two ways to approach this:

1. The ideal solution

Would be to add a builder with Lombok (@Builder).
That way, older tests that don't need points could construct the object without worrying about that field (the builder would assign 0 by default). This avoids having to pass unnecessary values ​​in constructors and gives us flexibility for the future.

Disadvantage: To do this properly, we would have to refactor all the tests that currently create instances with constructors and migrate them to the builder → it's a pretty big change.

2. The practical solution (what we'll do now)

We declare points as int (0 by default) and in the older tests, we simply replace nulls with 0.
This is the fastest way to keep the tests stable without having to rewrite them all.

What you should add

New, specific tests that test the behavior of points:

Make it 0 by default.

Make it possible to correctly assign a different value.

ivilarop (changes requested):

itachallenge-user/src/main/java/com/itachallenge/user/document/UserDocument.java
```diff
@@ -34,6 +34,10 @@ public class UserDocument {
     @Field("bookmark_challenges")
     private Set<UUID> bookmarkChallenges;
 
+    @Builder.Default
+    @Field
+    private Integer points = 0;
```
ivilarop, line 39: If the concept of "null points" doesn't exist in our domain, the best solution is to declare it as an int.

**Advantages**

- Always 0 by default → no need for @Builder.Default or manual initialization.
- Avoids NPE when performing comparisons or additions.
- More semantic: 0 means "no points," not "unknown value."

**When to keep Integer**

- Only if you really need to:
- Differentiate between "unreported field" (null) and "reported field with value 0."
- Use frameworks that require a wrapper (Integer) to allow null (e.g., optional validations or partial persistence).

ToniR90:
@zohra-b @MarTorrijos 
Could you check the changes?
https://github.com/IT-Academy-BCN/ita-challenges-backend/pull/977#discussion_r2413628334

ivilarop (approved):

PR #986, opened 2025-10-03
refactor(challenge): replace @Autowired annotations with @RequiredArgsConstructor (#718)

MarTorrijos:
## Summary

This PR addresses [task 718](https://tree.taiga.io/project/luisvi-ita-challenges/task/718)

It comes from [the technical debt of this task](https://tree.taiga.io/project/luisvi-ita-challenges/issue/641), which this PR addresses specifically for the challenge microservice. The rest of the microservices, as well as the test refactors, are out of scope for this PR.

The goal is to remove all @Autowired annotations from ChallengeServiceImpl.java, from the Challenge microservice), and to use constructor injection via Lombok’s @RequiredArgsConstructor.

### Changes

- ChallengeServiceImpl:
   - Removed 7 @Autowired annotations and used @RequiredArgsConstructor instead.
   - Marked the dependencies as final
   - Renamed ILanguageService to iLanguageService to follow correct naming conventions. Otherwise SonarQube analysis was giving an error
- ChallengeServiceImplCacheTest:
   - Removed unused imports
   - Changed tests a little bit because otherwise they wouldn't work with the @RequiredArgsConstructor from the service.
      - Added .cache() to mocked Mono/Flux returns. With @RequiredArgsConstructor, Spring proxies the service. Reactive streams get re-subscribed on each call, which breaks caching in tests. .cache() makes the mock replayable for multiple subscriptions.
      - Replaced verify(..., times(1)) with verify(..., atLeastOnce())
     Constructor injection with Spring proxies can cause extra internal calls. Using atLeastOnce() avoids false test failures.
	  - Removed verifyNoMoreInteractions(...). Extra calls from Spring proxies or constructor-based wiring would trigger this and fail tests unnecessarily.
	  
	  
### Some context of why the tests were slightly changed:

Switching from @Autowired to @RequiredArgsConstructor makes Spring wrap the service in a proxy for things like caching. This can trigger extra calls or re-subscriptions on reactive streams, which breaks the original test mocks. We fix it by making the mocked Monos/Fluxes “hot” with .cache(), so repeated calls return the same values.	  

### Justification for the Refactor

- Improves code quality and testability: Dependencies are injected via the constructor, making unit testing easier without needing to load the full Spring context.

- Enhances readability: Dependencies are explicitly declared and made mandatory.

- Promotes immutability: Dependencies can now be marked as final, enforcing best practices.

### Notes

- No new tests were needed but had to change the existing ones a little.

- Following [Taiga's guidelines](https://tree.taiga.io/project/luisvi-ita-challenges/wiki/guidehow-to-update-changelog-version),the changelog was not updated since this is a refactor and doesn't require it.

### PR author self-check

- [x] I tested my changes locally and verified they work as expected
- [x] The PR has a clear and focused purpose
- [x] Title and description are filled in correctly. No changelog edit needed since it's a refactor
- [x] There are no leftover merge conflicts or unrelated changes from previous branches
- [x] All automated checks (like SonarCloud) have passed
- [x] All existing tests pass and no new ones were needed

### DIFF
--- a/itachallenge-challenge/src/main/java/com/itachallenge/challenge/service/ChallengeServiceImpl.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/challenge/service/ChallengeServiceImpl.java
@@ -17,6 +17,7 @@ import io.micrometer.common.util.StringUtils;
 import lombok.RequiredArgsConstructor;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.cache.annotation.CacheEvict;
 import org.springframework.cache.annotation.Cacheable;
 import org.springframework.stereotype.Service;
@@ -48,8 +49,8 @@ public class ChallengeServiceImpl implements IChallengeService {
     private final ChallengeRepository challengeRepository;
     private final ILanguageService iLanguageService;
     private final SolutionRepository solutionRepository;
-    private final DocumentToDtoConverter<ChallengeDocument, ChallengeDto> challengeConverter;
-    private final DocumentToDtoConverter<SolutionDocument, SolutionDto> solutionConverter;
+    private final DocumentToDtoConverter<ChallengeDocument, ChallengeDto> challengeConverter = new DocumentToDtoConverter<>();
+    private final DocumentToDtoConverter<SolutionDocument, SolutionDto> solutionConverter = new DocumentToDtoConverter<>();
     private final IUserService userService;
     private final ITagService tagService;
 
--- a/itachallenge-challenge/src/main/java/com/itachallenge/challenge/service/ChallengeServiceImpl.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/challenge/service/ChallengeServiceImpl.java
@@ -49,8 +49,8 @@ public class ChallengeServiceImpl implements IChallengeService {
     private final ChallengeRepository challengeRepository;
     private final ILanguageService iLanguageService;
     private final SolutionRepository solutionRepository;
-    private final DocumentToDtoConverter<ChallengeDocument, ChallengeDto> challengeConverter = new DocumentToDtoConverter<>();
-    private final DocumentToDtoConverter<SolutionDocument, SolutionDto> solutionConverter = new DocumentToDtoConverter<>();
+    private final DocumentToDtoConverter<ChallengeDocument, ChallengeDto> challengeConverter;
+    private final DocumentToDtoConverter<SolutionDocument, SolutionDto> solutionConverter;
     private final IUserService userService;
     private final ITagService tagService;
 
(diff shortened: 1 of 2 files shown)


zohra-b (approved):
Explanations are clear, changes are justified.

ToniR90 (approved):
Well-documented changes and a solid implementation.

ivilarop:
@MarTorrijos @zohra-b @ToniR90 

The name of the PR is a bit generic, isn't it?

itachallenge-challenge/src/main/java/com/itachallenge/challenge/service/ChallengeServiceImpl.java
```diff
@@ -14,6 +14,7 @@
 import com.itachallenge.challenge.repository.ChallengeRepository;
 import com.itachallenge.challenge.repository.SolutionRepository;
 import io.micrometer.common.util.StringUtils;
+import lombok.RequiredArgsConstructor;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
```
ivilarop, line 20: should not be deleted?

itachallenge-challenge/src/test/java/com/itachallenge/challenge/service/ChallengeServiceImplCacheTest.java
ivilarop, line 1: Note (non-blocking, for future PRs):
Thanks for documenting the changes in tests. However, it's important to note that moving from @Autowired to constructor injection doesn't introduce proxies per se; proxies appear through annotations like @Cacheable, @Transactional, @Validated, etc.
If there were indeed failures due to additional calls, it would be helpful to reference the exact test/stack and the annotation causing the proxy.

In general, for a DI refactor, the following are expected:
• Minimal adjustments to tests (construction of the SUT via constructor / @InjectMocks and updating imports via renames).
• Maintain the same behavioral assertions; there's no need to relax times(1) to atLeastOnce or remove verifyNoMoreInteractions unless the contract changes (in which case it's a good idea to separate it into another PR).
• The use of .cache() should be limited to mocks in tests and documented; This shouldn't be necessary if the test doesn't rely on internal call counters.

Also: when you detect a risk of going out of scope, please mention it first in the PR before applying changes to avoid increasing complexity. Propose options (revert, separate into another PR/commit, or open a technical debt) and explain the impact. With the time available, it's difficult to validate whether all these adjustments were essential, and we prefer to avoid introducing changes that would later require another round of review.

We approved this PR so as not to slow down the flow, but going forward, let's try to narrow the scope of the refactoring and separate DI changes from test strategy changes (or accurately document the technical cause when necessary). Thanks 🙌

ivilarop (changes requested):

Sh1ng0:
iLanguageService should be kept "ILanguageService" I think

zohra-b (approved):

ivilarop (approved):

PR #996, opened 2025-10-06
PR 1/4 - feat(challenge): create and test FieldErrorDto and APIErrorResponse dto (#731)

aurelien-darbellay:
## Summary

This PR matches **Taiga task 731** and introduces a **standardized error response format** for the Challenge microservice when the payload submitted by the front-end fails to meet the validation criteria (triggering _MethodArgumentNotValidException_), starting the error-handling consistency initiative started in user story 712.

This is the first of a group of 4 PRs: (Task 731, #996); (Task 734, #997); (Task 770, #998); (Task 756, #999)

---

## Checklist Alignment
✅ Introduces `APIErrorResponse` and `FieldErrorDto` to formalize error structure  
✅ Adds serialization-focused unit tests for consistent JSON output  

---

### Before
-No DTOs to handle MethodArgumentNotValidExceptions. 

### After
- **New DTOs**  
  - `APIErrorResponse`: encapsulates status, error, message, and a list of field-specific errors.  
  - `FieldErrorDto`: details which object and field failed, with a human-friendly message.   
- **Serialization Tests:**  
  Added dedicated unit tests ensuring predictable JSON serialization, including:  
  - Multiple field errors serialization  
  - Handling of empty or null error lists  
  - Correct metadata field inclusion  

---

### `APIErrorResponse`
```java
@JsonInclude(JsonInclude.Include.NON_EMPTY)
@AllArgsConstructor
@Getter
@Builder
public class APIErrorResponse{
    private final Instant timestamp;
    private final int status;
    private final String error;
    private final String message;
    private final String path;
    private final List<FieldErrorDto> errors;
}
```

### `FieldErrorDto`
```java
@JsonInclude(JsonInclude.Include.NON_EMPTY)
@Builder
@AllArgsConstructor
@Getter
public class FieldErrorDto {
    private final String objectName;
    private final String field;
    private final String message;
}
```
---

## Changelog

As a feat, this task require a version bump. Still, for simplicity, I suggest we bundle all the work done in the 4 PRs into one version bump, that will be performed in the PR #999.

---

## How to Test

Run the test suites in `APIErrorResponseTest` and `ErrorFieldDto`to verify serialization correctness:

✅ Structured error JSON includes status, message, and detailed field errors  
✅ Empty error lists are handled according to `@JsonInclude.NON_EMPTY` behavior  
✅ Valid JSON structure confirmed in all test cases  

---

## PR Author Self-check

- [x] I tested my changes locally and verified they work as expected  
[...]

PR #997, opened 2025-10-06
PR 1/3 - chore(challenge): configure DTO error messages in message.properties & rebase with error-core (#734 & #797)

aurelien-darbellay:
### Summary
This PR corresponds to **Task 734** and introduces the initial **`messages.properties`** file in the `itachallenge-challenge` microservice.  

Initially, it built on  **Task 731** realized in PR #996. **But after a change of strategy,** it is now building on task #[796](error-response-core/src/test/java/com/itchallenge/errorcore/exceptionhandler/BaseExceptionHandlerUnitTest.java), realized in PR #1010. Accordingly, the branch feature/#734_configure_error_messages_in_message.properties has been rebased on feature/add-error-core. 

The goal of this task is to centralize all validation and error messages in one location to prepare the system for future internationalization and consistent message management across DTOs and controllers.

This is the first of a group of 3 PRs:  (Task 734, #997); (Task 770, #998); (Task 799), aiming at refactoring error handling in challenge-service by appealing to the error-response-core module, developped in task #[796](error-response-core/src/test/java/com/itchallenge/errorcore/exceptionhandler/BaseExceptionHandlerUnitTest.java).
As such, it introduces the error-response-core module as a dependency in itachallenge-challenge service.

### Scope observation
In this PR we focus exclusively on VALIDATION EXCEPTIONS LINKED TO DTO, that can be trown by the system, AS IT IS NOW. Some DTOs should have added validation, and some methods that do not throw validation exception now should be enhanced to do so.  
For instance, the following method should have a "Valid" annotation before its argument - which would allow it to throw a MethodArgumentNotValidException: 

Captura de pantalla 2025-10-10 145959

But such modifications were'nt done in this PR. When realizing these improvements, the messages.properties file will need to be enriched accordingly.  

Translations into Castillan and Catalan have not been included in the messages.properties file. This could be an interesting enhancement, to implement when realizing the general strategy. 

#### Before
- No messages.properties file existed
- error-response-core not included as a dependency. 

#### After
- Added a **`messages.properties`** file under `src/main/resources/`.  
- Organized messages into sections per DTO with clear and consistent naming conventions (e.g. `solution.text.notEmpty`, `challenge.title.notEmpty`, `resource.id.notNull`).  
- Included top-level `*.dto.invalid` entries for general validation fallback messages, to be used by the global error handler in future tasks.
-  added: implementation project (":error-response-core") to build.gradle

Example content:
```properties
# SolutionDto Validation Messages
solution.text.notEmpty=The solution text cannot be empty.
solution.languageId.invalid=The language ID must be a valid UUID.
solution.challengeId.invalid=The challenge ID must be a valid UUID.
solution.dto.invalid=The provided solution data is invalid.
```

### Checklist Alignment
✅ Introduces a centralized `messages.properties` file  
✅ Defines clear and consistent message keys per DTO  

### Impact for API Consumers
- No changes to API behavior or responses yet.  
- This PR only affects message organization at the project level.  
- Future releases will leverage these keys to return localized error responses.  

### Changelog

Since there is no behaviour changes, there is no version bump. The whole PR chain will be bundled into one bump in the end.

### How to Test
1. Verify that the file `src/main/resources/messages.properties` is included in the build.  
2. Confirm that message keys follow the naming structure:  
   - `solution.*`  
   - `challenge.*`  
   - `resource.*`  
3. Confirm that the classes from error-response-core are available in itachallenge-challenge. 

### PR Author Self-Check
- [x] I tested that the application builds and runs with the new `messages.properties` file  
- [x] The PR focuses solely on message centralization (no code or logic changes)  
- [x] No unrelated commits or merge conflicts  
- [x] SonarLint shows no warnings or errors  

### Future Considerations
1. **Update DTO annotations** to reference message keys instead of inline text (e.g. `@NotEmpty(message = "{challenge.title.notEmpty}")`) - task 770
2. **Integrate message resolution** into the `GlobalExceptionHandler` for localized validation responses - task 756 
3. **Extend the same strategy** to handle constraint-based validations (e.g. `ConstraintViolationException` for `@RequestParam`, `@PathVariable`, `@RequestHeader`).  
4. **Review all DTOs and Controller's method in ita-challenge to ensure validation consistency
5. **Establish a general validation policy** to ensure consistency across all microservices.  
6. **Translations**: in the general validation policy, we might want to include translations into Castillan and Catalan.

### DIFF
--- a/itachallenge-challenge/build.gradle
+++ b/itachallenge-challenge/build.gradle
@@ -60,6 +60,8 @@ dependencies {
 
     implementation project(":jwt-core")
 
+    implementation project (":error-response-core")
+
     testImplementation 'org.springframework.boot:spring-boot-starter-test:3.0.6'
     testImplementation 'org.springframework:spring-test:5.3.13'
     testImplementation 'org.junit.jupiter:junit-jupiter'
--- a/itachallenge-challenge/src/main/resources/messages.properties
+++ b/itachallenge-challenge/src/main/resources/messages.properties
@@ -1,0 +1,31 @@
+# ===============================
+# SolutionDto Validation Messages
+# ===============================
+solution.text.notEmpty=The solution text cannot be empty.
+solution.languageId.invalid=The language ID must be a valid UUID.
+solution.challengeId.invalid=The challenge ID must be a valid UUID.
+solution.text.size=The solution text must be between 1 and 500 characters long.
+solution.dto.invalid=The provided solution data is invalid.
+# ===============================
+# ChallengeCreateDto Validation Messages
+# ===============================
+challenge.title.notEmpty=The challenge title cannot be empty.
+challenge.description.notEmpty=The description cannot be empty.
+challenge.language.notEmpty=The language field cannot be empty.
+challenge.solution.notEmpty=The solution text cannot be empty.
+challenge.topic.notNull=The topic must be provided.
+challenge.tags.notEmpty=At least one tag must be provided for the challenge.
+challenge.dto.invalid=The provided challenge data is invalid or incomplete.
+# ===============================
+# ResourceDto Validation Messages
+# ===============================
+resource.id.notNull=The resource ID cannot be null.
+resource.title.notEmpty=The resource title cannot be empty.
+resource.description.notEmpty=The resource description cannot be empty.
+resource.url.notEmpty=The resource URL cannot be empty.
+resource.topic.notNull=The topic must be specified.
+resource.contentType.notNull=The content type must be specified.
+resource.challengeIds.notNull=The list of associated challenge IDs cannot be null.
+resource.associationType.notNull=The association type must be specified.
+resource.dto.invalid=The provided resource data is invalid or incomplete.
+
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,13 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+[itachallenge-challenge-3.0.1-RELEASE] - 2025-11-11
+
+### Changed
+- Added dependency to error-response module in challenge service / created message.properties files to store i18n messages. (Taiga [#734] & [#797], PR [#997])
+  - introduced external validation messages;
+  - does not break APIs, but may alter error texts.
+
 ### [itachallenge-user-3.0.2-RELEASE] - 2025-11-05
 
 ### Added
--- a/conf/.env.CI.dev
+++ b/conf/.env.CI.dev
@@ -5,7 +5,7 @@ MICROSERVICE_DEPLOY_itachallenge-user=itachallenge-user
 MICROSERVICE_VERSION_itachallenge-user=3.0.2-RELEASE
 
 MICROSERVICE_DEPLOY_itachallenge-challenge=itachallenge-challenge
-MICROSERVICE_VERSION_itachallenge-challenge=3.0.0-RELEASE
+MICROSERVICE_VERSION_itachallenge-challenge=3.0.1-RELEASE
 
 MICROSERVICE_DEPLOY_itachallenge-document=itachallenge-document
 MICROSERVICE_VERSION_itachallenge-document=1.0.2-RELEASE


MarTorrijos (changes requested):
The comments I left on PR 1/4 still apply to this one. Since this is part of a chain of PRs, I'm not entirely sure if the feedback from that one should also be addressed here before approving it, but I assume it should.

messages.properties has clear and well-structured messages, so no feedback there.

Good work!

aurelien-darbellay:
@Mar Torrijos: thanks again for the review, please check my comments on your comments to the previous PR.

ToniR90 (approved):
The file message.properties is well structured, assuming the request changes on PR [996](https://github.com/IT-Academy-BCN/ita-challenges-backend/pull/996) everything looks fine!!

MarTorrijos (approved):
Cool! I'll approve this PR then since the file that changes here on 2/4 is okay

Meriekje (changes requested):
On `messages.properties`, I suggest to use "text" or not for both contraints as the message is identical:
Line 4: solution.text.notEmpty=The solution text cannot be empty
Line 15: empty.challenge.solution.notEmpty=The solution text cannot be empty.

Although it's not part of this PR, `FieldErrorDtoTest` here has several messages in Spanish while they are in English in PR996 as they must be.

There is also the changelog issue as Mar pointed although in 2/4 would not be needed at all because it's purely internal. You can add it as not needed here.

Meriekje (approved):

aurelien-darbellay:
@Meriekje --- 
(1) I left message.properties as it is because it reflects the structure of existing Dtos. 
(2) I rebased on 1/4 to get rid of the spanish messages. 
(3) as we discussed, the changelog conversation is good as it is.

ToniR90 (approved):
Considering the scope of the PR, the changes are well implemented and the messages are clea

itachallenge-challenge/build.gradle
```diff
@@ -60,6 +60,8 @@ dependencies {
 
     implementation project(":jwt-core")
 
+    implementation project (":error-response-core")
```
ivilarop, line 63: CHANGELOG (required): Documents that the challenge module:

adds a dependency to error-response-core;

introduces external validation messages;

does not break APIs, but may alter error texts.

itachallenge-challenge/src/main/resources/messages.properties
ivilarop, line 1: Could we rename the messages.properties file to error-core-messages.properties to explicitly indicate that it's the core's global properties file? This way, in the technical request you previously created regarding the starter creation, we can add the task: "Move the message configuration to the core's boot-starter so that microservices can consume it without manual configuration."

ivilarop (changes requested):

itachallenge-challenge/build.gradle
aurelien-darbellay: Done. Version proposed: 3.0.1

itachallenge-challenge/src/main/resources/messages.properties
aurelien-darbellay: I would not do that - first the files does belong to the challenge service, not the error-core: MessageSource work like this - it can combine local and root *.properties files: it first tries to find the key in the local file, then if not found, search the root file. In the present configuration, the messageSource bean is configured to search for a messages.properties file in the service, and then if the key isn't found, search the core-messages.properites which is in the error-response-core module. Here is my last exchange with the AI on the topic:

**Me**:  I'm sorry, but I'm struggling to see why it is good. i already have a core-messages.properties file in the core - that fix framework level messages. then, I have local files that provide messages for service-level exceptions. the bean in the core/starter is configured so as to read both. 

**AI**: That’s actually a very good question — and you’re not wrong to be skeptical.
In fact, from what you’re describing, your current setup already follows a clean, layered design, and the proposed rename might even reduce clarity, not improve it.
Let’s reason this out carefully and pragmatically 👇

🧩 Your current model (from what you just described)
In the core

You already have core-messages.properties (or error-core-messages.properties — name doesn’t matter much).

It provides framework-level / generic messages, like:

error.notFound={0} not found
error.badRequest=Invalid request

This file lives inside the core, and the ErrorResponseBuilder / starter knows to load it.

In each service
You have a local messages.properties file that defines domain-specific messages, e.g.:

user.notFound=User not found
user.invalidEmail=The provided email is invalid

The bean configured in the starter is already set up to chain / aggregate both:

The core message bundle (core-messages)

The service-local bundle (messages)

That’s exactly how MessageSource chaining in Spring is meant to work.

ivilarop: Ok understood! thks

itachallenge-challenge/build.gradle
ivilarop: ok thsk!

ivilarop (approved):

PR #998, opened 2025-10-06
PR 2/3 - refactor(challenge): update validation annotations in dtos (#770 & #798)

aurelien-darbellay:
### Summary
This PR implements **Task 770 – Update validation annotations to use message keys from `messages.properties`** as part of the broader User Story *“Improve Error Messages in the Challenge Microservice.”*

It builds on task 796 (PR #1010 ) and task 734 (PR #997). Task 734 previously built on a Task 731, but the overall strategy for error handling has been redesigned, and the current git pipeline rebased on feature/add-error-core which develops a error handling module that can be used in all microservices. 

It updates validation annotations in the `itachallenge-challenge` microservice so that they now reference message keys defined in `messages.properties`, replacing hardcoded or inline message strings.  
This ensures consistent, maintainable, and easily localizable validation feedback across all DTOs.
Only the DTOs that already include validation annotations have been updated. For consistency, the whole microservice should be reviewed so that the all DTOs that deserve validation are validated. 

This is the third of a group of 4 PRs: (Task 731, #996); (Task 734, #997); (Task 770, #998); (Task 756, #999)

### Checklist Alignment
✅ Replaces all hardcoded validation messages with message keys from `messages.properties`  
✅ Ensures message keys exist and are correctly defined in the centralized message file  
✅ Maintains existing validation behavior in controllers using `@Valid`  
✅ Adds self-contained unit tests validating message resolution and constraint behavior  

### Context & Changes

**Before**
Validation annotations used inline message text, e.g.:
```java
@NotEmpty(message = "cannot be empty")
@NotNull(message = "cannot be null")
```
### After

Validation annotations now reference centralized message keys:

```java
@NotEmpty(message = "{challenge.title.notEmpty}")
@NotNull(message = "{resource.id.notNull}")
```
### Scope of Changes

**Updated DTOs:**
- `SolutionDto`
- `ChallengeCreateDto`
- `ResourceDto`

**Updated Tests:**
- `SolutionDto` (created)
- `ChallengeCreateDto`(created)
- `ResourceDto`(enriched)
- `ChallengeControllerTest` (modification to adapt to new messages in SolutionDto)

### Validation Behavior

DTO-level unit tests confirm that message interpolation works correctly **outside the Spring context**.

---

### Impact for API Consumers

✅ No breaking changes in endpoint structure or payloads.  
✅ Only the source of validation messages changed — responses remain semantically identical.  
[...]

PR #999, opened 2025-10-06
PR 4/4 - feat(challenge): adjust exception handler to use error dtos and messages.properties (#756)

aurelien-darbellay:
### **Summary**
This PR implements Task 756 – Adjust Exception Handler to use message keys from messages.properties, APIErrorResponse and ErrorFieldDto in the challenge microservice GlobalExceptionHandler as part of the broader User Story 712 “Modify GlobalExceptionHandler in the Challenge microservice to return a JSON with messages defined in messages.properties.”
The scope of the User Story has been narrowed down to deal only with _MethodArgumentNotValidException_.

It builds on task 731 (PR #996), task 734 (PR #997) and task 770 (PR #998).

---

### **Context**
This PR addresses the ongoing effort to standardize and validate request input handling across the **itachallenge-challenge** microservice.

The goal is to:
- Centralize validation error processing in the `GlobalExceptionHandler`
- Ensure consistent in message resolution through `MessageSource`
- Enrich the Error Responses
---
### **Changes**
- ✅ Updated `GlobalExceptionHandler`:
    - Dependency Injection of MessageSource bean  
    - Modification of the handleMethodArgumentNotValidException signature to include a HttpServletRequest argument to extract path
    - Creation of the buildValidationErrorResponse private method that returns and APIErrorResponse on the basis of an objectName, a list of FieldErrorsDto and an HttpServletRequest
    - Modification of handleMethodArgumentNotValidException so that ObjectName and FieldErrors are extracted from the _MethodArgumentNotValidException_, mapped to Dtos, and then passed to buildValidationErrorResponse to generate the apropriate APIErrorResponse  
- ✅ Updated `GlobalExceptionHandlerTest` to adjust to the modifications in the previous point
- ✅ Created `GlobalExceptionHandlerValidationTest` :
    - to unit test the new logic
    - to avoid making the GlobalExceptionHandlerTest even bigger and more confused 
- ✅ Created `ValidationIntegrationTest` to verify:
  - Validation of `ChallengeController` and `ResourceController`
  - The MessageSource bean works as expected in the application context
---
### 🧩 Standardized Validation Error Response

As a result of these changes, the error response returned in case of _MethodArgumentNotValidException_ has now the following structure:

```json
{
  "timestamp": "2025-10-06T22:51:58.293353800Z",
  "status": 400,
  "error": "Bad Request",
  "message": "Validation failed for one or more fields in challengeCreateDto.",
  "path": "/itachallenge/api/v1/challenge/challenges",
  "errors": [
    {
      "objectName": "challengeCreateDto",
      "field": "tags",
      "message": "At least one tag must be provided for the challenge."
    },
    {
      "objectName": "challengeCreateDto",
      "field": "topic",
      "message": "The topic must be provided."
    },
    {
      "objectName": "challengeCreateDto",
      "field": "challengeTitle",
      "message": "The challenge title cannot be empty."
    }
  ]
}
```

### **Justification / Rationale**
This refactor consolidates all field validation error handling into a single, reusable layer.  
By delegating to `GlobalExceptionHandler`, we ensure:
- Consistent HTTP 400 responses across all controllers  
- Readable, translatable error messages for client-side handling  
- Easier maintenance and testability of validation logic  
The integration test suite verifies that validation works correctly under WebFlux reactive conditions.
---

### **Testing**
- ✅ Unit tests for `GlobalExceptionHandler` (`GlobalExceptionHandlerValidationTest`)
- ✅ Integration tests for `ChallengeController` and `ResourceController` validation (`ValidationIntegrationTest`)
- ✅ Local execution: `./gradlew test` passed successfully  
- ✅ SonarLint run locally — no warnings  
- ✅ Manual Integration Testing via Postman 

---

### **Version Impact**
This change applies to the **`itachallenge-challenge`** microservice.  
Proposed version increment: **3.1.0-RELEASE**
Both Changelog and env.CI.dev have been updated.

---

### **Future Work**
- 🔄 Extend error validation structured response to ConstraintValidationExceptions
- 🌐Refactor all Dtos in the challenge microservice to introduce consistency
- Document a general strategy for validation error response to be implemented across all microservices
---

### **Author Self-check**
- [x] I tested my changes locally and verified they work as expected  
- [x] The PR has a clear and focused purpose (single feature/fix/refactor)  
- [x] Title, description, and changelog are correctly filled  
- [x] All automated checks (SonarCloud, CI) pass  
- [x] I responded to all previous review comments  
- [x] No unrelated commits or merge leftovers remain  

---

### DIFF
--- a/itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/APIErrorResponse.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/APIErrorResponse.java
@@ -5,15 +5,18 @@ import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Getter;
 
+import java.time.Instant;
 import java.util.List;
 
 @JsonInclude(JsonInclude.Include.NON_EMPTY)
 @AllArgsConstructor
 @Getter
 @Builder
-public class ErrorResponseDto {
-    private int status;
-    String error;
-    String message;
+public class APIErrorResponse {
+    private final Instant timestamp;
+    private final int status;
+    private final String error;
+    private final String message;
+    private final String path;
     private final List<FieldErrorDto> errors;
 }
--- a/itachallenge-challenge/src/main/java/com/itachallenge/challenge/exception/GlobalExceptionHandler.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/challenge/exception/GlobalExceptionHandler.java
@@ -157,7 +157,7 @@ public class GlobalExceptionHandler {
 
         String message = String.format(
                 "Validation failed for one or more fields in %s.",
-                objectName.replace("Dto", "").toLowerCase()
+                objectName
         );
 
         return APIErrorResponse.builder()
--- a/itachallenge-challenge/src/main/java/com/itachallenge/challenge/exception/GlobalExceptionHandler.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/challenge/exception/GlobalExceptionHandler.java
@@ -99,23 +99,11 @@ public class GlobalExceptionHandler {
                         .message(messageSource.getMessage(error, locale))
                         .build())
                 .toList();
-
+        
         // Build unified structured response
         String objectName = ex.getBindingResult().getObjectName();
-        String message = String.format(
-                "Validation failed for one or more fields in %s.",
-                objectName
-        );
-        APIErrorResponse response = APIErrorResponse.builder()
-                .status(HttpStatus.BAD_REQUEST.value())
-                .error(HttpStatus.BAD_REQUEST.getReasonPhrase())
-                .message(message)
-                .errors(fieldErrors)
-                .timestamp(Instant.now())
-                .path(request.getRequestURI())
-                .build();
-
-        return ResponseEntity.badRequest().body(response);
+        return ResponseEntity.badRequest()
+                .body(buildValidationErrorResponse(objectName, fieldErrors, request));
     }
 
     @ExceptionHandler(BadUUIDException.class)
@@ -160,5 +148,25 @@ public class GlobalExceptionHandler {
             }
         }
         return Optional.empty();
+    }
+
+    private APIErrorResponse buildValidationErrorResponse(
+            String objectName,
+            List<FieldErrorDto> fieldErrors,
+            HttpServletRequest request) {
+
+        String message = String.format(
+                "Validation failed for one or more fields in %s.",
+                objectName.replace("Dto", "").toLowerCase()
+        );
+
+        return APIErrorResponse.builder()
+                .status(HttpStatus.BAD_REQUEST.value())
+                .error(HttpStatus.BAD_REQUEST.getReasonPhrase())
+                .message(message)
+                .errors(fieldErrors)
+                .timestamp(Instant.now())
+                .path(request.getRequestURI())
+                .build();
     }
 }
--- a/itachallenge-challenge/src/main/java/com/itachallenge/challenge/exception/GlobalExceptionHandler.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/challenge/exception/GlobalExceptionHandler.java
@@ -1,28 +1,36 @@
 package com.itachallenge.challenge.exception;
 
+import com.fasterxml.jackson.databind.JsonMappingException.Reference;
 import com.fasterxml.jackson.databind.exc.InvalidFormatException;
+import com.itachallenge.challenge.dto.APIErrorResponse;
+import com.itachallenge.challenge.dto.FieldErrorDto;
 import com.itachallenge.challenge.dto.MessageDto;
+import jakarta.servlet.http.HttpServletRequest;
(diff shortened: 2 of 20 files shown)


zohra-b (changes requested):
**In GlobalExceptionHandlerValidationTest.java** : 
- why don't you use the DisplayName annotation as you used to do in other tests files you created ? it is a good prasctice that should be extended to the whole project
- you could add edge tests (empty fieldErrors or null bindingResults)


**In ValidationIntegrationTest.java** : 
- You could add more edge tests (multiple errors, invalid formats)

zohra-b (approved):
Changes made as requested :
- @DisplayName added to tests to harmonize with other test classes
- 2 edge tests added.

Approved

ToniR90 (approved):
Great work on this PR — I don’t see anything I could add from a code perspective, it’s really well done. I’m not sure how commented lines are handled or whether they might be considered by the mentors (as @rioacosta mentioned). If I had to suggest something, maybe consider adding an errorCode field to the ApiErrorResponse to support debugging and frontend handling — but that’s more of a nice-to-have. For me, this looks excellent 👏

itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/APIErrorResponse.java
```diff
@@ -0,0 +1,22 @@
...
+
+import java.time.Instant;
+import java.util.List;
+
+@JsonInclude(JsonInclude.Include.NON_EMPTY)
+@AllArgsConstructor
+@Getter
+@Builder
+public class APIErrorResponse {
+    private final Instant timestamp;
```
rioacosta, line 16: Maybe here you could use a direct initialization like 
@Builder.Default
private final Instant timestamp = Instant.now();
Instead of setting it in the GlobalExceptionHandler (Line 168).

If the APIErrorResponse constructor already does this automatically, you will be reducing the handler boilerplate and also be able to omit that line and simplify the handler code.

A DTO that autocompletes is easier for other developers to use because it prevents them from having to remember to include Instant.now() every time they build it.

This suggestion could be an improvement but can be dismissed too because is fuctional as it is.

rioacosta (approved):
Apart from the commented line on in the code and the issues addressed by @ToniR90 and @zohra-b, I think you've done an excellent job in this case.
 The error structure is clear and easy to document.
Good separation of responsibilities, use of DTO, and reuse of logic.
The error responses are more readable, translatable, and consistent.
Solid test coverage, both in the internal logic and in the behavior in the context of WebFlux.

I would remove all comments marked with // in the code, unless it is absolutely essential to leave them there. You can change it or wait for @ivilarop clarifications on this particular case.

itachallenge-challenge/src/main/java/com/itachallenge/challenge/exception/GlobalExceptionHandler.java
```diff
@@ -75,8 +83,27 @@ public ResponseEntity<MessageDto> handleNotFoundException(NotFoundException ex)
     }
 
     @ExceptionHandler(MethodArgumentNotValidException.class)
-    public ResponseEntity<MessageDto> handleMethodArgumentNotValidException(MethodArgumentNotValidException ex) {
-        return ResponseEntity.badRequest().body(new MessageDto(ex.getMessage()));
+    public ResponseEntity<APIErrorResponse> handleMethodArgumentNotValidException(
```
zohra-b, line 86: In a Spring WebFlux reactive application (with Mono/Flux), it is not recommended to use HttpServletRequest for several important reasons:
- Architectural incompatibility: HttpServletRequest is part of the Servlet API (blocking), while WebFlux is based on Netty/Reactor and operates in a non-blocking manner. Using HttpServletRequest goes against the reactive model.
- Thread blocking: Accessing HttpServletRequest can block the reactive thread, negating the performance benefits of the reactive approach.
- Not available by default: If you're using Netty as your server (the default in WebFlux), HttpServletRequest is simply not available.

Recommended Alternatives
Use ServerHttpRequest or ServerWebExchange

zohra-b (changes requested):
Changes to be done in GlobalExceptionHandler and GlobalExceptionHandlerTest to avoid using HttpServletRequest

aurelien-darbellay:
Thanks @zohra-b . Just changed it.

aurelien-darbellay:
@zohra-b : I thought it was a plisplas... but no. Turned out some tests started failing. And it was far from obvious the reason why. I finally got to the core of it: the challenge microservice isn't really reactive. While it uses reactive syntaxis in the code, it still runs the Tomcat blocking servlet. And hence, the exception types and request wrappers objects generated by Tomcat do not belong to the reactive realm. That means that, once I passed a reactive wrapper to the method that is supposed to catch the exception, it stopped being called, because this reactive wrapper is never generated by Tomcat in the first-place. And I don't think there is any fix to it. So, I'll leave the code as it is. thanks anyway, it was a very interesting investigation.

zohra-b (approved):

aurelien-darbellay:
close for the benefit of the general error handling strategy based on error-response-core.

PR #1005, opened 2025-10-15
PR1/2 - refactor(user) - refactor userAPIErrorResponse Dto (#787 & #800)

zohra-b:
### SUMMARY

In this PR, we import the new error-response-core module in itachallenge-user service, to implement consistent treatment of exception thoughout the whole system ([#800](https://tree.taiga.io/project/luisvi-ita-challenges/task/800)). 

It is a revamp of an older PR that intended to standardize Error handling without using a shared module (task 787). 

-------------------------------------
- [Context and Changes](#context-and-changes)
- [Modified Files](#modified-files)
- [NotaBene](#nota-bene)
- [Definition of Done](#definition-of-done)
- [Author Self-check](#self-check)

-----------------------------------------------------------------------------------------
<a id="context-and-changes"></a>
### Context and Changes
**Before**
-Error-response-core module not imported.
-No message.properties file in the service. 
-Error messages hard-coded in the service's Dtos. 

**After**
-Error-response-core module imported (in build.gradle (project(":error-response-core")) and ErrorHandlingConfig.class in the Spring Application).

-------------
<a id="modified-files"></a>
**Modified Files**
- [x] build.gradle
- [x] App.java
- [x] created messages.properties
- [x] modified fields in AdminCreateUserRequestDto and in UserSolutionRequestDto so that they use messages.properties
- [x] refactored UserControllerTest : added MessageSource Class to use messages.properties
- [x] refactored AdminCreateUserRequestDtoTest : in order to use messages.properties

----------------------------------------------
### 🧾 Versioning Note  
This pull request does **not include a version bump**, as it does not introduce any functional or behavioural changes.  
The modifications are limited to **internal refactoring and structural improvements** aimed at increasing code clarity and maintainability.  
Runtime behaviour, API contracts, and outputs remain unchanged.

-----------------------------------------------
<a id="nota-bene"></a>
**Nota Bene** : SonarCloud detects 2 issues : these are 2 comments block corresponding to a handler and its test that are out of scope and where the new dto should be implemented in PR #1008 

-----------------------------------------------------
<a id="definition-of-done"></a>
**DEFINITION OF DONE :**
- [x] modified build.gradle and App.java
- [x] create messages.properties file and implement it through the microservice
- [x] refactor / create related tests

--------------------------------------------
<a id="self-check"></a>
**Author Self-check**
- [x]  I tested my changes locally and verified they work as expected
- [x]  The PR has a clear and focused purpose (single feature/fix/refactor)
[...]

PR #790, opened 2025-02-04
Feature/778 users collection

davidcuadrado:

Issue #788, opened 2025-02-03
Add additional languages to database


Gatchan1:
So far there are 4 programming languages in the database: PHP, Java, JavaScript and Python.
Now we have to add **SQL** and **TypeScript**, in order to match the frontend design (view figma).
Add documents for them in the "languages" collection, in Challenge microservice.

- [x] edit database feeding json file with both new languages.
- [x] add equally both languages in remote database.

PR #1006, opened 2025-10-16
PR2.1/2 - refactor(user) - 4 user exception handlers (#789)

zohra-b:
**SUMMARY**
Refactorize core exception handlers to return standardized ApiErrorResponseDto (#1005 )

-------

**NOTA BENE** : 

SonarCloud detects 2 issues : these are 2 comments block corresponding to a handler and its test that are out of scope and where the new dto should be implemented in PR #1008 

-------

**MODIFIED FILES**
- UserGlobalExceptionHandler : refactored 4 handlers
- UserGlobalExceptionHandlerTest : refactored tests for the 4 handlers

**DETAILS**
This PR continues the work on standardizing error responses across the Challenge microservice. Building on #996, we now refactor the following handlers to produce structured, consistent error responses:

- [x] handleAny(Exception e)
- [x] handleIllegalArgument(IllegalArgumentException e)
- [x] handleValidationExceptions(ConstraintViolationException ex)
- [x] handleTypeMismatchException(MethodArgumentTypeMismatchException ex)

Each handler now returns an ApiErrorResponseDto that includes:
- Timestamp – when the error occurred
- HTTP Status – corresponding to the exception type 
- Message – human-friendly description 
- Field-level errors – via FieldErrorDto, if applicable

This ensures all validation and exception errors are clear, consistent, and user-friendly, while technical details are safely logged rather than exposed.

**OUT OF SCOPE**
Handlers refactored in [Taiga #788](https://tree.taiga.io/project/luisvi-ita-challenges/task/788) and [Taiga #790](https://tree.taiga.io/project/luisvi-ita-challenges/t/790)

**IMPACT**
API error responses are now standardized across the microservice
Existing unit tests updated and verified green
Messages are more user-friendly and secure (exception details are logged, not returned)

**DEFINITION OF DONE**

- [x] 4 handlers refactored to return ApiErrorResponseDto
- [x] Unit tests updated and passing

**Author Self-check**

- [x]  I tested my changes locally and verified they work as expected
- [x]  The PR has a clear and focused purpose (single feature/fix/refactor)
- [x]  Title, description, and changelog are correctly filled
- [x]  All automated checks (SonarCloud, CI) pass
- [x] I responded to all previous review comments
- [x] No unrelated commits or merge leftovers remain

### DIFF
--- a/itachallenge-user/src/main/java/com/itachallenge/user/dto/APIErrorResponse.java
+++ b/itachallenge-user/src/main/java/com/itachallenge/user/dto/APIErrorResponse.java
@@ -3,12 +3,14 @@ import com.fasterxml.jackson.annotation.JsonInclude;
 import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Getter;
+import lombok.NoArgsConstructor;
 
 import java.time.Instant;
 import java.util.List;
 
 @JsonInclude(JsonInclude.Include.NON_EMPTY)
 @AllArgsConstructor
+@NoArgsConstructor(force = true)
 @Getter
 @Builder
 public class APIErrorResponse {
--- a/itachallenge-user/src/main/java/com/itachallenge/user/dto/AdminCreateUserRequestDto.java
+++ b/itachallenge-user/src/main/java/com/itachallenge/user/dto/AdminCreateUserRequestDto.java
@@ -7,7 +7,7 @@ import lombok.*;
 @Setter
 public class AdminCreateUserRequestDto {
 
-    @NotBlank(message = "Username must not be blank")
+    @NotBlank(message = "{adminCreateUser.username.notBlank}")
     private String username;
 
 }
--- a/itachallenge-user/src/main/java/com/itachallenge/user/dto/FieldErrorDto.java
+++ b/itachallenge-user/src/main/java/com/itachallenge/user/dto/FieldErrorDto.java
@@ -1,21 +1,16 @@
 package com.itachallenge.user.dto;
+
 import com.fasterxml.jackson.annotation.JsonInclude;
 import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Getter;
 
-import java.time.Instant;
-import java.util.List;
-
 @JsonInclude(JsonInclude.Include.NON_EMPTY)
+@Builder
 @AllArgsConstructor
 @Getter
-@Builder
-public class APIErrorResponse {
-    private final Instant timestamp;
-    private final int status;
-    private final String error;
+public class FieldErrorDto {
+    private final String objectName;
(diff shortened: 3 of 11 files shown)


Meriekje (changes requested):
All 4 targeted exception handlers correctly return the new standardized APIErrorResponse DTO structure.
On the other hand, SonarCloud reports 2 issues. Please, resolve them or document the reason that part of the code is commented.

zohra-b:
Thanks for your review @Meriekje .
The 2 issues are related to 2 blocks commented (one in the GlobalExceptionHandler, the other in GlobalExceptionHandler test). I commented t'hem because they are out off scope. (Your user story)

Meriekje:
> Thanks for your review @Meriekje . The 2 issues are related to 2 blocks commented (one in the GlobalExceptionHandler, the other in GlobalExceptionHandler test). I commented t'hem because they are out off scope. (Your user story)

Yes, that's why I said to document it so it is clear for anyone who can check this PR appart from me.

itachallenge-user/src/main/java/com/itachallenge/user/dto/APIErrorResponse.java
```diff
@@ -1,27 +1,23 @@
...
-
-    @JsonProperty("message")
-    String message;
-
-    @JsonProperty("timestamp")
-    Instant timestamp;
-
-}
-
+    private final Instant timestamp;
```
rioacosta, line 17: I gave this comment to @aurelien-darbellay too.

Maybe here you could use a direct initialization like:

@Builder.Default
private final Instant timestamp = Instant.now();

Instead of setting it in the GlobalExceptionHandler (Line 168).

If the APIErrorResponse constructor already does this automatically, you will be reducing the handler boilerplate and also be able to omit that line and simplify the handler code.

A DTO that autocompletes is easier for other developers to use because it prevents them from having to remember to include Instant.now() every time they build it.

This suggestion could be an improvement but can be dismissed too because is functional as it is.

rioacosta (changes requested):
I'm following the branching with this task of harmonizing the error response across the whole app and as far as the definition of this task scope, I think the coding is coherent whit the rest of the work done but I'll make eco for my previous requests for the improvement of the PR documentation, in this case at least the summary needs to be explicative in a friendlier way.

itachallenge-user/src/main/java/com/itachallenge/user/dto/APIErrorResponse.java
zohra-b: Thank @rioacosta . I think this comment is related to PR #1005 as this one just applies the dto into the handlers. I'll discuss it with Aurelien so that we follow the same modifications if needed.

PR #1010, opened 2025-10-22
core(error-response-core): Introduce Shared Error Handling Module — `error-response-core` (v1.0.0) (#796)

aurelien-darbellay:
# 🧩 PR: Refactor and Extend Shared Error Handling Module — `error-response-core` (v3.2.0)

## 🎯 Goal

This PR refactors and extends the shared Gradle module **`error-response-core`** to provide a robust, localized, and ISO-8601–compliant foundation for unified error handling across all backend microservices in the **ITA Challenge platform**.

It enhances the original implementation from v1.0.0 by introducing a **new `BaseApiException`**, improved **`ErrorResponseBuilder`** logic, refined **exception mapping**, and **ISO-8601 timestamp serialization** using `Instant`.  
It keeps full compatibility with **Spring Boot 3.0.6** and all existing service integrations.

---

### ✅ What’s Introduced / Updated

| Component | Description |
|------------|-------------|
| **APIErrorResponse** | Unified DTO for structured error payloads (includes ISO-8601 `timestamp`, `status`, `message`, `path`, and optional field errors). |
| **FieldErrorDto** | Represents field-level validation errors (`field`, `objectName`, `message`). |
| **BaseApiException** | Abstract base class for service-defined exceptions — encapsulates ApiErrorInfo as a field. |
| **ApiErrorInfo** | Record class which encapsulates all the information needed to build error objects with i18n messages: httpStatus, messagekey (reference to messages.properties file), args to populate the message if needed . |
| **ErrorResponseBuilder** | Core utility for building standardized `APIErrorResponse` objects. Handles Spring framework exceptions, constraint violations, and custom `BaseApiException` instances. Includes localized message resolution and ISO-8601 timestamp formatting. |
| **GlobalExceptionHandler** | `@RestControllerAdvice` that intercepts framework exceptions and BaseApiException, and delegates to `ErrorResponseBuilder`. Can be overridden with `@Order(HIGHEST_PRECEDENCE)` in local handlers. |
| **core-messages.properties** | Default message bundle providing localized error messages (e.g. `validation.type_mismatch`, `error.internal`). |

All microservices can now import `error-response-core` to get a unified, localized, and testable error-handling system.  
Custom exceptions extending `BaseApiException` automatically benefit from consistent JSON output and message resolution.

N.B: ApiErrorInfo has a factory method ".of" that allows hiding the constructor in the BaseApiException subclasses, to improve maintainability. 
---

### 🧾 Example Error Response

A standardized JSON structure for frontend clarity and debugging consistency:

```json
{
  "timestamp": "2025-10-22T10:15:30.012Z",
  "status": 400,
  "error": "Bad Request",
  "message": "Parameter 'age' has invalid value 'abc'. Expected type: Integer.",
  "errors": [
    {
      "objectName": "UserController",
      "field": "age",
      "message": "Parameter 'age' has invalid value 'abc'. Expected type: Integer."
    }
  ],
  "path": "/api/users"
}
```

---

## 🏗️ Module Structure

```
error-response-core/
├── build.gradle
├── src/
│   ├── main/java/com/itachallenge/errorcore/
│   │   ├── builder/ErrorResponseBuilder.java
│   │   ├── dto/
│   │   │   ├── APIErrorResponse.java
│   │   │   └── FieldErrorDto.java
│   │   ├── exception/BaseApiException.java
│   │   └── exceptionhandler/GlobalExceptionHandler.java
│   └── resources/core-messages.properties
│
└── test/java/com/itachallenge/errorcore/
    ├── builder/ErrorResponseBuilderTest.java
    ├── dto/APIErrorResponseTest.java
    ├── dto/FieldErrorDtoTest.java
    └── exceptionhandler/GlobalExceptionHandlerTest.java
```

---

## 🧪 Testing Strategy

Comprehensive test suites verify correctness and consistency across all layers:

| Test Suite | Purpose | Notes |
|-------------|----------|-------|
| **APIErrorResponseTest** | Validates DTO construction, builder pattern, and JSON serialization. | Ensures stable, ISO-8601 timestamp serialization. |
| **FieldErrorDtoTest** | Tests mapping of validation errors. | Includes equality and message formatting checks. |
| **ErrorResponseBuilderTest** | Verifies exception-to-message mapping, localization, and error payload integrity. | Includes edge cases for missing messages and unmapped exceptions. |
| **GlobalExceptionHandlerTest** | Confirms delegation to `ErrorResponseBuilder` and correct `ResponseEntity` construction. | Unit test only (no Spring context). |
| **ErrorCoreIntegration** | Confirmes the whole module is well integrated | Minimal spring context. |

- Frameworks: **JUnit 5 + AssertJ + Mockito**  
- Coverage: tracked via **JaCoCo (≥70%)**  
- Report: `build/reports/jacoco/test/html/index.html`

---

## 🧩 JaCoCo & Sonar Integration

Integrated coverage and quality verification with **JaCoCo** and **SonarQube**:

```groovy
test {
    useJUnitPlatform()
    finalizedBy jacocoTestReport
}

jacocoTestReport {
    dependsOn test
    reports {
        xml.required = true
        csv.required = false
        html.outputLocation = layout.buildDirectory.dir('jacocoHtml')
    }
}

jacocoTestCoverageVerification {
    violationRules {
        rule {
            limit {
                counter = 'LINE'
                value = 'COVEREDRATIO'
                minimum = 0.7
            }
        }
    }
}

sonar {
    properties {
        property "sonar.coverage.jacoco.xmlReportPaths",
            layout.buildDirectory.file("reports/jacoco/test/jacocoTestReport.xml").get().asFile.absolutePath
    }
}
```

**Highlights:**
- Enforces **≥70% line coverage**
- Generates **HTML** + **XML** coverage reports
- Seamless CI integration via GitHub Actions
- Ensures SonarQube quality gates remain green

---

## ⚙️ Gradle & Development Notes

- `bootJar` disabled → not a standalone app  
- `jar` enabled → reusable shared library  
- Built with **Java 21**, **Spring Boot 3.0.6**, and **Jakarta Validation 3.0.2**  
- Uses **`java-library`** plugin for proper dependency exposure  
- **Lombok** included as `compileOnly`  
- **`jakarta.servlet-api`** set as `compileOnly` (runtime provided by container)  
- `Instant` timestamp in `APIErrorResponse` outputs ISO-8601 by default — `jackson-datatype-jsr310` is auto-registered by Spring Boot 3.x  
- `messageArgs` in `BaseApiException` marked `transient` to comply with SonarQube serialization rules  
- Fully stateless, dependency-free, and compatible with all ITA Challenge microservices

---

## 🚀 Integration Instructions

Add dependency:

```groovy
dependencies {
    implementation project(":error-response-core")
}
```

Register in `settings.gradle`:

```groovy
include(":error-response-core", ":user-service", ":challenge-service")
```

Import in the service’s main Spring configuration:

```java
@Import(ErrorHandlingConfig.class)
```
Custom service exceptions that extends BaseApiException will be automatically handled by the GlobalExceptionHandler. To create subClasses of BaseApiException: 
   1. Write the subclass, using the ApiErrorInfo factory ".of" to hide the constructor and shield your expression against future enhancements. 
   2. Configure the messageKey and it's corresponding message in the local messages.properties. 

Shall you need to define a local handler (but think twice before doing it - in principle it shouldn't be necessary):

```java
@RestControllerAdvice
@RequiredArgsConstructor
public class UserExceptionHandler {
    private final ErrorResponseBuilder responseBuilder;

    @ExceptionHandler(UserNotFoundException.class)
    public ResponseEntity<APIErrorResponse> handle(UserNotFoundException ex, HttpServletRequest req) {
        return ResponseEntity.status(ex.getStatus())
                             .body(responseBuilder.buildError(ex, req));
    }
}
```

Override default handler behavior with:

```java
@Order(Ordered.HIGHEST_PRECEDENCE)
```

To customize messages, define:

```
src/main/resources/messages.properties
```

Spring merges this automatically with `core-messages.properties`.

---

## 🔍 Expected Impact

✅ Centralized and extensible error-handling framework  
✅ Unified JSON response format across all services  
✅ Localized message resolution and i18n support  
✅ Reduced duplication and simpler maintenance  
✅ ISO-8601 timestamps and SonarQube-compliant implementation  
✅ Easier debugging and improved frontend consistency  

---

## 🧩 Version & Artifact

| Key | Value |
|-----|--------|
| **Version** | `1.0.0` |
| **Artifact** | `error-response-core` |
| **Group** | `com.itachallenge.errorcore` |

Stable release introducing `BaseApiException`, refined builder logic, transient serialization fix, and ISO-8601 timestamp support.  
Foundation for upcoming refactors in `user-service` and `challenge-service`.

---

## 📈 Next Steps

1. Rebase open PRs (#997, #998, #1005) to depend on this module.  
3. Remove redundant DTOs and handlers from each microservice.  
4. Verify message localization and consistency per environment.  
5. Confirm CI and SonarQube metrics on the new module.  
6. Migrate existing custom exceptions to `BaseApiException`.  
7. Document usage patterns in the internal backend wiki.  
8. `Error-response-core` could be significantly improved by implementing the technical debt explained in issue #[813](https://tree.taiga.io/project/luisvi-ita-challenges/issue/813) and issue #[814](https://tree.taiga.io/project/luisvi-ita-challenges/issue/814).

---

## ✅ Definition of Done

- [x] Independent Gradle module builds successfully  
- [x] DTOs, builder, and handlers implemented and tested  
- [x] Unit + integration coverage validated via JaCoCo  
- [x] SonarQube integration confirmed  
- [x] ISO-8601 timestamps validated  
- [x] Ready for consumption by all backend microservices  

---

## 👥 Authors & Reviewers

**Authors:**  
- Aurélien Darbellay — Developer  

**Suggested Reviewers:**  
`@backend-team`  
`@qa-team`  
`@devops-team` (for CI & Sonar verification)

---

## ✍️ PR Author Self-Check

- [x] Tested changes locally with sample services  
- [x] Clear, single-purpose PR  
- [x] Title, description, and changelog updated  
- [x] No leftover merge conflicts or unrelated changes  
- [x] All CI and Sonar checks pass  
- [x] Responded to all prior review comments  
- [x] Ran SonarLint locally and confirmed no warnings  

---

### 🧩 TL;DR

> Refactors and stabilizes the shared **error-handling module** with `BaseApiException`, ISO-8601 timestamps, improved localization, and full SonarQube compliance — establishing a consistent, maintainable, and extensible foundation for backend error management across all ITA Challenge microservices.


vicenlu:
@codex review

zohra-b (changes requested):
### File : **error-response-core/build.gradle** : 

**Change**
```
implementation 'org.springframework.boot:spring-boot-starter-web:3.0.6'
implementation 'org.springframework.boot:spring-boot-starter-validation:3.0.6'
// These dependencies are not transitive, meaning that dependent projects must declare them manually
```

**to**

```
api 'org.springframework.boot:spring-boot-starter-web'
api 'org.springframework.boot:spring-boot-starter-validation'
// Dependencies are declared using api, which makes them public and transitive. Dependent modules automatically inherit the exposed dependencies
```


**Remove these dependencies**

`implementation 'jakarta.validation:jakarta.validation-api:3.1.0'   // already included in spring-boot-starter-validation
testImplementation 'org.mockito:mockito-core:4.11.0'                // laready included in spring-boot-starter-test
testImplementation 'org.mockito:mockito-junit-jupiter:4.11.0'    // altready included in spring-boot-starter-test
`   

--------------------------------------------------------------

### File **error-response-core/src/main/java/com/itachallenge/errorcore/builder/ErrorResponseBuilder.java** : 
File **error-response-core/src/main/java/com/itachallenge/errorcore/exceptionhandler/BaseExceptionHandler.java**

in a WebFlux (reactive) context, you shouldn’t use HttpServletRequest which belongs to the Servlet API (javax.servlet.http), which is part of Spring MVC — a blocking, thread-per-request model.
Instead you should use Instead of HttpServletRequest, WebFlux provides reactive equivalents: **ServerHttpRequest** or **ServerWebExchange**


-------------------------------------------------------------------
### File : **itachallenge-user/src/test/java/com/itachallenge/user/dto/APIErrorResponseTest.java**

test : **shouldSerializeErrorResponseWithMultipleErrorsAndMetadata()**  : you should separate the assertions in 2 blocks to make it more readable : 

```
assertThat(json)    // fields related to APIErrorResponse

                .startsWith("{").endsWith("}")
                .contains("\"status\":400")
                .contains("\"error\":\"Bad Request\"")
                .contains("\"message\":\"Validation failed\"")
                .contains("\"path\":\"/api/v1/challenges\"")
                .contains("\"timestamp\":\"2025-10-06T09:00:00Z\"");

assertThat(json)     // fields related to FieldErrorDto
                .contains("\"errors\"")
                .contains("\"objectName\":\"challengeDto\"")
                .contains("\"field\":\"title\"")
                .contains("\"message\":\"The title cannot be empty\"")
                .contains("\"field\":\"difficulty\"")
                .contains("\"message\":\"The difficulty is invalid\"");
```

*** you should add a test to verify that it serializes APIErrorResponse with simple errors (without FieldErrorDto)

-----------------------------------------------------------------------

### File : **error-response-core/src/main/java/com/itachallenge/errorcore/exceptionhandler/BaseExceptionHandler.java**

Avoid using e.getMessage() as it may expose internal messages

------------------------------------------------------------------------
### File : **error-response-core/src/test/java/com/itachallenge/errorcore/testapp/TestController.java**

Please add logging to help debugging

error-response-core/src/test/java/com/itachallenge/errorcore/testapp/TestController.java
```diff
@@ -11,32 +13,42 @@
...
 
+    Logger log = LoggerFactory.getLogger(TestController.class);
+
     @GetMapping("/mismatch")
     public String mismatch(@RequestParam("age") @Min(10) int age) {
+        log.debug("Made it to /mismatch path. " +
+                "To trigger mismatch type, pass an unconvertible string as param; " +
+                "to trigger constrain violation, pass a number smalle than 10"
+        );
         return "OK";
```
zohra-b: L22 : ""to trigger constrain violation, pass a number smalle than 10"
To trigger with majuscule /constraint / smalleR

zohra-b (changes requested):

error-response-core/src/test/java/com/itachallenge/errorcore/testapp/TestController.java
aurelien-darbellay: thanks @zohra-b . I think the other observations have been addressed!

zohra-b (approved):

rioacosta (approved):
Great work consolidating error handling in a shared module and applying the reviewers' comments. 
Solid and extensible design: good separation of responsibilities between DTOs, builder, and handler. The abstraction in BaseExceptionHandler facilitates reuse.
Testing has good coverage, although it could be improved in the future.
I liked APIErrorResponseTest; it will be very useful.
The instructions for integration with microservices are clear and well documented for importing, extending, and overwriting messages. It would be ideal to add a section in the main wiki on how to extend the handler and document common errors already covered.

In summary, this PR establishes a robust foundation for consistent, maintainable, and easily integrable error handling across all services. Great work!

error-response-core/src/main/java/com/itachallenge/errorcore/exceptionhandler/BaseExceptionHandler.java
```diff
@@ -0,0 +1,119 @@
...
+import org.springframework.web.bind.annotation.ExceptionHandler;
+import org.springframework.web.method.annotation.MethodArgumentTypeMismatchException;
+import org.springframework.web.server.ResponseStatusException;
+import org.springframework.web.server.ServerWebInputException;
+
+@Slf4j
+@RequiredArgsConstructor
+public abstract class BaseExceptionHandler {
+
+    protected final ErrorResponseBuilder responseBuilder;
```
ivilarop: abstract class + protected … responseBuilder

This introduces inheritance coupling and breaks encapsulation: by exposing internals with `protected`, the core becomes tied to how microservices inherit from it. Every future need will require "opening up" more things, and the core will become fragile.

Impact:

- Core refactors can break subclasses in several microservices. 
- Testing and evolving become more complicated (it depends on hierarchies). 
- It goes against best practices: composition > inheritance in shared modules.

Recommendation: Let's not force inheritance or expose internals with `protected`. Error handling should be done through composition, and its details should be kept encapsulated.

error-response-core/build.gradle
```diff
@@ -0,0 +1,77 @@
...
+    mavenCentral()
+}
+
+dependencies {
+    // --- Lombok ---
+    compileOnly 'org.projectlombok:lombok:1.18.34'
+    annotationProcessor 'org.projectlombok:lombok:1.18.34'
+
+    // --- Spring Boot Core + Validation ---
+    implementation 'org.springframework.boot:spring-boot-starter-web'
```
ivilarop: A core should be a lightweight library (minimal framework), not dragging along the entire boot stack. Starters belong in microservices (or in a separate boot starter), not in the core.

error-response-core/build.gradle
```diff
@@ -0,0 +1,77 @@
+plugins {
+    id 'org.springframework.boot' version '3.0.6'
```
ivilarop: A shared core must be pure java-library and inherit versions from the root.

ivilarop (changes requested):

error-response-core/src/main/java/com/itachallenge/errorcore/exceptionhandler/BaseExceptionHandler.java
aurelien-darbellay: Hey! I'm not sure whether the problem was only with the protected or with the protected and with the abstract class. I've  replaced protected with private. I've added a getter to be able to retrieve the responsebuilder in the subclasses.

error-response-core/build.gradle
aurelien-darbellay: I've replaced the starter by a list of the exact dependencies the module needs. in order to do so, I had to remove the integration test that runs with SpringBootTest.

error-response-core/build.gradle
aurelien-darbellay: I've remove this plugin and replaced it with : dependencyManagement {
    imports {
        mavenBom "org.springframework.boot:spring-boot-dependencies:3.0.6"
    }
} If I've understood correctly, this is less heavy but it still doesn't mean that versions are inherited from the root. But in order to get version inheritance, this block code would have to be added to the root build.gradle, which I'd rather avoid touching. This might be considered in later improvements, maybe at the same time as the starter and the versions are removed from the other shared core (github-core and jwt-core).

aurelien-darbellay: Btw: thanks for the comments! Now I understand better what libraries and shared modules should be.

error-response-core/src/main/java/com/itachallenge/errorcore/exceptionhandler/BaseExceptionHandler.java
aurelien-darbellay: I've gotten rid of the abstract class too!

error-response-core/src/main/java/com/itachallenge/errorcore/exceptionhandler/BaseExceptionHandler.java
```diff
@@ -0,0 +1,124 @@
...
+import org.springframework.web.bind.annotation.RestControllerAdvice;
+import org.springframework.web.method.annotation.MethodArgumentTypeMismatchException;
+import org.springframework.web.server.ResponseStatusException;
+import org.springframework.web.server.ServerWebInputException;
+
+@Slf4j
+@RestControllerAdvice
+@RequiredArgsConstructor
+@Order(Ordered.LOWEST_PRECEDENCE)
+public class BaseExceptionHandler {
```
ivilarop: 1.-Mark the class as final

Prevent anyone from accidentally extending it again in the future.

A clear indication that it's a component for direct use, not an extension.

2.-BaseExceptionHandler is no longer "base". We can rename it to something like GlobalErrorHandler so it's not confused with and extended from it.

error-response-core/README.md
```diff
@@ -0,0 +1,166 @@
...
+## 🧩 Module Contents
+
+### **Main Components**
+
+| Package                                       | Class | Description |
+|-----------------------------------------------|--------|-------------|
+| `com.itachallenge.errorcore.builder`          | **`ErrorResponseBuilder`** | Core utility for building structured `APIErrorResponse` objects. Handles argument validation, type mismatches, constraint violations, and general exceptions. |
+| `com.itachallenge.errorcore.dto`              | **`APIErrorResponse`** | DTO representing the unified error payload returned to clients (includes timestamp, status, message, error details, path). |
+|                                               | **`FieldErrorDto`** | DTO for detailed field-level validation errors (`field`, `objectName`, `message`). |
+| `com.itachallenge.errorcore.exceptionhandler` | **`BaseExceptionHandler`** | Abstract `@RestControllerAdvice` that defines centralized exception handling methods. Can be extended by any microservice. |
```
ivilarop: That contradicts the change we have made (it is no longer abstract nor should it be extended).

error-response-core/README.md
```diff
@@ -0,0 +1,166 @@
...
+    implementation project(":error-response-core")
+}
+```
+Ensure the module is declared in your root settings.gradle:
+
+```groovy
+include(":error-response-core", ":user-service", ":challenge-service")
+```
+
+### 2. Extend the Base Exception Handler
```
ivilarop: That contradicts the change we have made (it is no longer abstract nor should it be extended).

ivilarop (changes requested):

error-response-core/src/main/java/com/itachallenge/errorcore/exceptionhandler/BaseExceptionHandler.java
aurelien-darbellay: cool! I think it is done.

error-response-core/README.md
aurelien-darbellay: sorry. I forgot to revise the Readme. It's done now

error-response-core/README.md
aurelien-darbellay: same

error-response-core/src/main/java/com/itachallenge/errorcore/builder/ErrorResponseBuilder.java
```diff
@@ -0,0 +1,204 @@
...
+        return APIErrorResponse.builder()
+                .status(status.value())
+                .error(status.getReasonPhrase())
+                .message(resolveMessage(message))
+                .path(request != null ? request.getRequestURI() : null)
+                .build();
+    }
+
+    public APIErrorResponse buildNotFoundError(RuntimeException ex, HttpServletRequest request){
+        return buildError(HttpStatus.NOT_FOUND,ex.getMessage(),request);
```
ivilarop: Do not expose e.getMessage() to the client.

Do not use ex.getMessage() here; the builder/service should resolve an i18n message

error-response-core/src/main/java/com/itachallenge/errorcore/exceptionhandler/GlobalExceptionHandler.java
```diff
@@ -0,0 +1,124 @@
...
+
+    private final ErrorResponseBuilder responseBuilder;
+
+    @ExceptionHandler(Exception.class)
+    public ResponseEntity<APIErrorResponse> handleAny(Exception e, HttpServletRequest request) {
+        log.error("Unexpected error happened: {}", e.getMessage());
+        return ResponseEntity
+                .status(HttpStatus.INTERNAL_SERVER_ERROR)
+                .body(responseBuilder
+                        .buildError(HttpStatus.INTERNAL_SERVER_ERROR, e.getMessage(), request)
```
ivilarop: Remove that e.getMessage() from the payload (the builder/service should decide the message via i18n)

error-response-core/src/main/java/com/itachallenge/errorcore/exceptionhandler/GlobalExceptionHandler.java
```diff
@@ -0,0 +1,124 @@
...
+                );
+    }
+
+    @ExceptionHandler(IllegalArgumentException.class)
+    public ResponseEntity<APIErrorResponse> handleIllegalArgument(IllegalArgumentException e, HttpServletRequest request) {
+        log.error("Illegal argument happened: {}", e.getMessage());
+        return ResponseEntity
+                .status(HttpStatus.BAD_REQUEST)
+                .body(responseBuilder
+                        .buildError(HttpStatus.BAD_REQUEST, e.getMessage(), request)
```
ivilarop: Remove that e.getMessage() from the payload (the builder/service should decide the message via i18n)

error-response-core/src/main/java/com/itachallenge/errorcore/builder/ErrorResponseBuilder.java
```diff
@@ -0,0 +1,204 @@
...
+                request
+        );
+    }
+
+    public APIErrorResponse buildStatusErrorResponse(ResponseStatusException ex, HttpServletRequest request) {
+        HttpStatusCode statusCode = ex.getStatusCode();
+        Object[] detailMessageArguments = ex.getDetailMessageArguments();
+        String errorMessage;
+
+        if (detailMessageArguments == null || detailMessageArguments.length == 0) {
```
ivilarop: Do not use ex.getMessage() or compose the text from detailMessageArguments for the payload. Resolve it using i18n based on the status.

error-response-core/src/main/java/com/itachallenge/errorcore/builder/ErrorResponseBuilder.java
```diff
@@ -0,0 +1,204 @@
...
+
+@RequiredArgsConstructor
+public class ErrorResponseBuilder {
+
+    private final MessageSource messageSource;
+
+    /** Build general error response.*/
+    public APIErrorResponse buildError(
+            HttpStatus status,
+            String message,
```
ivilarop: This isn't "a single line" with `getMessage`, but it's the source of the problem: this method allows any String to be passed from outside (and `e.getMessage()` is being passed to it).

Critical action: stop using this signature in the handler (don't pass `String message`). The builder/service must receive the exception and decide on the message.

ivilarop (changes requested):

error-response-core/src/main/java/com/itachallenge/errorcore/exceptionhandler/BaseExceptionHandler.java
ivilarop: Have you changed your name?

error-response-core/build.gradle
```diff
@@ -0,0 +1,94 @@
...
+    id 'jacoco'
+    id 'java-library'
+}
+
+group = 'com.itachallenge.errorcore'
+version = '1.0.0'
+
+dependencyManagement {
+    imports {
+        mavenBom "org.springframework.boot:spring-boot-dependencies:3.0.6"
```
ivilarop, line 12: Spring Boot BOM pinned within the module

Why it's wrong: You're pinning versions (3.0.6) in the submodule. In a monorepo, versions are managed from the root. This causes drift with respect to the other micromodules.

What to do: Remove that local dependencyManagement and inherit from the root project's BOM. (If the root doesn't already have it, move it there.)

error-response-core/build.gradle
```diff
@@ -0,0 +1,94 @@
...
+    annotationProcessor 'org.projectlombok:lombok:1.18.34'
+
+    // --- Spring (minimal) ---
+            implementation 'org.springframework:spring-context'
+    implementation 'org.springframework:spring-web'
+    implementation 'org.springframework:spring-webmvc'
+
+    // --- Jakarta APIs ---
+    implementation 'jakarta.validation:jakarta.validation-api:3.0.2'
+    implementation 'jakarta.servlet:jakarta.servlet-api:6.0.0'
```
ivilarop: Change jakarta.servlet-api from implementation to compileOnly (the runtime is provided by the micro/server).

ivilarop (changes requested):

error-response-core/build.gradle
```diff
@@ -0,0 +1,94 @@
...
+
+dependencies {
+    // --- Lombok ---
+    compileOnly 'org.projectlombok:lombok:1.18.34'
+    annotationProcessor 'org.projectlombok:lombok:1.18.34'
+
+    // --- Spring (minimal) ---
+            implementation 'org.springframework:spring-context'
+    implementation 'org.springframework:spring-web'
+    implementation 'org.springframework:spring-webmvc'
```
ivilarop: Remove spring-webmvc: in a core you don't need the MVC stack; spring-web and spring-context are enough.

ivilarop (changes requested):

error-response-core/src/main/java/com/itachallenge/errorcore/config/ErrorHandlingConfig.java
```diff
@@ -0,0 +1,38 @@
...
+    @Bean
+    public ErrorResponseBuilder errorResponseBuilder(MessageSource messageSource) {
+        return new ErrorResponseBuilder(messageSource);
+    }
+
+    // This is abstract, so we don't instantiate it.
+    // But exposing it as a bean definition allows Spring to process its metadata
+    // if you ever have non-abstract variants in your shared module.
+    @Bean
+    public BaseExceptionHandler baseExceptionHandler(ErrorResponseBuilder builder) {
```
ivilarop: The core should not register web beans (@RestControllerAdvice) or create anonymous subclasses of the handler. This can cause conflicts/double registration in microservices and reintroduces coupling.

Incorrect comment: it's no longer abstract, so the comment is misleading.

Agreed design: composition. The handler should be registered in a boot-starter (auto-config) or in each microservice, not in the core.

Risk: by exposing this bean from the core, we force a global decision and limit the configuration of each microservice.

Analyze what the best solution would be?

error-response-core/src/main/java/com/itachallenge/errorcore/exceptionhandler/BaseExceptionHandler.java
aurelien-darbellay: yes... the push had failed. but I think it is okay now.

ivilarop (changes requested):

error-response-core/src/main/java/com/itachallenge/errorcore/exceptionhandler/GlobalExceptionHandler.java
aurelien-darbellay: done

error-response-core/src/main/java/com/itachallenge/errorcore/builder/ErrorResponseBuilder.java
aurelien-darbellay: done.

error-response-core/src/main/java/com/itachallenge/errorcore/dto/APIErrorResponse.java
```diff
@@ -0,0 +1,23 @@
...
+import java.time.Instant;
+import java.util.List;
+
+@JsonInclude(JsonInclude.Include.NON_EMPTY)
+@AllArgsConstructor
+@Getter
+@Builder
+public class APIErrorResponse {
+    @Builder.Default
+    private final Instant timestamp = Instant.now();
```
ivilarop, line 17: Ensure the timestamp is in ISO-8601 format. With Instant, you need the runtime to register jackson-datatype-jsr310; otherwise, it may output in epoch-millis. (A clear alternative is to use OffsetDateTime in UTC.)

ivilarop (changes requested):

error-response-core/src/main/java/com/itachallenge/errorcore/config/ErrorHandlingConfig.java
```diff
@@ -0,0 +1,35 @@
...
+        return src;
+    }
+
+    @Bean
+    public ErrorResponseBuilder errorResponseBuilder(MessageSource messageSource) {
+        return new ErrorResponseBuilder(messageSource);
+    }
+
+    @Bean
+    public GlobalExceptionHandler globalExceptionHandler(ErrorResponseBuilder builder) {
```
ivilarop, line 33: The core should not register web beans (@RestControllerAdvice) or create anonymous subclasses of the handler. This can cause conflicts/double registration in microservices and reintroduces coupling.

Incorrect comment: it's no longer abstract, so the comment is misleading.

Agreed design: composition. The handler should be registered in a boot-starter (auto-config) or in each microservice, not in the core.

Risk: by exposing this bean from the core, we force a global decision and limit the configuration of each microservice.

Analyze what the best solution would be?

error-response-core/src/test/java/com/itachallenge/errorcore/builder/ErrorResponseBuilderTest.java
```diff
@@ -0,0 +1,181 @@
...
+        request = mock(HttpServletRequest.class);
+        when(request.getRequestURI()).thenReturn("/api/test");
+    }
+
+    // ------------------------------------------------------------
+    // buildError
+    // ------------------------------------------------------------
+    @Test
+    void buildError_shouldBuildBasicResponseWithRealMessage() {
+        APIErrorResponse response = builder.buildError(HttpStatus.BAD_REQUEST, "validation.bad_request", request);
```
ivilarop: Problem: The test calls `buildError(HttpStatus, "validation.bad_request", request). This forces the builder to maintain an API that accepts free strings.

What needs to change: The test should not pass messages. It should invoke a method that receives the exception (or use case) and have the builder determine the text using i18n.

error-response-core/src/test/java/com/itachallenge/errorcore/builder/ErrorResponseBuilderTest.java
```diff
@@ -0,0 +1,181 @@
...
+        assertThat(response.getStatus()).isEqualTo(400);
+    }
+
+    @Test
+    void buildNotFoundError_shouldReturnNotFoundResponse() {
+        // Given
+        RuntimeException ex = new RuntimeException("Resource not found");
+
+        // When
+        APIErrorResponse response = builder.buildNotFoundError(request,ex.getMessage());
```
ivilarop: Problem: It passes `ex.getMessage()` to the builder and expects the same text in the response. This reopens the door to arbitrary messages.

What needs to change: The test should not pass a random message; it should invoke the route that constructs the 404 error and expect the `i18n` error.not_found message (with safe arguments if available).

error-response-core/src/test/java/com/itachallenge/errorcore/builder/ErrorResponseBuilderTest.java
```diff
@@ -0,0 +1,181 @@
...
+        assertThat(response.getErrors().getFirst().getField()).isEqualTo("email");
+        assertThat(response.getErrors().getFirst().getMessage()).isEqualTo("email cannot be null");
+        assertThat(response.getMessage()).contains("object 'testDto'");
+    }
+
+    // ------------------------------------------------------------
+    // buildStatusErrorResponse (ResponseStatusException)
+    // ------------------------------------------------------------
+    @Test
+    void buildStatusErrorResponse_shouldUseValidationFailedWhenNoArgs() {
```
ivilarop: Problem: They expect the payload to contain `ex.getMessage()(404 NOT_FOUND "Not found")` or the join of `detailMessageArguments("a", "b", "c"). This forces the exposure of internal/non-i18n text.

What needs to change: The test should expect an i18n message based on the status (e.g., for 404, the `error.not_found` message; for 400, the `validation.bad_request` message), with controlled arguments if applicable.

error-response-core/src/test/java/com/itachallenge/errorcore/builder/ErrorResponseBuilderTest.java
```diff
@@ -0,0 +1,181 @@
...
+        ResponseStatusException ex = new ResponseStatusException(HttpStatus.NOT_FOUND, "Not found");
+
+        APIErrorResponse response = builder.buildStatusErrorResponse(ex, request);
+
+        assertThat(response.getStatus()).isEqualTo(404);
+        assertThat(response.getMessage()).isEqualTo("404 NOT_FOUND \"Not found\"");
+    }
+
+    @Test
+    void buildStatusErrorResponse_shouldJoinDetailMessageArguments() {
```
ivilarop: Problem: They expect the payload to contain `ex.getMessage()(404 NOT_FOUND "Not found")` or the join of `detailMessageArguments("a", "b", "c"). This forces the exposure of internal/non-i18n text.

What needs to change: The test should expect an i18n message based on the status (e.g., for 404, the `error.not_found` message; for 400, the `validation.bad_request` message), with controlled arguments if applicable.

ivilarop (changes requested):

error-response-core/src/test/java/com/itachallenge/errorcore/exceptionhandler/GlobalExceptionHandlerUnitTest.java
```diff
@@ -0,0 +1,156 @@
...
+
+    private final APIErrorResponse dummyResponse = APIErrorResponse.builder()
+            .status(400)
+            .error("Bad Request")
+            .message("dummy")
+            .build();
+
+
+    @Test
+    void handleAny_shouldReturnInternalServerError() {
```
ivilarop, line 52: Why it blocks: It requires a dangerous signature from the builder that accepts String messages (and in production, this ended up coming from e.getMessage()).
What to ask: Rewrite these tests so that the handler doesn't pass text. It should delegate to builder/service methods that don't accept Strings (e.g., fromException(e, req), fromIllegalArgument(e, req), fromInvalidFormat(e, req), etc.). In the checks, it should verify delegation and status; not messages.

error-response-core/src/test/java/com/itachallenge/errorcore/exceptionhandler/GlobalExceptionHandlerUnitTest.java
```diff
@@ -0,0 +1,156 @@
...
+                .thenReturn(dummyResponse);
+
+        ResponseEntity<APIErrorResponse> response = handler.handleAny(new Exception("boom"), request);
+
+        assertThat(response.getStatusCode()).isEqualTo(HttpStatus.INTERNAL_SERVER_ERROR);
+        verify(builder).buildError(eq(HttpStatus.INTERNAL_SERVER_ERROR), contains("boom"), eq(request));
+    }
+
+    @Test
+    void handleIllegalArgument_shouldReturnBadRequest() {
```
ivilarop, line 63: Why it blocks: It requires a dangerous signature from the builder that accepts String messages (and in production, this ended up coming from e.getMessage()).
What to ask: Rewrite these tests so that the handler doesn't pass text. It should delegate to builder/service methods that don't accept Strings (e.g., fromException(e, req), fromIllegalArgument(e, req), fromInvalidFormat(e, req), etc.). In the checks, it should verify delegation and status; not messages.

error-response-core/src/test/java/com/itachallenge/errorcore/exceptionhandler/GlobalExceptionHandlerUnitTest.java
```diff
@@ -0,0 +1,156 @@
...
+        when(builder.buildStatusErrorResponse(ex, request)).thenReturn(dummyResponse);
+
+        ResponseEntity<APIErrorResponse> response = handler.handleResponseStatusException(ex, request);
+
+        assertThat(response.getStatusCode()).isEqualTo(HttpStatus.NOT_FOUND);
+        verify(builder).buildStatusErrorResponse(ex, request);
+    }
+
+    @Test
+    void handleInvalidFormat_shouldReturnBadRequestAndDelegateToBuilder() {
```
ivilarop, line 119: Why it blocks: It requires a dangerous signature from the builder that accepts String messages (and in production, this ended up coming from e.getMessage()).
What to ask: Rewrite these tests so that the handler doesn't pass text. It should delegate to builder/service methods that don't accept Strings (e.g., fromException(e, req), fromIllegalArgument(e, req), fromInvalidFormat(e, req), etc.). In the checks, it should verify delegation and status; not messages.

error-response-core/src/test/java/com/itachallenge/errorcore/exceptionhandler/GlobalExceptionHandlerUnitTest.java
```diff
@@ -0,0 +1,156 @@
...
+
+        ResponseEntity<APIErrorResponse> response = handler.handleInvalidFormat(ex, request);
+
+        assertThat(response.getStatusCode()).isEqualTo(HttpStatus.BAD_REQUEST);
+        assertThat(response.getBody()).isEqualTo(dummyResponse);
+        verify(builder).buildError(HttpStatus.BAD_REQUEST, "Invalid or malformed request.", request);
+    }
+
+    @Test
+    void handleWebFluxBindingErrors_withInvalidFormatCause_shouldDelegateToHandleInvalidFormat() {
```
ivilarop, line 132: Why it blocks: It requires a dangerous signature from the builder that accepts String messages (and in production, this ended up coming from e.getMessage()).
What to ask: Rewrite these tests so that the handler doesn't pass text. It should delegate to builder/service methods that don't accept Strings (e.g., fromException(e, req), fromIllegalArgument(e, req), fromInvalidFormat(e, req), etc.). In the checks, it should verify delegation and status; not messages.

error-response-core/src/test/java/com/itachallenge/errorcore/exceptionhandler/GlobalExceptionHandlerUnitTest.java
```diff
@@ -0,0 +1,156 @@
...
+                .thenReturn(dummyResponse);
+
+        ResponseEntity<APIErrorResponse> response = handler.handleWebFluxBindingErrors(ex, request);
+
+        assertThat(response.getStatusCode()).isEqualTo(HttpStatus.BAD_REQUEST);
+        verify(builder).buildError(HttpStatus.BAD_REQUEST, "Invalid or malformed request.", request);
+    }
+
+    @Test
+    void handleWebFluxBindingErrors_withoutInvalidFormatCause_shouldReturnGenericBadRequest() {
```
ivilarop, line 147: Why it blocks: It requires a dangerous signature from the builder that accepts String messages (and in production, this ended up coming from e.getMessage()).
What to ask: Rewrite these tests so that the handler doesn't pass text. It should delegate to builder/service methods that don't accept Strings (e.g., fromException(e, req), fromIllegalArgument(e, req), fromInvalidFormat(e, req), etc.). In the checks, it should verify delegation and status; not messages.

error-response-core/src/test/java/com/itachallenge/errorcore/exceptionhandler/GlobalExceptionHandlerUnitTest.java
```diff
@@ -0,0 +1,156 @@
...
+
+    @Test
+    void handleAny_shouldReturnInternalServerError() {
+        when(builder.buildError(eq(HttpStatus.INTERNAL_SERVER_ERROR), anyString(), eq(request)))
+                .thenReturn(dummyResponse);
+
+        ResponseEntity<APIErrorResponse> response = handler.handleAny(new Exception("boom"), request);
+
+        assertThat(response.getStatusCode()).isEqualTo(HttpStatus.INTERNAL_SERVER_ERROR);
+        verify(builder).buildError(eq(HttpStatus.INTERNAL_SERVER_ERROR), contains("boom"), eq(request));
```
ivilarop: That contains("boom") / contains("invalid input") requires the builder to receive a String with the raw exception message, which is exactly what we want to prohibit (messages must come from i18n inside the builder/service).

error-response-core/src/test/java/com/itachallenge/errorcore/exceptionhandler/GlobalExceptionHandlerUnitTest.java
```diff
@@ -0,0 +1,156 @@
...
+    @Test
+    void handleIllegalArgument_shouldReturnBadRequest() {
+        when(builder.buildError(eq(HttpStatus.BAD_REQUEST), anyString(), eq(request)))
+                .thenReturn(dummyResponse);
+
+        ResponseEntity<APIErrorResponse> response = handler.handleIllegalArgument(
+                new IllegalArgumentException("invalid input"), request);
+
+        assertThat(response.getStatusCode()).isEqualTo(HttpStatus.BAD_REQUEST);
+        verify(builder).buildError(eq(HttpStatus.BAD_REQUEST), contains("invalid input"), eq(request));
```
ivilarop: That contains("boom") / contains("invalid input") requires the builder to receive a String with the raw exception message, which is exactly what we want to prohibit (messages must come from i18n inside the builder/service).

ivilarop (changes requested):

error-response-core/src/main/java/com/itachallenge/errorcore/exceptionhandler/GlobalExceptionHandler.java
aurelien-darbellay: done

error-response-core/src/main/java/com/itachallenge/errorcore/builder/ErrorResponseBuilder.java
aurelien-darbellay: done

error-response-core/src/main/java/com/itachallenge/errorcore/builder/ErrorResponseBuilder.java
aurelien-darbellay: done

error-response-core/build.gradle
aurelien-darbellay: I understand, but I'd suggest we do this later, at the same time as we take the version out of the other core modules in the system (jwt-core; github-core)

error-response-core/build.gradle
aurelien-darbellay: done. Had to add it as testImplementation.

error-response-core/build.gradle
aurelien-darbellay: done

error-response-core/src/main/java/com/itachallenge/errorcore/config/ErrorHandlingConfig.java
aurelien-darbellay: The core doesn't create an anonymous subclass of the handler anymore. 

I'd suggest we keep the configuration in the core for now, as in github-core, and we leave it for a further assignement to move the configuration to a boot-starter or, alternatively, in each microservice. 

Furthermore, since the goal is to have consistent error handling across microservices, it seems that having a single global configuration is pretty harmless at least for now.

error-response-core/src/main/java/com/itachallenge/errorcore/config/ErrorHandlingConfig.java
aurelien-darbellay: The core doesn't create an anonymous subclass of the handler anymore.

I'd suggest we keep the configuration in the core for now, as in github-core, and we leave it for a further assignement to move the configuration to a boot-starter or, alternatively, in each microservice.

Furthermore, since the goal is to have consistent error handling across microservices, it seems that having a single global configuration is pretty harmless at least for now.

error-response-core/src/main/java/com/itachallenge/errorcore/dto/APIErrorResponse.java
aurelien-darbellay: The Instant, LocalDateTime, OffsetDateTime, etc. classes come from Java 8’s JSR-310 date/time API.

Jackson doesn’t automatically support those types —
you must register the jackson-datatype-jsr310 module.

✅ 3. Two ways to fix this
✅ Option A: Keep using Instant, register the module globally

If you prefer Instant (the cleanest type for timestamps):

Add dependency (if not already in your microservices):

implementation 'com.fasterxml.jackson.datatype:jackson-datatype-jsr310'


Then register the module.

If you’re using Spring Boot (3.x+), it’s already auto-registered
via Jackson2ObjectMapperBuilder —
so you only need to ensure it’s on the classpath.

✅ Then your Instant will automatically serialize in ISO-8601 UTC:

"timestamp": "2025-10-30T10:04:07.123Z"

aurelien-darbellay: as we are using spring 3.0.6 across the system, we are covered.

error-response-core/src/test/java/com/itachallenge/errorcore/builder/ErrorResponseBuilderTest.java
aurelien-darbellay: done

error-response-core/src/test/java/com/itachallenge/errorcore/builder/ErrorResponseBuilderTest.java
aurelien-darbellay: done

error-response-core/src/test/java/com/itachallenge/errorcore/builder/ErrorResponseBuilderTest.java
aurelien-darbellay: done

error-response-core/src/test/java/com/itachallenge/errorcore/builder/ErrorResponseBuilderTest.java
aurelien-darbellay: done

error-response-core/src/test/java/com/itachallenge/errorcore/exceptionhandler/GlobalExceptionHandlerUnitTest.java
aurelien-darbellay: done

error-response-core/src/test/java/com/itachallenge/errorcore/exceptionhandler/GlobalExceptionHandlerUnitTest.java
aurelien-darbellay: done

error-response-core/src/test/java/com/itachallenge/errorcore/exceptionhandler/GlobalExceptionHandlerUnitTest.java
aurelien-darbellay: done

error-response-core/src/test/java/com/itachallenge/errorcore/exceptionhandler/GlobalExceptionHandlerUnitTest.java
aurelien-darbellay: done

error-response-core/src/test/java/com/itachallenge/errorcore/exceptionhandler/GlobalExceptionHandlerUnitTest.java
aurelien-darbellay: done

error-response-core/src/test/java/com/itachallenge/errorcore/exceptionhandler/GlobalExceptionHandlerUnitTest.java
aurelien-darbellay: done

error-response-core/src/test/java/com/itachallenge/errorcore/exceptionhandler/GlobalExceptionHandlerUnitTest.java
aurelien-darbellay: done

error-response-core/src/main/java/com/itachallenge/errorcore/config/ErrorHandlingConfig.java
ivilarop: Okay, we accept the current configuration in the core as a temporary measure to maintain consistency across microprocessors.

**Known risks:** coupling when registering web beans from the core and potential double registration/limitation of customization per microprocessor.

When possible, please note the technical requirements for moving this configuration to a Spring Boot starter (auto-config, opt-in) and removing web dependencies from the core. Include this information in the pull request description.

error-response-core/src/main/java/com/itachallenge/errorcore/dto/APIErrorResponse.java
ivilarop: Perfect, confirmed then.

error-response-core/src/test/java/com/itachallenge/errorcore/exceptionhandler/GlobalExceptionHandlerUnitTest.java
ivilarop: Ok you can resolve review ;)

error-response-core/src/test/java/com/itachallenge/errorcore/exceptionhandler/GlobalExceptionHandlerUnitTest.java
ivilarop: Ok you can resolve review ;)

error-response-core/src/test/java/com/itachallenge/errorcore/exceptionhandler/GlobalExceptionHandlerUnitTest.java
ivilarop: Ok you can resolve review ;)

error-response-core/src/test/java/com/itachallenge/errorcore/exceptionhandler/GlobalExceptionHandlerUnitTest.java
ivilarop: Ok you can resolve review ;)

error-response-core/src/test/java/com/itachallenge/errorcore/exceptionhandler/GlobalExceptionHandlerUnitTest.java
ivilarop: Ok you can resolve review ;)

error-response-core/build.gradle
ivilarop: Okay, we'll leave that for when we unify dependency management for all core modules in the root dependencyManagement.

Add a technical debt note for that global task. Will it be possible to create only one technical debt with this improvement and what was discussed earlier regarding ErrorHandlingConfig?

ivilarop (approved):

PR #862, opened 2025-05-05
Feature/277 creation method endpoint solved controller

RustyGearBox:
A new controller has been added '`ChallengeSolvedController.java`' in this PR for the solvedCounter feature.

The `addChallengeToSolved` method is a Spring WebFlux POST endpoint that marks a challenge as solved for a user. It extracts the `challengeId` from the URI and the user ID from the `Authorization` header using `jwtService`. Errors like invalid headers or missing challenges are handled and logged.

PR #854, opened 2025-04-30
Add MarcBernabeuRodriguez to contributors.yml

trisk910:
Added my name in contributors.md

PR #1021, opened 2025-11-12
PR 2/5 - refactor(user): Implement FavoriteService (#841)

itsAidenDev:
This PR corresponds to Task #841 and introduces the new FavoriteService layer, responsible for handling all logic related to user favorites.
It represents the third step in the ongoing refactor towards decoupling user interaction data from the UserService.

The purpose of this implementation is to centralize favorite-related operations into a dedicated service, improving modularity, scalability, and testability of the codebase.

It follows the work completed in task #840 — Creation of Favorite documents and repositories.

**🔹 Changes Introduced**

Created the new FavoriteService interface and FavoriteServiceImpl implementation class.

Implemented the main method:

`    public Mono<Set<UUID>> getUserFavorites(String userId)
`

This method, getUserFavorites, retrieves a set of favorite challenge IDs for a given user. Here's a breakdown of its functionality:

1. It takes a userId as a string parameter and returns a Mono<Set<UUID>> (a reactive type that will eventually emit a set of UUIDs).
2. The method performs the following operations:
- First, it validates the input userId using parseAndValidateUUID()
- Then, it checks if a user with that ID exists in the database using userRepository.existsById()
- If the user doesn't exist, it returns a Mono.error with a NotFoundException
- If the user exists, it fetches all favorites for that user using favoriteRepository.findByUserId()
- It transforms the result to extract just the challenge IDs using map(FavoriteDocument::getChallengeId)
- Finally, it collects these IDs into a Set<UUID>, or returns an empty set if the user has no favorites

Logic mirrors the previous favorite retrieval mechanism inside UserService, but is now fully decoupled and repository-driven.

Ensured that all operations are reactive-friendly (Flux/Mono) to maintain consistency with the reactive stack.

**🔹 Advantages**

Promotes separation of concerns, isolating favorites logic from user management.

Improves maintainability and testability of the backend architecture.

Sets the foundation for future tasks involving the GET controller (Task #842) and integration tests (Task #843).

Ensures a reactive, non-blocking design aligned with the new interactions architecture.

**✅ Created / Updated Files**

FavoriteService.java

FavoriteServiceImpl.java

**💡 Notes**

The service implementation directly uses the FavoriteRepository introduced in Task #840.

Future controller logic (Task #842) will consume this service to expose REST endpoints for retrieving user favorites.

Technical debt: [#854] — Use ErrorCore in the new Favorite Service so that exceptions return ErrorDto instead of strings.

Tecnical debt: [#862] — [BE] Optimize User Existence Check in getUserFavorites.

**✅ PR Author Self-Check**

- [x] I tested my changes locally and verified they work as expected.
- [x] The PR has a clear and focused purpose.
- [x] Title and description are filled in correctly.
- [x] There are no leftover merge conflicts or unrelated changes from previous branches.
- [x] All automated checks (like SonarCloud) have passed.
- [x] I responded to all previous review comments and only resolved those I truly addressed.
- [x] I ran SonarLint locally on the modified files and confirmed there are no warnings or errors
- [x] changelog.md updated


FlavioKde (commented):
Reviewing the history, it appears that `getUserFavorites` was correctly removed from `UserServiceImpl` but then reintroduced. I'm not sure if this is correct.

Since the logic is already implemented in `FavoriteServiceImpl`, this method should be removed from both `UserServiceImpl` and its interface (`UserService`) to avoid duplication and comply with HU #841.

Please correct me if I'm wrong.

Once this is done, the pull request would be ready for validation.

Pablo2203 (changes requested):
Hello Aiden! 
Hope you are doing great.
First of all, good job! I just have one concern regarding the class itachallenge-user/src/main/java/com/itachallenge/user/service/UserServiceImpl.java

Detected issues:

- [ ] Concatenated message bug in getUserFavorites Problem: USER_NOT_FOUND + userId is used, where USER_NOT_FOUND is "User not found" (without a separator). 

This produces messages like "User not foundf3b..." and also does not use the specific constant for IDs. Fix: use USER_NOT_FOUND_WITH_ID + userId. 

Example replacement: public Mono<Set<UUID>> getUserFavorites(String userId) { return parseAndValidateUUID(userId) .flatMap(userUuid -> userRepository.findById(userUuid) .switchIfEmpty(Mono.error(new NotFoundException(USER_NOT_FOUND_WITH_ID + userId))) .map(user -> Optional.ofNullable(user.getFavoriteChallenges()).orElseGet(HashSet::new)) ); }

- [ ] UUID parsing/validation in getUserById Problem: getUserById uses UUID.fromString(userId) directly. If userId is not a valid UUID it will throw IllegalArgumentException before entering the reactive flow and before you can return a controlled error (and the current handler maps IllegalArgumentException to 404, which may not be desired).  

Recommended fix: reuse parseAndValidateUUID (already available in the class) to normalize validation and be able to return 400 (BadUUIDException) or 404 depending on the semantics you prefer. Example replacement: public Mono<UserDocument> getUserById(String userId) { return parseAndValidateUUID(userId) .flatMap(userUuid -> userRepository.findById(userUuid) .switchIfEmpty(Mono.error(new NotFoundException(USER_NOT_FOUND_WITH_ID + userId))) ); }

Make sure parseAndValidateUUID emits Mono.error(new BadUUIDException(...)) when the id is invalid if you want the handler to return 400. If it currently throws IllegalArgumentException, consider changing it to BadUUIDException for clarity.

itsAidenDev:
@FlavioKde Fixed! The method has been removed from both UserService and UserServiceImpl. Logic now resides in FavoriteService and FavoriteServiceImpl as intended. Thanks for catching that!

itsAidenDev:
@Pablo2203 Thanks for the feedback, Pablo! I appreciate the thorough review. 

The first issue has been addressed — the message now correctly uses the constant USER_NOT_FOUND_WITH_ID instead of the hardcoded "User not found with id:".

Regarding the second point about getUserById, that change is currently out of scope for this PR, since this task focuses specifically on the FavoriteService implementation and related logic.

FlavioKde (approved):
Ok guys, gooddd job!!

Pablo2203 (approved):

ivilarop:
@itsAidenDev @Pablo2203 @FlavioKde 

The PR title does not meet the checklist.

itsAidenDev:
@ivilarop Done! I’ve just updated the PR title now

itachallenge-user/src/main/java/com/itachallenge/user/service/UserServiceImpl.java
```diff
@@ -18,6 +18,9 @@
 @Service
 public class UserServiceImpl implements UserService {
 
+    private static final String USER_NOT_FOUND = "User not found";
```
ivilarop: Are these changes part of the scope?

itachallenge-user/src/main/java/com/itachallenge/userinteraction/service/favorite/FavoriteServiceImpl.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    @Override
+    public Mono<Set<UUID>> getUserFavorites(String userId) {
+        return parseAndValidateUUID(userId)
+                .flatMap(userUuid ->
+                        userRepository.existsById(userUuid)
+                                .flatMap(exists -> {
+                                    if (!exists) {
+                                        return Mono.error(new NotFoundException(USER_NOT_FOUND_WITH_ID + userId));
```
ivilarop, line 46: Good work reusing the existing user exceptions in this pull request. To avoid future duplication and deviations, once the new shared exceptions module is merged, we should consolidate all domain-level exceptions there and update the "user interaction" usages accordingly.

Please create a brief technical debt ticket and attach it to the PR description, indicating that it remains pending.

itachallenge-user/src/main/java/com/itachallenge/userinteraction/service/favorite/FavoriteServiceImpl.java
```diff
@@ -0,0 +1,55 @@
...
+                                    }
+                                    return favoriteRepository.findByUserId(userUuid)
+                                            .map(FavoriteDocument::getChallengeId)
+                                            .collect(Collectors.toSet())
+                                            .defaultIfEmpty(Set.of()); // user exists but no favorites
```
ivilarop: `collect(Collectors.toSet())` turns the Flux into a Monkey that always emits a Set (if the Flux was empty, the Set will also be empty). `defaultIfEmpty(...)` only applies if that Monkey doesn't emit any value. Since `collect(...)` always emits one, `defaultIfEmpty(Set.of())` never comes into play and can be removed.

itachallenge-user/src/main/java/com/itachallenge/userinteraction/service/favorite/FavoriteServiceImpl.java
```diff
@@ -0,0 +1,55 @@
...
+
+        try {
+            return Mono.just(UUID.fromString(id));
+        } catch (IllegalArgumentException ex) {
+            return Mono.error(new BadUUIDException("Invalid ID format"));
+        }
+    }
+
+    @Override
+    public Mono<Set<UUID>> getUserFavorites(String userId) {
```
ivilarop, line 40: Currently, we first check `users.existsById(userId)` and then read `favorites`. I propose reversing the order:

1. Check `favorites` by `userId` first.

    - If it returns results, we respond with a 200 value with the list of IDs and don't call `users`.

2. Only if `favorites` is empty should we call `users.existsById(userId)` to distinguish between:

    - `true` → 200 value with an empty list (user exists without favorites).
    - `false` → 404 value (user doesn't exist).

**Why this is better:**

- Cheaper happy path: most users with favorites avoid an extra call to `users`.
- Lower latency and less I/O: we only access `users` when it's empty.
- Contract intact: we still differentiate between 200 and 404 without ambiguity.
- Clear semantics: user existence only matters when there is no data.

ivilarop (changes requested):

itachallenge-user/src/main/java/com/itachallenge/userinteraction/service/favorite/FavoriteServiceImpl.java
itsAidenDev: Thanks! We’ve created a technical debt (ticket #854) to use the ErrorCore in the new Favorite Service so that exceptions return ErrorDto instead of strings. I’ve also indicated it in the PR description as well.

itachallenge-user/src/main/java/com/itachallenge/userinteraction/service/favorite/FavoriteServiceImpl.java
ivilarop: Technical Debt: https://tree.taiga.io/project/luisvi-ita-challenges/issue/862

ivilarop (approved):

PR #1025, opened 2025-11-13
chore(contributors): add Alex Perez to contributors.md (#857)

aperez1haad:
Add Alex Pérez to contributors.md like onboarding wiki description


eduzavarce (changes requested):
@aperez1haad 
* The branch name should be as stated on the wiki:
image

* The PR name should also follow the conventions as tin the same file:

image

PR #1069, opened 2026-01-22
Fix(submission) 400 Bad Request on "Show Solution" with Empty Input (#34)

aperez1haad:
**Context:**

In the current implementation of the Challenges module, users have the option to Save, Finish, or Show Solution. Currently, when a user clicks on "Show Solution" without having typed any text in the input field, the server returns a 400 Bad Request error. This happens because the backend (Java MVC) expects a non-empty string or a specific data structure that is not being met when the field is blank.

**Objective:**

1. Modify the validation logic in the Controller and/or the Request DTO and implements on UserSolutionServiceImpl and allow null or empty string values specifically for the "Show Solution" action. The system should process the request successfully and display the solution regardless of whether the user has provided a draft answer.

With this changes:

SHOW_SOLUTION → doesn't require text

SAVE → can be empty (save progress)

FINISH → should require text

2. Now, the request is vinculated to Challenge Microservice and if it is down, the request save correctly to finish button but we have the 500 server internal error. To avoid this, this pr implements OnErrorResume.

3. Actualize Tests and postman works correctly

**Impact:**

User Experience: Improves the flow by allowing users to skip a challenge and see the solution immediately without being forced to type placeholder text.

Stability: Eliminates unhandled 400 errors in the frontend logs, leading to a cleaner and more professional application behavior.

API Consistency: Aligns the API behavior with the functional requirement that "Show Solution" is an informative action, not necessarily dependent on prior input.

This improve will be apply on the branch fix(submission)-invalid-blank-submission-showsolution

Captura de pantalla 2026-01-28 a las 11 50 20

Captura de pantalla 2026-01-28 a las 11 51 36
Captura de pantalla 2026-01-28 a las 12 52 59

image

image
image
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-### [itachallenge-user-3.1.3-RELEASE] - 2026-01-27
+### [itachallenge-user-3.1.4-RELEASE] - 2026-01-27
 
 ### Changed
 - Updated UserSolutionRequestDto to include blank solution possibility to the User
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,10 +3,11 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-### [itachallenge-user-3.1.4-RELEASE] - 2026-01-27
+### [itachallenge-user-3.2.1-RELEASE] - 2026-02-10
 
 ### Changed
 - Updated UserSolutionRequestDto to include blank solution possibility to the User
+- 
 ### [itachallenge-user-3.2.0-RELEASE] - 2026-02-05
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### [itachallenge-user-3.1.3-RELEASE] - 2026-01-27
+
+### Changed
+- Updated UserSolutionRequestDto to include blank solution possibility to the User
+
 ### [itachallenge-challenge-3.1.0-RELEASE] - 2026-01-14
 
 ### Changed
--- a/itachallenge-user/src/main/java/com/itachallenge/user/dto/UserSolutionRequestDto.java
+++ b/itachallenge-user/src/main/java/com/itachallenge/user/dto/UserSolutionRequestDto.java
@@ -3,7 +3,6 @@ package com.itachallenge.user.dto;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.itachallenge.user.annotations.GenericUUIDValid;
 import com.itachallenge.user.document.enums.SolutionAction;
-import jakarta.validation.constraints.NotBlank;
 import jakarta.validation.constraints.NotNull;
 import lombok.*;
 
--- a/itachallenge-user/src/main/java/com/itachallenge/user/dto/UserSolutionRequestDto.java
+++ b/itachallenge-user/src/main/java/com/itachallenge/user/dto/UserSolutionRequestDto.java
@@ -33,7 +33,7 @@ public class UserSolutionRequestDto {
     private SolutionAction action;
 
     @JsonProperty(value ="solution_text")
-    @NotBlank(message = "Solution text is required")
+    @NotNull(message = "Solution cannot be null")
     private String solutionText;
 
 }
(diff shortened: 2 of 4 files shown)


DiegoBalaguer (approved):
Good job Alex

anaberod (approved):
Well done Alex

ivilarop:
@aperez1haad @DiegoBalaguer @anaberod 

The PR title is missing details about the GitHub task

itachallenge-user/src/main/java/com/itachallenge/user/dto/UserSolutionRequestDto.java
```diff
@@ -33,7 +32,7 @@ public class UserSolutionRequestDto {
     private SolutionAction action;
 
     @JsonProperty(value ="solution_text")
-    @NotBlank(message = "Solution text is required")
+    @NotNull(message = "Solution cannot be null")
```
ivilarop: There's an important design detail to consider. The UserSolutionRequestDto is used for several different actions (SAVE, SHOW_SOLUTION, and FINISH), but each has different rules regarding the solutionText.

SHOW_SOLUTION → doesn't require text

SAVE → can be empty (save progress)

FINISH → should require text

Therefore, having global validation in the DTO doesn't quite fit with the business model. Previously, @NotBlank was too restrictive (it broke SHOW_SOLUTION). Now, @NotNull is too permissive (it allows finishing with an empty string). Validation should really be conditional based on the action.

It's recommended to move this rule to the service/controller, for example: validate that solutionText is only required when action == FINISH, checking for null or blank.

Furthermore, the attached test in Postman returns a 409 Conflict ("already submitted"), so the DTO validation isn't being executed. To properly validate the "finishing with an empty solution" scenario, you should test with a challenge in IN_PROGRESS status or without a prior submission.

In short, the change improves the UX for "Show Solution," but it would be ideal to adjust the validation to depend on the action rather than applying it globally to the DTO.

CHANGELOG.md
```diff
@@ -3,6 +3,11 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### [itachallenge-user-3.1.3-RELEASE] - 2026-01-27
```
ivilarop: It's a bad version.

ivilarop (changes requested):

itachallenge-user/src/main/java/com/itachallenge/user/service/UserSolutionServiceImpl.java
```diff
@@ -97,7 +102,15 @@ private Mono<SubmitSolutionResponseDto> buildSubmitSolutionResponse(UserSolution
                             .isSolved(solvedDto.isSolved())
                             .timesSolved(solvedDto.getTimesSolved())
                             .status(status.name())
-                            .build());
+                            .build())
+                    .onErrorResume(e -> {
```
ivilarop: Why this change? I don't see it described in the PR.

ivilarop (changes requested):
conversations are missing responses and see comment in UserSolutionImpl

CHANGELOG.md
aperez1haad: changed

itachallenge-user/src/main/java/com/itachallenge/user/dto/UserSolutionRequestDto.java
aperez1haad: changed, now this filter are in the UserSolutionServiceImpl

itachallenge-user/src/main/java/com/itachallenge/user/service/UserSolutionServiceImpl.java
aperez1haad: Now when you POST the json of the UserSolution to the Database, the connection with Challenge Microservice can go wrong and if Challenges microservice is down throws a 500 Interal Server Error. However,  the database saves the request. To avoid this, i implement this code to catch the error if the Challenges Microservice is down because the request goes well.

ivilarop: Is this change within the scope of the task? Could it be considered technical debt with more time to analyze whether it's the right solution?

ivilarop (changes requested):

itachallenge-user/src/main/java/com/itachallenge/user/service/UserSolutionServiceImpl.java
aperez1haad: yes, its out oh scope and its not conmpletely necessary. So i delete this part of code.

itachallenge-user/src/main/java/com/itachallenge/user/dto/UserSolutionRequestDto.java
ivilarop: ok thks! well done!

CHANGELOG.md
```diff
@@ -3,6 +3,10 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### [itachallenge-user-3.1.4-RELEASE] - 2026-01-27
```
ivilarop: incorrect version

ivilarop (changes requested):

ivilarop (approved):

CHANGELOG.md
ivilarop: done! thks!

itachallenge-user/src/main/java/com/itachallenge/user/service/UserSolutionServiceImpl.java
ivilarop: ok thks!

CHANGELOG.md
ivilarop: ok thks!

PR #1070, opened 2026-01-22
PR 1/2 - feat(challenge): add submission POST business logic (service) (#73)

anaberod:
🎯 Summary

This PR corresponds to subtask #73 of issue #22.

It implements the business logic required to replicate the submission POST behavior from the User microservice into the Challenge microservice, at service level only.

🛑 Context

Currently, the submission POST logic resides in the User microservice.
As part of a gradual migration strategy, this logic needs to be replicated in the Challenge microservice, without changing the current production flow.

🧠 Solution

This PR prepares the Challenge microservice by implementing:
- Submission POST business rules (SAVE / SUBMIT / GIVE_UP)
- Action → status mapping inside the SubmissionAction enum
- Domain validation and blocking of completed submissions
- Side-effect call to increment solved challenges only when a submission is completed

No HTTP endpoint is exposed in this PR.

🔹 Scope of this PR

- Service-layer business logic for submission POST
- New DTOs and enums required by the logic
- Repository finder to support overwrite behavior
- Domain exception for unmodifiable submissions

🚫 Out of scope

- Submission controller and HTTP endpoint
- Swagger / OpenAPI updates
- WebTestClient tests
- Frontend or User microservice changes

🔹 This task is part of a 2-PR series:

- PR 1/2 (this PR): Submission POST business logic (service)
- PR 2/2: Submission POST controller, validations, Swagger and tests

🔄 Changelog

- Added submission POST business logic to Challenge service layer
- Added SubmissionAction enum and submission POST DTOs
- Added repository finder to support submission overwrite behavior

✅ PR Author Self-Check

- [ ] Changes tested locally
- [x] PR has a clear and focused purpose
- [x] No unrelated changes included
- [x] Sonar / CI checks passing

### DIFF
--- a/itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/submission/SubmissionActionRequestDto.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/submission/SubmissionActionRequestDto.java
@@ -12,7 +12,7 @@ import java.util.UUID;
 @Getter
 @Setter
 @Builder
-public class SubmissionRequestDto {
+public class SubmissionActionRequestDto {
 
     @NotNull
     @JsonProperty(value = "uuid_challenge")
--- a/itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/submission/SubmissionRequestDto.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/submission/SubmissionRequestDto.java
@@ -1,0 +1,24 @@
+package com.itachallenge.challenge.dto.submission;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.*;
+
+@AllArgsConstructor
+@NoArgsConstructor
+@Getter
+@Setter
+@Builder
+public class SubmissionRequestDto {
+
+    @JsonProperty(value = "uuid_challenge")
+    private String challengeId;
+
+    @JsonProperty(value = "uuid_language")
+    private String languageId;
+
+    @JsonProperty("action")
+    private String action;
+
+    @JsonProperty(value = "submission_text")
+    private String submissionText;
+}
--- a/itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/submission/SubmissionRequestDto.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/submission/SubmissionRequestDto.java
@@ -1,6 +1,7 @@
 package com.itachallenge.challenge.dto.submission;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.itachallenge.submission.enums.SubmissionAction;
 import jakarta.validation.constraints.NotNull;
 import lombok.*;
 
@@ -23,7 +24,7 @@ public class SubmissionRequestDto {
 
     @NotNull
     @JsonProperty("action")
-    private String action;
+    private SubmissionAction action;
 
     @NotNull
     @JsonProperty(value = "submission_text")
--- a/itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/submission/SubmissionRequestDto.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/submission/SubmissionRequestDto.java
@@ -26,7 +26,6 @@ public class SubmissionRequestDto {
     @JsonProperty("action")
     private SubmissionAction action;
 
-    @NotNull
     @JsonProperty(value = "submission_text")
     private String submissionText;
 }
--- a/itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/submission/SubmissionRequestDto.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/submission/SubmissionRequestDto.java
@@ -1,7 +1,6 @@
 package com.itachallenge.challenge.dto.submission;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
-import com.itachallenge.challenge.annotations.ValidUUID;
 import jakarta.validation.constraints.NotNull;
 import lombok.*;
 
@@ -15,12 +14,10 @@ import java.util.UUID;
 public class SubmissionRequestDto {
 
     @NotNull
-    @ValidUUID
     @JsonProperty(value = "uuid_challenge")
     private UUID challengeId;
 
     @NotNull
-    @ValidUUID
     @JsonProperty(value = "uuid_language")
     private UUID languageId;
 
--- a/itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/submission/SubmissionRequestDto.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/submission/SubmissionRequestDto.java
@@ -1,7 +1,11 @@
 package com.itachallenge.challenge.dto.submission;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.itachallenge.challenge.annotations.ValidUUID;
+import jakarta.validation.constraints.NotNull;
 import lombok.*;
+
+import java.util.UUID;
 
 @AllArgsConstructor
 @NoArgsConstructor
@@ -10,15 +14,21 @@ import lombok.*;
 @Builder
 public class SubmissionRequestDto {
 
+    @NotNull
+    @ValidUUID
     @JsonProperty(value = "uuid_challenge")
-    private String challengeId;
+    private UUID challengeId;
 
+    @NotNull
+    @ValidUUID
     @JsonProperty(value = "uuid_language")
-    private String languageId;
+    private UUID languageId;
 
+    @NotNull
     @JsonProperty("action")
     private String action;
 
+    @NotNull
     @JsonProperty(value = "submission_text")
     private String submissionText;
 }
--- a/itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/submission/SubmissionResponseDto.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/submission/SubmissionResponseDto.java
@@ -1,0 +1,24 @@
+package com.itachallenge.challenge.dto.submission;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.*;
+
+@AllArgsConstructor
+@NoArgsConstructor
+@Getter
+@Setter
+@Builder
+public class SubmissionResponseDto {
+
+    @JsonProperty(value = "submission_text")
+    private String submissionText;
+
+    @JsonProperty(value = "is_solved")
+    private Boolean isSolved;
+
+    @JsonProperty(value = "times_solved")
+    private Integer timesSolved;
+
+    @JsonProperty("status")
+    private String status;
+}
--- a/itachallenge-challenge/src/main/java/com/itachallenge/submission/service/SubmissionService.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/submission/service/SubmissionService.java
@@ -1,7 +1,7 @@
 package com.itachallenge.submission.service;
 
 import com.itachallenge.challenge.dto.submission.SubmissionDto;
-import com.itachallenge.challenge.dto.submission.SubmissionRequestDto;
+import com.itachallenge.challenge.dto.submission.SubmissionActionRequestDto;
 import com.itachallenge.challenge.dto.submission.SubmissionActionResponseDto;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
@@ -10,7 +10,7 @@ public interface SubmissionService {
 
     Flux<SubmissionDto> getAllSubmissionsByUser(String userId);
 
-    Mono<SubmissionActionResponseDto> processSubmissionAction(String userId, SubmissionRequestDto request);
+    Mono<SubmissionActionResponseDto> processSubmissionAction(String userId, SubmissionActionRequestDto request);
 
 
 }
--- a/itachallenge-challenge/src/main/java/com/itachallenge/submission/service/SubmissionService.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/submission/service/SubmissionService.java
@@ -10,6 +10,7 @@ public interface SubmissionService {
 
     Flux<SubmissionDto> getAllSubmissionsByUser(String userId);
 
-    Mono<SubmissionActionResponseDto> createOrUpdateSubmission(String userId, SubmissionRequestDto request);
+    Mono<SubmissionActionResponseDto> processSubmissionAction(String userId, SubmissionRequestDto request);
+
 
 }
--- a/itachallenge-challenge/src/main/java/com/itachallenge/submission/service/SubmissionService.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/submission/service/SubmissionService.java
@@ -2,8 +2,7 @@ package com.itachallenge.submission.service;
 
 import com.itachallenge.challenge.dto.submission.SubmissionDto;
 import com.itachallenge.challenge.dto.submission.SubmissionRequestDto;
-import com.itachallenge.challenge.dto.submission.SubmissionResponseDto;
-import com.itachallenge.submission.document.SubmissionDocument;
+import com.itachallenge.challenge.dto.submission.SubmissionActionResponseDto;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
@@ -11,6 +10,6 @@ public interface SubmissionService {
 
     Flux<SubmissionDto> getAllSubmissionsByUser(String userId);
 
-    Mono<SubmissionResponseDto> createOrUpdateSubmission(String userId, SubmissionRequestDto request);
+    Mono<SubmissionActionResponseDto> createOrUpdateSubmission(String userId, SubmissionRequestDto request);
 
 }
--- a/itachallenge-challenge/src/main/java/com/itachallenge/submission/service/SubmissionService.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/submission/service/SubmissionService.java
@@ -1,10 +1,16 @@
 package com.itachallenge.submission.service;
 
 import com.itachallenge.challenge.dto.submission.SubmissionDto;
+import com.itachallenge.challenge.dto.submission.SubmissionRequestDto;
+import com.itachallenge.challenge.dto.submission.SubmissionResponseDto;
 import com.itachallenge.submission.document.SubmissionDocument;
 import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
 
 public interface SubmissionService {
 
     Flux<SubmissionDto> getAllSubmissionsByUser(String userId);
+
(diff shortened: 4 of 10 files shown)


itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/submission/SubmissionActionRequestDto.java
```diff
@@ -0,0 +1,24 @@
+package com.itachallenge.challenge.dto.submission;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.*;
+
+@AllArgsConstructor
```
aperez1haad, line 10: Why the request dont' have GenericUUIDValid or NotNull  like UserSolutionRequestDto? Is replacing this request? Why you move this logic to service?

aperez1haad (commented):
Hi Ana. this changes works good? Its a good point to include Postman screenshots

itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/submission/SubmissionActionRequestDto.java
anaberod: Thanks for the review, Alex. I’ve now aligned SubmissionRequestDto with the User request style by adding @NotNull + UUID validation and using @Valid in the controller.
This does not replace the current User endpoint; it mirrors the behavior in Challenge for a future switch of callers.
Validation/parsing and action handling live in the service to keep business logic in one place. I’ll include Postman screenshots as suggested.

itachallenge-challenge/src/main/java/com/itachallenge/submission/service/SubmissionServiceImpl.java
```diff
@@ -27,6 +37,74 @@ public Flux<SubmissionDto> getAllSubmissionsByUser(String userId) {
                 );
     }
 
+    @Override
+    public Mono<SubmissionResponseDto> createOrUpdateSubmission(String userId, SubmissionRequestDto request) {
+
+        Mono<UUID> userUuidMono = validateAndParseUuid(userId);
+
```
DiegoBalaguer, line 44: Since submission RequestDto comes from a @Valid in the controller, check that the null validation is not being duplicated in the dto and in the method.

DiegoBalaguer (changes requested):

DiegoBalaguer (commented):
In the title of the PR you have put the general task number (22) at the end, but you should actually put the subtask number.

DiegoBalaguer (approved):
In the title of the PR you have put the general task number (22) at the end, but you should actually put the subtask number.

aperez1haad (approved):
Good job Ana

itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/submission/SubmissionResponseDto.java
```diff
@@ -0,0 +1,24 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.*;
+
+@AllArgsConstructor
+@NoArgsConstructor
+@Getter
+@Setter
+@Builder
+public class SubmissionResponseDto {
```
ivilarop: -----------
1 Naming DTO
-------------

I would suggest adjusting the naming of the SubmissionResponseDto.

Currently, we have:

SubmissionDto → used as the GET response (represents the "submission" resource)

SubmissionResponseDto → used as the POST response (submit action)

The problem is that the name SubmissionResponseDto sounds like a generic response for the resource, when in reality it only represents the result of an action (submit/save/give_up) and not the same model returned by GET (it doesn't contain uuid_user, uuid_challenge, etc.).

In real-world APIs, when GET and POST return different models, it's better for the name to reflect the operation to avoid confusion.

That's why I suggest renaming it to something more specific, for example:

SubmissionActionResponseDto or

SubmitSubmissionResponseDto

This makes it clear that:

GET → resource (SubmissionDto)

POST → action result (Action/Submit response)

Ideally, we could unify both responses into one, but in this case, it's not possible because the fields are different and represent different things (resource vs. result).

It's just a name change, but it greatly helps keep the API clearer and more maintainable in the future. 🙂

----------------
2 Create DTOs in PR bussines logic
----------------

We're not going to move the DTOs, but for future improvements:

DTOs with @JsonProperty/validations are HTTP contracts, and that belongs to PR 2/2 (controller + validations + swagger + tests).

Why (business criteria)?

In a service-only PR, you want it to be deploy-safe and transport-aggregated.

A DTO with @JsonProperty, @NotNull, and @ValidUUID is already clearly part of the API layer.

Mixing it here makes PR 1 no longer "logic-only" and complicates the review (and the split becomes pointless).

What alternative to use in PR 1?

In the service, you can use an internal model (command) without HTTP annotations, for example:

SubmissionCommand(userId, challengeId, languageId, action, submissionText)

or even direct parameters.

And in PR 2 you map:
SubmissionRequestDto → SubmissionCommand
and CommandResult → SubmissionActionResponseDto.

itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/submission/SubmissionRequestDto.java
```diff
@@ -0,0 +1,34 @@
...
+import lombok.*;
+
+import java.util.UUID;
+
+@AllArgsConstructor
+@NoArgsConstructor
+@Getter
+@Setter
+@Builder
+public class SubmissionRequestDto {
```
ivilarop: -----------------
1 Rename suggestion (request DTO)
------------------------------------

Small naming suggestion for clarity 🙂

We currently have:

SubmissionDto → used in GET responses (resource model)

SubmissionRequestDto → intended for the POST action-based endpoint

The name SubmissionRequestDto is a bit too generic: it sounds like “the request DTO for any submission endpoint”, but this request is specifically for the action workflow (SAVE / SUBMIT / GIVE_UP).

To avoid ambiguity as the API grows, it’s better if the name reflects the operation.

👉 Suggested rename:

SubmitSubmissionRequestDto

That way it’s immediately clear that this DTO is tied to the action/command endpoint, not a generic “submission request”.

----------------
2 Create DTOs in PR bussines logic
----------------

We're not going to move the DTOs, but for future improvements:

DTOs with @JsonProperty/validations are HTTP contracts, and that belongs to PR 2/2 (controller + validations + swagger + tests).

Why (business criteria)?

In a service-only PR, you want it to be deploy-safe and transport-aggregated.

A DTO with @JsonProperty, @NotNull, and @ValidUUID is already clearly part of the API layer.

Mixing it here makes PR 1 no longer "logic-only" and complicates the review (and the split becomes pointless).

What alternative to use in PR 1?

In the service, you can use an internal model (command) without HTTP annotations, for example:

SubmissionCommand(userId, challengeId, languageId, action, submissionText)

or even direct parameters.

And in PR 2 you map:
SubmissionRequestDto → SubmissionCommand
and CommandResult → SubmissionActionResponseDto.

itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/submission/SubmissionActionRequestDto.java
```diff
@@ -0,0 +1,34 @@
...
+@NoArgsConstructor
+@Getter
+@Setter
+@Builder
+public class SubmissionRequestDto {
+
+    @NotNull
+    @ValidUUID
+    @JsonProperty(value = "uuid_challenge")
+    private UUID challengeId;
```
ivilarop, line 19: Small suggestion about the UUID fields 🙂

I see you changed challengeId/languageId from String to UUID (which is actually a good improvement 👍).

Just note that if the type is already UUID, the custom validation (@ValidUUID / similar) becomes redundant. Spring/Jackson already validates the format while deserializing and will return 400 Bad Request automatically if the value is not a valid UUID.

So we can simplify to:

UUID + @NotNull

and remove the custom UUID validator.

(Custom UUID validators only make sense when the field type is String.)

If you want to be extra safe and ensure we don’t get any “weird” errors, we can add a small controller test (WebTestClient) sending an invalid UUID and assert we return a clean 400 response.

itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/submission/SubmissionActionRequestDto.java
```diff
@@ -0,0 +1,34 @@
...
+    @JsonProperty(value = "uuid_language")
+    private UUID languageId;
+
+    @NotNull
+    @JsonProperty("action")
+    private String action;
+
+    @NotNull
+    @JsonProperty(value = "submission_text")
+    private String submissionText;
```
ivilarop, line 30: Have you been able to talk to Alex about the change he made in https://github.com/IT-Academy-BCN/ita-challenges/issues/34 regarding this field? Can we implement the change in this pull request or should we leave it as technical debt?

itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/submission/SubmissionRequestDto.java
```diff
@@ -0,0 +1,34 @@
...
+    private UUID challengeId;
+
+    @NotNull
+    @ValidUUID
+    @JsonProperty(value = "uuid_language")
+    private UUID languageId;
+
+    @NotNull
+    @JsonProperty("action")
+    private String action;
```
ivilarop: action would be better as an enum (SubmissionAction) instead of String. This prevents invalid values, improves Swagger docs, and lets Spring return 400 automatically when an unsupported action is sent.

itachallenge-challenge/src/main/java/com/itachallenge/submission/service/SubmissionServiceImpl.java
```diff
@@ -35,4 +113,11 @@ private Mono<UUID> validateAndParseUuid(String userId) {
                 .onErrorMap(IllegalArgumentException.class,
                         ex -> new BadRequestException("The 'userId' parameter must be a valid UUID."));
     }
-}
+
+    private Mono<UUID> validateAndParseUuid(String value, String fieldName) {
```
ivilarop: What is the purpose of this method?

itachallenge-challenge/src/main/java/com/itachallenge/submission/service/SubmissionServiceImpl.java
ivilarop: @anaberod @DiegoBalaguer 

Conversations cannot be resolved without adding the changes made.

itachallenge-challenge/src/main/java/com/itachallenge/submission/service/SubmissionServiceImpl.java
```diff
@@ -27,6 +37,74 @@ public Flux<SubmissionDto> getAllSubmissionsByUser(String userId) {
...
+                            .findByUserIdAndChallengeIdAndLanguageId(userUuid, challengeUuid, languageUuid)
+                            .flatMap(existing -> {
+                                if (existing.getStatus() == SubmissionStatus.SUBMITTED_COMPLETE
+                                        || existing.getStatus() == SubmissionStatus.SUBMITTED_INCOMPLETE) {
+                                    return Mono.error(new UnmodifiableSubmissionException(
+                                            "Submission cannot be modified once submitted."));
+                                }
+
+                                existing.setStatus(targetStatus);
+                                existing.setSubmissionText(request.getSubmissionText());
```
ivilarop, line 77: We’re replicating the behavior from the User endpoint. In User (PR #1069 / issue #34) solution_text is optional for SAVE/GIVE_UP and required only for SUBMIT (the validation was moved to the service).

To keep both endpoints consistent, we should apply the same rule here:

allow empty submissionText for SAVE/GIVE_UP

enforce non-blank submissionText only when action == SUBMIT

This will likely require:

removing the always-required constraint from the request DTO, and

adding SUBMIT-only validation in the service (similar to UserService).

 Can we address this now, or should we incur technical debt?

itachallenge-challenge/src/main/java/com/itachallenge/submission/service/SubmissionServiceImpl.java
```diff
@@ -27,6 +37,74 @@ public Flux<SubmissionDto> getAllSubmissionsByUser(String userId) {
...
+                .switchIfEmpty(Mono.error(new BadRequestException("The 'languageId' parameter cannot be null.")));
+
+
+        return Mono.zip(userUuidMono, challengeUuidMono, languageUuidMono)
+                .flatMap(tuple -> {
+                    UUID userUuid = tuple.getT1();
+                    UUID challengeUuid = tuple.getT2();
+                    UUID languageUuid = tuple.getT3();
+
+                    SubmissionAction action = SubmissionAction.fromString(request.getAction());
```
ivilarop: Note: although this PR focuses on replicating the current behavior from the User micro, it's still important to review existing code critically and spot potential robustness issues.

SubmissionAction.fromString should handle null/blank explicitly (similar to SubmissionStatus.fromString) and avoid catching generic Exception.
Otherwise we risk NPEs or inconsistent 400 error handling.

This is outside the current scope, so we can track it as technical debt if we prefer not to address it now.

itachallenge-challenge/src/main/java/com/itachallenge/submission/service/SubmissionServiceImpl.java
```diff
@@ -27,6 +37,74 @@ public Flux<SubmissionDto> getAllSubmissionsByUser(String userId) {
                 );
     }
 
+    @Override
+    public Mono<SubmissionResponseDto> createOrUpdateSubmission(String userId, SubmissionRequestDto request) {
```
ivilarop: For that service method, `createOrUpdateSubmission` doesn't accurately describe the intent, because it's not a CRUD upsert: it's processing an action (SAVE / SUBMIT / GIVE_UP) with rules (409 block, increment resolved, etc.).

In business, it's more common to name it by use case:

✅ My recommendation:

`handleSubmissionAction` or `processSubmissionAction`

itachallenge-challenge/src/main/java/com/itachallenge/submission/service/SubmissionService.java
```diff
@@ -1,10 +1,16 @@
...
+import com.itachallenge.challenge.dto.submission.SubmissionResponseDto;
 import com.itachallenge.submission.document.SubmissionDocument;
 import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
 
 public interface SubmissionService {
 
     Flux<SubmissionDto> getAllSubmissionsByUser(String userId);
+
+    Mono<SubmissionResponseDto> createOrUpdateSubmission(String userId, SubmissionRequestDto request);
```
ivilarop: For that service method, `createOrUpdateSubmission` doesn't accurately describe the intent, because it's not a CRUD upsert: it's processing an action (SAVE / SUBMIT / GIVE_UP) with rules (409 block, increment resolved, etc.).

In business, it's more common to name it by use case:

✅ My recommendation:

`handleSubmissionAction` or `processSubmissionAction`

ivilarop (changes requested):

itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/submission/SubmissionResponseDto.java
anaberod: I renamed SubmissionResponseDto to SubmissionActionResponseDto to better reflect its purpose as a POST action response.
The rename was applied in the service PR and propagated to the controller, Swagger and tests.

itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/submission/SubmissionRequestDto.java
anaberod: Got it, thanks for the clarification.
Using an internal command in the service layer and keeping HTTP DTOs in the controller PR makes sense. I’ll apply this approach in future PRs.

itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/submission/SubmissionResponseDto.java
ivilarop: Thanks for applying the DTO renaming — SubmissionActionResponseDto looks good and better reflects that this is an action result rather than a resource.

Regarding point 2 (DTOs belonging to the API layer only), please reply confirming you understand the reasoning behind keeping HTTP DTOs separate from business logic. Also, remember not to resolve review conversations yourself — reply first and let the reviewer close them.

itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/submission/SubmissionActionRequestDto.java
ivilarop: OK, @ValidUUID doesn't appear

itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/submission/SubmissionActionRequestDto.java
ivilarop: I see that the `@NotBlank` and `@NotNull` attributes have been correctly removed from the `DO` clause, and the following validation has been added to the service:

- Allowing blank text when the action does NOT require text (SHOW_SOLUTION / SAVE)
- Requiring text when the action DOES require it (SUBMIT )

itachallenge-challenge/src/main/java/com/itachallenge/submission/service/SubmissionServiceImpl.java
ivilarop: I see that the `@NotBlank` and `@NotNull` attributes have been correctly removed from the `DO` clause, and the following validation has been added to the service:

Allowing blank text when the action does NOT require text (SHOW_SOLUTION / SAVE)

Requiring text when the action DOES require it (SUBMIT)

itachallenge-challenge/src/main/java/com/itachallenge/submission/service/SubmissionServiceImpl.java
ivilarop: resolved

ivilarop (approved):

itachallenge-challenge/src/main/java/com/itachallenge/challenge/dto/submission/SubmissionRequestDto.java
ivilarop: done

itachallenge-challenge/src/main/java/com/itachallenge/submission/service/SubmissionServiceImpl.java
ivilarop: done

itachallenge-challenge/src/main/java/com/itachallenge/submission/service/SubmissionService.java
ivilarop: done

itachallenge-challenge/src/main/java/com/itachallenge/submission/service/SubmissionServiceImpl.java
ivilarop: Change made, request action is of type enum and no longer applies fromstring

PR #1073, opened 2026-01-23
Refactor(User/Favorites): Secure DELETE /users/{id}/favorites contract & HTTP Alignment (#41)

AlvaroLMC:
**Description**
This PR completes the refactoring for the favorite removal functionality. This update aligns the DELETE method with resource-based URI standards, enhances security through identity cross-checking, and maintains backward compatibility.

**Architectural Justification (DELETE)**
The deletion logic has been migrated from a flat route (DELETE /{challengeId}) to a structured sub-resource path: DELETE /users/{userId}/favorites/{challengeId}.
- Resource Hierarchy: Deletions are now explicitly performed on the favorite collection of a specific user, matching the underlying domain model.
- Enhanced Security (IDOR Protection): A cross-validation check has been implemented. The system now verifies that the userId provided in the URL matches the UUID extracted from the JWT token. This prevents users from deleting favorites belonging to other accounts.
- Redundancy Elimination: The microservice prefix has been streamlined to /itachallenge/api/v1/challenge, removing repetitive path segments like "favorite/challenges".

**Key Changes**
1. Technical Implementation
- New Endpoint: Created removeFavorite under the new hierarchical path.
- Identity Validation: Added ID comparison logic and BadRequestException handling for mismatches.
- Deprecation Management: The original method is now marked as @Deprecated, returning Deprecation and Link headers to ensure a smooth transition for existing consumers.
2. HTTP Consistency
- HTTP 200 OK: Maintained the 200 status for the deletion since the service returns the FavoriteDto with the updated state, following REST semantics for responses with a body.
3. Quality & Test Coverage
Unit tests have been refactored and expanded using StepVerifier to cover three critical scenarios:
- Success (200): Successful deletion with matching IDs.
- Unauthorized (400): Error when no valid token is provided or the header is missing.
- ID Mismatch (400): (New) Specifically validates that a user cannot delete another user's favorites, ensuring data integrity.

**Commit History**
- refactor (delete favorite): secure new API contract and handle deprecation (Includes: URI hierarchy cleanup, HTTP code alignment, and security logic).
- Test (Challenge): Refactor the favorite tests due to changes in FavoriteController (SonarQube visibility fixes and method signature updates).
- Test: ID Mismatch validation (Specific test case for the security cross-validation).

**Validation Checklist**
[x] Delete route follows the /users/{userId}/favorites/{challengeId} pattern.
[x] IDOR protection logic is fully implemented and tested.
[x] Legacy endpoint correctly returns deprecation headers.
[x] All JUnit 5 tests follow SonarQube's visibility best practices.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,18 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### [itachallenge-user-3.1.1] - 2026-01-27
+
+### Changed
+- Refactored favorite challenge removal logic to follow the hierarchical model: DELETE /users/{userId}/favorites/{challengeId}.
+- Deprecated legacy delete endpoint DELETE /{challengeId} in favor of the new user-centric structured path.
+- Updated FavoriteControllerTest to include validation for reactive deletion flows.
+
+### Added
+- Implemented RFC 8594 compliant Deprecation and Link headers in the legacy delete endpoint to guide consumer migration.
+- Added security validation in the deletion flow to ensure users can only remove challenges from their own favorites list.
+- Increased unit test coverage for the delete block, ensuring correct construction of ResponseEntity and its metadata.
+
 ### [itachallenge-challenge-3.1.0-RELEASE] - 2026-01-14
 
 ### Changed
--- a/conf/.env.CI.dev
+++ b/conf/.env.CI.dev
@@ -5,7 +5,7 @@ MICROSERVICE_DEPLOY_itachallenge-user=itachallenge-user
 MICROSERVICE_VERSION_itachallenge-user=3.1.3-RELEASE
 
 MICROSERVICE_DEPLOY_itachallenge-challenge=itachallenge-challenge
-MICROSERVICE_VERSION_itachallenge-challenge=3.1.0-RELEASE
+MICROSERVICE_VERSION_itachallenge-challenge=3.1.1-RELEASE
 
 MICROSERVICE_DEPLOY_itachallenge-document=itachallenge-document
 MICROSERVICE_VERSION_itachallenge-document=1.0.2-RELEASE
--- a/itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/FavoriteController.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/FavoriteController.java
@@ -51,7 +51,6 @@ public class FavoriteController {
                 .map(ResponseEntity::ok);
     }
 
-    // NEW REST CONTRACT
     @DeleteMapping("/users/{userId}/favorites/{challengeId}")
     @Operation(
             operationId = "Remove a challenge from User's favorites (new REST contract).",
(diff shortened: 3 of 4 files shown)


aperez1haad (commented):
H Alvaro. Pls, check the correct form of PR title. (https://tree.taiga.io/project/luisvi-ita-challenges/wiki/pr-review-checklist). I think that your pr is a refactor. 
Then the SonarQube is not ready yet. You don't have enough tests.

DiegoBalaguer (changes requested):
Here too you would have to use the new generated version instead of the 2.0.4 in Deprecated.

AlvaroLMC:
This commit synchronizes the documentation, environment configuration, and source code for the favorite removal feature:
- Documentation: Updated CHANGELOG.md to detail the migration from DELETE /{challengeId} to the user-centric DELETE /users/{userId}/favorites/{challengeId}.
- Version Management: Bumped version to 3.1.1-RELEASE in .env.CI.dev to align with the current development milestone.
- Controller Refinement: Adjusted the @Deprecated metadata in FavoriteController to ensure consistency with the release roadmap.
- Quality Assurance: Implemented a targeted test in FavoriteControllerTest to verify that the legacy delete endpoint correctly injects Deprecation and Link headers, ensuring 100% coverage of the new reactive mapping block.

DiegoBalaguer (approved):
You've made the changes mentioned and explained why you made them. Good job.

aperez1haad (approved):
Good job Alvaro. However, be careful with comments on the code. If isn't totally necessary pls delete comments before send the pr to the next step.

AlvaroLMC:
@aperez1haad I removed the comments as you suggested.

ivilarop (changes requested):
@AlvaroLMC @anaberod @DiegoBalaguer

The objective of this task (as described in the User Handbook) was to regularize/refactor the FavoriteController paths within the user microservice. In other words, the focus of the change is the API of the user microservice, not the public API of the challenge microservice.

1.- Impact on the Frontend (Favorites GET)

It was explicitly stated that the Frontend performs the Favorites GET by calling the user microservice directly (it doesn't go through the challenge microservice).

Therefore:

- If the FavoriteController path in the user microservice is changed, the Favorites GET in the Frontend will break.
- This means that, as part of the migration, it is necessary to:
    - temporarily maintain the legacy endpoint (deprecated), or
    - update the Frontend to point to the new path (according to the agreed-upon strategy).

In summary: there is an impact on the Frontend, because the Frontend directly consumes that endpoint of the user microservice.

2.- Impact on Challenge (POST/DELETE favorites)

However, for adding or deleting favorites (POST/DELETE), the current logic is different:

- The Frontend does not call the user microcontroller for these operations.
- The Frontend calls the challenge microcontroller, and it is the challenge microcontroller that then communicates internally with the user microcontroller.

Therefore, when the path of the FavoriteController in the user microcontroller is regularized:

- The impact on challenge should NOT be on its controllers (public API).
- The only place where challenge should be changed is the internal service/client that makes the HTTP call to the user microcontroller.

As we've discussed several times, my role as a senior developer is to detail the objective, context, and impacts, but it's the developer's responsibility to analyze the full scope and take ownership of it.

In microservices, it's key to:

- understand who owns the change
- identify consumers
- limit the scope
- and if in doubt, ask before expanding the impact

The user guide clearly defined that the refactor belonged to the micro-user.
If there was any doubt about also addressing the challenge, it had to be validated beforehand.
This type of impact analysis is part of the developer's job, not just the implementation.

PR #1076, opened 2026-01-27
Refactor(User/Favorites): New REST contract Get User Favorites (#38)

AlvaroLMC:
**Architectural Justification**

The endpoint to retrieve a user's favorite challenges has been moved from a top-level resource to a user sub-resource: GET /users/{userId}/favorites.
- Resource Hierarchy: Favorites are now correctly modeled as a property of a user. This follows the REST maturity model where the URI represents the relationship between resources.
- Path Uniformity: By removing the redundant /userinteraction/favorites segments and nesting the resource under /users, we achieve a cleaner and more predictable API structure.

**Key Changes**
1. Implementation & Migration
- New Hierarchical Route: Created the new endpoint under the /users/{userId}/favorites path.
- Legacy Support: The previous endpoint GET /favorites/{userId} is now marked as @Deprecated. It remains functional but includes technical signals for consumers to migrate.
- Deprecation Headers: Implemented Deprecation: true and a Link header pointing to the new URI, following RFC 8594 standards for API evolution.
2. Quality & Standards
- SonarQube Compliance: Added structured Javadoc with the @deprecated tag to ensure clear documentation and pass static analysis gates (Rule RSPEC-1123).
- HTTP Semantics: Maintained 200 OK for successful retrievals, with proper error handling for 400 (Bad Request) on malformed IDs and 404 (Not Found) if the user profile does not exist.

**Related Commits**
- refactor (get favorites): secure new API contract and handle deprecation for user favorites
- test (get favorites): update test suite for new URI

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,18 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### [itachallenge-user-3.1.4-RELEASE] - 2026-01-27
+
+### Changed
+- Refactored favorite challenges retrieval logic to follow the hierarchical model: GET /users/{userId}/favorites.
+- Deprecated legacy endpoint GET /favorites/{userId} in favor of the new user-centric structured path.
+- Updated FavoriteControllerTest to include validation for reactive retrieval flows and pagination (if applicable).
+
+### Added
+- Implemented RFC 8594 compliant Deprecation and Link headers in the legacy GET endpoint to guide consumer migration to the new resource URI.
+- Added security validation to the retrieval flow to ensure users can only access their own favorites list.
+- Increased unit test coverage for the GET block, verifying the correct mapping of the favorite challenges list and response metadata.
+
 ### [itachallenge-challenge-3.1.0-RELEASE] - 2026-01-14
 
 ### Changed
--- a/conf/.env.CI.dev
+++ b/conf/.env.CI.dev
@@ -2,7 +2,7 @@ MICROSERVICE_DEPLOY_itachallenge-auth=itachallenge-auth
 MICROSERVICE_VERSION_itachallenge-auth=2.0.5-RELEASE
 
 MICROSERVICE_DEPLOY_itachallenge-user=itachallenge-user
-MICROSERVICE_VERSION_itachallenge-user=3.1.3-RELEASE
+MICROSERVICE_VERSION_itachallenge-user=3.1.4-RELEASE
 
 MICROSERVICE_DEPLOY_itachallenge-challenge=itachallenge-challenge
 MICROSERVICE_VERSION_itachallenge-challenge=3.1.0-RELEASE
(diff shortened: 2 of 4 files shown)


AlvaroLMC:
**Changes summary by file:**
- CHANGELOG.md: Documented the migration path from GET /favorites/{userId} to the new user-centric resource /users/{userId}/favorites, including security and header improvements.
- conf/.env.CI.dev: Synchronized CI/CD environment variables with the new versioning (3.1.4-RELEASE) and specific microservice deployment keys.
- FavoriteController.java: Fixed a versioning mismatch in the @Deprecated annotation to ensure accurate API documentation and alignment with the release roadmap.
- FavoriteControllerTest.java: Verified the implementation of Deprecation and Link headers for the GET flow, ensuring the reactive pipeline is fully covered.

DiegoBalaguer (approved):
Good job Alvaro

aperez1haad (changes requested):
Hi Alvaro, pls check the comments on the code. And when your Pr have 3 steps, pls explain on the description the enviroment (the global task and subtasks) and the Pr title should be PR (1/3, 2/4, for example)

AlvaroLMC:
@aperez1haad I removed the comments as you suggested. Regarding the route nomenclature (x/x), I didn't use it because they are separate tasks.

aperez1haad (approved):
ok! Good job Alvaro!

ivilarop (changes requested):
@AlvaroLMC @anaberod @DiegoBalaguer

The objective of this task (as described in the User Handbook) was to regularize/refactor the FavoriteController paths within the user microservice. In other words, the focus of the change is the API of the user microservice, not the public API of the challenge microservice.

1.- Impact on the Frontend (Favorites GET)

It was explicitly stated that the Frontend performs the Favorites GET by calling the user microservice directly (it doesn't go through the challenge microservice).

Therefore:

- If the FavoriteController path in the user microservice is changed, the Favorites GET in the Frontend will break.
- This means that, as part of the migration, it is necessary to:
    - temporarily maintain the legacy endpoint (deprecated), or
    - update the Frontend to point to the new path (according to the agreed-upon strategy).

In summary: there is an impact on the Frontend, because the Frontend directly consumes that endpoint of the user microservice.

2.- Impact on Challenge (POST/DELETE favorites)

However, for adding or deleting favorites (POST/DELETE), the current logic is different:

- The Frontend does not call the user microcontroller for these operations.
- The Frontend calls the challenge microcontroller, and it is the challenge microcontroller that then communicates internally with the user microcontroller.

Therefore, when the path of the FavoriteController in the user microcontroller is regularized:

- The impact on challenge should NOT be on its controllers (public API).
- The only place where challenge should be changed is the internal service/client that makes the HTTP call to the user microcontroller.

As we've discussed several times, my role as a senior developer is to detail the objective, context, and impacts, but it's the developer's responsibility to analyze the full scope and take ownership of it.

In microservices, it's key to:

- understand who owns the change
- identify consumers
- limit the scope
- and if in doubt, ask before expanding the impact

The user guide clearly defined that the refactor belonged to the micro-user.
If there was any doubt about also addressing the challenge, it had to be validated beforehand.
This type of impact analysis is part of the developer's job, not just the implementation.

PR #1082, opened 2026-02-04
Chore(user): #111 remove deprecated bookmarks legacy endpoint

adolcc:
Description
This PR focuses on removing technical debt by deleting the deprecated legacy endpoint for user bookmarks. The domain logic has been successfully migrated to the new subresource-based structure.

Changes Made
Deleted: BookmarkLegacyController.java and its associated tests to clean up the codebase.

Updated: application.yml in the challenge microservice to point to the new correct endpoint path: /itachallenge/api/v1/users/{userId}/bookmarks.

Refactored: BookmarkControllerTest.java to ensure all test cases (including error scenarios) target the new resource path.

Verification Results
Unit Tests: All tests in BookmarkControllerTest are passing with the new URI.

Configuration: Verified that the challenge service configuration no longer uses the redundant /user/ segment or the unnecessary {challengeId} path variable.

Impact
Prevents 404 errors in future integrations between the User and Challenge microservices.


adolcc:
@ivilarop the test don't past but i think is relative to the server.
Captura de pantalla 2026-02-04 a las 22 27 11

adolcc:
@ivilarop restore accidental deletion of version 3.2.0 in CHANGELOG and document legacy endpoint removal.

DiegoBalaguer (approved):
Hi Adolfo, good job finally removing the `BookmarkLegacyController` and updating the `application.yml` for the challenges microservice. The decoupling seems complete now that the routes have been corrected.

I’ve only noticed one detail in the **CHANGELOG.md**:

I see that the new entry has no assigned version or an `[Unreleased]` header. Even though this is a technical debt refactor, it is necessary to follow the project standard to ensure the record is traceable:

* **Suggestion:** If we don't have a release version assigned yet, you could move these changes under the `## [Unreleased]` header.

Other than that, the test refactoring to point to the new `/bookmarks` URI is correct and necessary to avoid false positives. Good job closing this cycle!

carlosPc1987 (commented):
**Good job removing the legacy endpoint and updating the routes.**

Just one important detail regarding the `CHANGELOG.md`:

The new entry cannot be placed above `itachallenge-challenge-3.1.0-RELEASE`, since that version is already closed.  
According to the project standard, all pending changes must go under:

## [Unreleased]

This ensures proper traceability and prevents mixing new changes with already published versions.

---

### 🔎 Issue 1 — You placed the change inside a version that has already been released

The block:
 [itachallenge-challenge-3.1.0-RELEASE] - 2026-01-14

is a closed release, with a date and finalized content.

➡️ **New changes cannot be added inside or above a closed release.**

---

### 🔎 Issue 2 — You didn’t use the `## [Unreleased]` header

The project standard (and *Keep a Changelog*) states:

- New changes must always go under `## [Unreleased]`
- When a new release is created, that block is moved to the corresponding version

By adding changes without a header, they end up “floating” and lose traceability.

---

Please move the new entry under the `## [Unreleased]` section to align with the project conventions.

adolcc:
@DiegoBalaguer @carlosPc1987 Hi team, I don't think we should update the changelog here. Since this is a refactoring, the documentation specifies that it's not necessary in this case. I'm basing my comment on that; please take another look and let me know your thoughts. So I can procedure to remove the changelog version from the pr.  Thanks for the review!
Captura de pantalla 2026-02-11 a las 9 33 05

ivilarop:
We're closing this PR because it's incomplete and due to lack of time we'll address it in the next sprint

PR #1092, opened 2026-02-06
feat(challenge): Secure Delete Challenge Endpoint with JWT validation #150

pantalois:
## 📌 Summary ##
Secures the DELETE challenge endpoint in the Challenge microservice by reusing the existing JWT validation approach already used in this controller (bookmarks/update).

The endpoint now requires a valid Authorization: Bearer <token> and returns 401 Unauthorized when the header is missing or invalid, without introducing Spring Security or new auth mechanisms.

## 🎯 Scope ##
MVP (required):

Protect DELETE /itachallenge/api/v1/challenge/challenges/{challengeId}

Manually read the Authorization header (same pattern as bookmarks/update)

Validate JWT using ChallengeJwtFacade + jwt-core

Extract UUID from JWT (via the facade)

Return 401 if Authorization header is missing/invalid

## 🧱 Architecture ##

No Spring Security Resource Server introduced.

No new security layers or mechanisms added.

To enforce the MVP behavior (401 when Authorization header is missing/invalid), the controller maps JWT validation errors to an HTTP status using Spring Web utilities:

- org.springframework.http.HttpStatus
- org.springframework.web.server.ResponseStatusException

Implementation stays at controller level and matches the pattern already present:

- Read Authorization header
- Delegate parsing/validation to ChallengeJwtFacade (backed by jwt-core)

JWT validation uses the JJWT exception type (io.jsonwebtoken.JwtException). To avoid conflicts with similarly named/custom exceptions, the code references the JJWT exception explicitly where needed.

## 🔒 Validations ##

A request is authorized only if:

Header exists and starts with Bearer

ChallengeJwtFacade.getUserUuIdFromAuthenticationHeader(...) successfully extracts a UUID

Missing/invalid header/token → 401 Unauthorized

## 🧪 Tests ##

Controller tests were updated to reflect the new DELETE JWT requirement and to avoid false positives:

### Updated deleteOneChallenge_notFound ###

This test now includes a valid Authorization: Bearer ... header and mocks the UUID extraction with ChallengeJwtFacade.
Reason: once DELETE is protected, a request without a token will return 401 before reaching the service, so to assert the real 404 Not Found path, the request must pass JWT validation first.
The test verifies:

JWT UUID extraction is successful (challengeJwtFacade.getUserUuIdFromAuthenticationHeader(authHeader) returns a userId)

The service is called once

The endpoint returns 404 with the expected error message when the challenge does not exist

### Added deleteOneChallenge_missingHeader_shouldReturn401 ###

New coverage for the MVP behavior: calling DELETE without the Authorization header must return 401 Unauthorized.
The test mocks the facade to throw io.jsonwebtoken.JwtException when the header is missing, and asserts:

Response status is 401

The service is never invoked (verify(..., never())) ensuring the request is blocked at the JWT validation step

## 🧹 Refactor suggestion (follow-up / tech debt) ##
It would be useful to refactor the controller to extract the repeated JWT validation logic into a small helper method (or shared utility), so we don’t duplicate the same pattern across endpoints (bookmarks/update/delete and others).

Example idea:

- Create a private method in the controller like requireUserId(authHeader) that:
- validates Authorization: Bearer ...
- delegates to ChallengeJwtFacade
- maps JWT errors to the expected HTTP status (401)
- Then each endpoint can simply call .flatMap(userId -> ...) and keep the business flow clean.

This keeps the sprint implementation unchanged, but improves readability, reduces duplication, and makes future changes to JWT handling safer and faster.

### DIFF
--- a/itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeController.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeController.java
@@ -3,6 +3,7 @@ package com.itachallenge.challenge.controller;
 import com.itachallenge.challenge.annotations.ValidGenericPattern;
 import com.itachallenge.challenge.config.PropertiesConfig;
 import com.itachallenge.challenge.dto.*;
+import com.itachallenge.challenge.exception.UnauthorizedException;
 import com.itachallenge.common.exception.BadRequestException;
 import com.itachallenge.challenge.exception.JwtException;
 import com.itachallenge.challenge.service.IChallengeService;
@@ -291,7 +292,7 @@ public class ChallengeController {
             @RequestHeader(name = "Authorization", required = false) String authHeader
     ) {
         return Mono.fromCallable(() -> challengeJwtFacade.getUserUuIdFromAuthenticationHeader(authHeader))
-                .onErrorMap(io.jsonwebtoken.JwtException.class, e -> new ResponseStatusException(HttpStatus.UNAUTHORIZED, "Missing or invalid authorization token", e))
+                .onErrorMap(e -> new UnauthorizedException("Missing or invalid authorization token"))
                 .flatMap(userId -> challengeService.deleteChallengeById(challengeId))
                 .map(ResponseEntity::ok)
                 .doOnError(error -> log.error("Error deleting challenge: {}", error.getMessage()));
--- a/itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeController.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeController.java
@@ -20,6 +20,8 @@ import org.springframework.cloud.client.discovery.DiscoveryClient;
 import org.springframework.http.ResponseEntity;
 import org.springframework.validation.annotation.Validated;
 import org.springframework.web.bind.annotation.*;
+import org.springframework.http.HttpStatus;
+import org.springframework.web.server.ResponseStatusException;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
@@ -285,10 +287,14 @@ public class ChallengeController {
             }
     )
     public Mono<ResponseEntity<DeleteResponseDto>> deleteChallenge(
-            @PathVariable String challengeId) {
-
-        return challengeService.deleteChallengeById(challengeId)
-                .map(ResponseEntity::ok);
+            @PathVariable String challengeId,
+            @RequestHeader(name = "Authorization", required = false) String authHeader
+    ) {
+        return Mono.fromCallable(() -> challengeJwtFacade.getUserUuIdFromAuthenticationHeader(authHeader))
+                .onErrorMap(io.jsonwebtoken.JwtException.class, e -> new ResponseStatusException(HttpStatus.UNAUTHORIZED, "Missing or invalid authorization token", e))
+                .flatMap(userId -> challengeService.deleteChallengeById(challengeId))
+                .map(ResponseEntity::ok)
+                .doOnError(error -> log.error("Error deleting challenge: {}", error.getMessage()));
     }
 
     @PostMapping("/challenges/{challengeId}/bookmarks")
--- a/itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeController.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeController.java
@@ -21,8 +21,6 @@ import org.springframework.cloud.client.discovery.DiscoveryClient;
 import org.springframework.http.ResponseEntity;
 import org.springframework.validation.annotation.Validated;
 import org.springframework.web.bind.annotation.*;
-import org.springframework.http.HttpStatus;
-import org.springframework.web.server.ResponseStatusException;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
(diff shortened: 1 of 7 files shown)


DiegoBalaguer (approved):
Good job Lois.

Lucy-SD (approved):
Well done. Looks good and clean.

itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeController.java
```diff
@@ -285,10 +287,14 @@ public Mono<ResponseEntity<Map<String, String>>> getVersion() {
...
     public Mono<ResponseEntity<DeleteResponseDto>> deleteChallenge(
-            @PathVariable String challengeId) {
-
-        return challengeService.deleteChallengeById(challengeId)
-                .map(ResponseEntity::ok);
+            @PathVariable String challengeId,
+            @RequestHeader(name = "Authorization", required = false) String authHeader
+    ) {
+        return Mono.fromCallable(() -> challengeJwtFacade.getUserUuIdFromAuthenticationHeader(authHeader))
+                .onErrorMap(io.jsonwebtoken.JwtException.class, e -> new ResponseStatusException(HttpStatus.UNAUTHORIZED, "Missing or invalid authorization token", e))
```
ivilarop: io.jsonwebtoken.JwtException.class: Import and use the simple name for readability.

itachallenge-challenge/src/test/java/com/itachallenge/challenge/controller/ChallengeControllerTest.java
```diff
@@ -582,16 +582,39 @@ void deleteOneChallenge_success() {
...
+
+        verify(challengeService, times(1)).deleteChallengeById(id);
+    }
+
+    @Test
+    void deleteOneChallenge_missingHeader_shouldReturn401() {
+        String id = "non_existing_id";
+
+        when(challengeJwtFacade.getUserUuIdFromAuthenticationHeader(null))
+                .thenThrow(new io.jsonwebtoken.JwtException("Missing or bad formatted Authorization header"));
```
ivilarop: io.jsonwebtoken.JwtException.class: Import and use the simple name for readability.

ivilarop (commented):
- Add screenshots of the tests performed.
- Missing changelog

itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/ChallengeController.java
pantalois: Hi Iván,

I used io.jsonwebtoken.JwtException.class directly inside the method instead of importing it, because there are some methods that use a custom exception also named JwtException. If I add the import, it creates a naming conflict and I can’t reference the Spring/JJWT one properly unless I refactor the custom exception and rename it. I was planning to add that as technical debt.

Is it okay to keep it as it is for now (using io.jsonwebtoken.JwtException.class in the method scope), or would you prefer I refactor the custom exception for compatibility?

pantalois: @ivilarop 

Hi Iván,

I used io.jsonwebtoken.JwtException.class directly inside the method instead of importing it, because there are some methods that use a custom exception also named JwtException. If I add the import, it creates a naming conflict and I can’t reference the Spring/JJWT one properly unless I refactor the custom exception and rename it. I was planning to add that as technical debt.

Is it okay to keep it as it is for now (using io.jsonwebtoken.JwtException.class in the method scope), or would you prefer I refactor the custom exception for compatibility?

pantalois:
## 📌 Summary ##

Secures the DELETE challenge endpoint by returning 401 Unauthorized when the Authorization header/token is missing, invalid, or expired, using a new custom UnauthorizedException handled by the GlobalExceptionHandler.
No Spring Security added. No new auth mechanism. Same “manual validation” approach as the controller’s existing protected endpoints.

## 🎯 Scope ##

MVP (required):

- Protect DELETE /itachallenge/api/v1/challenge/challenges/{challengeId}
- Manually read Authorization header (same controller-level pattern)
- Validate/extract UUID via ChallengeJwtFacade (backed by jwt-core)
- Map JWT validation failures to UnauthorizedException
- Global handler returns 401 Unauthorized

## 🧱 Architecture ##

- No Spring Security Resource Server introduced.
- No new filters/security layers added.
- The controller enforces the 401 behavior by converting token-validation failures to the app-level exception:
- UnauthorizedException (custom, in Challenge)
- handled in GlobalExceptionHandler → 401 with MessageDto
- This avoids relying on ResponseStatusException and avoids coupling the controller to io.jsonwebtoken.JwtException.

## 🔒 Validations ##

- A request is authorized only if:
1. Header exists and starts with Bearer
2. ChallengeJwtFacade.getUserUuIdFromAuthenticationHeader(...) successfully extracts a UUID
3. Missing/invalid/expired token → UnauthorizedException → 401 Unauthorized

## 🧪 Tests ##

Controller tests were updated to reflect the new DELETE JWT requirement and avoid false positives:

- Updated deleteOneChallenge_notFound
- Now includes a valid Authorization: Bearer ... and mocks UUID extraction so the request passes JWT validation and reaches the service.
Verifies:

- UUID extraction succeeds (challengeJwtFacade.getUserUuIdFromAuthenticationHeader(...) returns a userId)
- Service called once
- Endpoint returns 404 when the challenge does not exist
- Added deleteOneChallenge_missingHeader_shouldReturn401
- Covers MVP behavior: calling DELETE without the header returns 401 and the service is not invoked.

Verifies:

- Status is 401
- Service is never called (verify(..., never()))

## Invalid token ## 
image


## Valid token ## 

image

## Tests for the new behavior ##
image

PR #1105, opened 2026-02-17
fix(ci): update Testcontainers via BOM (Docker API compat)

ivilarop:


ivilarop:
We created a new PR with an MVP: disable integration tests that depend on testContainer

PR #1107, opened 2026-02-19
[BE] PR 1/3 (Shared): Infrastructure for Gamification Persistence (#215)

Lucy-SD:
## Chore: Infrastructure for Gamification Persistence

**closes sub-issue #https://github.com/orgs/IT-Academy-BCN/projects/26/views/2?pane=issue&itemId=158023376&issue=IT-Academy-BCN%7Cita-challenges%7C215**

###  📝 Description
This PR establishes the foundational data layer for the gamification module within the Challenge microservice. It migrates the "source of truth" for user points from the inconsistent `User` database to a dedicated, transaction-based collection in the `Challenge` database.

### ✅ Acceptance Criteria
* New MongoDB collection `user_score_history` is defined.
* The system can persist individual point transactions.
* Repository supports chronological queries by `userId`.

### 🔧 Technical Details
* **Document Structure:** `userId` (UUID), `points` (Integer), `challengeId` (UUID), `createdAt` (Date).
* **Package:** `com.itachallenge.challenge.document` and `repository`.
* **Indexes:** Compound index on `{userId: 1, createdAt: 1}` for query optimization.
* **Sum:** The total points will be calculated as the sum of the points in the history by userId.
* **Important Consideration:** The userId will not be received as an editable parameter in the future endpoint's JSON, but will instead be extracted from the Security Context (JWT). This ensures that one user cannot query or "inject" points into another user's history.

### 📂 New Directory Structure
To maintain consistency with the submission module, the Gamification architecture is organized by separating the domain (root) logic from the microservice entry points (challenge):

```
src/main/java/com/itachallenge
├── gamification/          
│   ├── document/
│   │   └── UserScoreDocument.java
│   └── repository/
│       └── UserScoreRepository.java
└── challenge/                    
    ├── controller/
    │   └── gamification/
    │       └── RankingController.java
    │       └── UserScoreHisotoryController.java
    └── dto/
            └── gamification/
           ├── PointsHistoryDto.java
           └── PointEntryDto.java
```
            
### 🏛️ Architectural Breakdown

* Pattern: Event Sourcing (Light). We don't store a fixed state, but rather an event log (user_score_history). This allows us to reconstruct the history at any time without data loss.

* Decoupling: The com.itachallenge.gamification package functions as a standalone module. If Gamification becomes its own microservice tomorrow, migration simply involves moving that folder.

* Data Access: Queries: Optimized using the composite index (userId, createdAt).

[...]

PR #1108, opened 2026-02-19
[BE] Gamification: package structure, record score events and unit tests(#228)

carlosPc1987:
# 🎯 Objective
Implement the gamification package structure inside the Challenge module, the logic that records scoring events when a submission reaches `SUBMITTED_COMPLETE`, and the tests that validate this behavior.  
The POST submission contract is not modified (points are **not** exposed in the response in this task).

---

# 🔗 Dependency
Requires the merge of **[BE] PR 1/3 (Shared): Infrastructure for Gamification Persistence (Feature#19b #215)** into the base branch.  
This includes `UserScoreDocument` and `UserScoreRepository`.

---

# 📝 Changes Implemented

## New files
- `com.itachallenge.gamification.service.PointsService` – interface with `recordPoints(userId, challengeId, points)`.
- `com.itachallenge.gamification.service.impl.PointsServiceImpl` – persists a document into the `user_score_history` collection using `UserScoreRepository`.
- `com.itachallenge.gamification.service.impl.PointsServiceImplTest` – unit tests with a mocked `UserScoreRepository`; verifies that the saved document contains `userId`, `challengeId`, `points`, and `createdAt`.

## Modified files
- `SubmissionServiceImpl` – injects `PointsService`; when `saved.getStatus() == SUBMITTED_COMPLETE`, it calls  
  `pointsService.recordPoints(userUuid, challengeUuid, 10)`.  
  The POST response does **not** include a `points` field.
- `SubmissionServiceImplTest` – mocks `PointsService`; verifies that `recordPoints` is called in the `SUBMITTED_COMPLETE` flow and **not** called in other cases.

---

# 🧪 How to test

Run:    ./gradlew clean build
(or `gradlew.bat clean build` on Windows) from the project root.

All tests must pass, including `PointsServiceImplTest` and `SubmissionServiceImplTest`.

---

# ✅ Acceptance Criteria

- Gamification package (service + impl) created and used by `SubmissionServiceImpl`.
- When a submission result is `SUBMITTED_COMPLETE`, a gamification record is persisted.
- The POST submission response contract remains unchanged.
- Unit tests for `PointsService`/`PointsServiceImpl` are green.
- `SubmissionServiceImpl` tests cover:
  - scoring event triggered on `SUBMITTED_COMPLETE`
  - no call to `PointsService` in other statuses
- Code follows project conventions and compiles successfully.

Captura de pantalla (BUILD DE GRADDLE)2026-02-19 145456

[...]

PR #1109, opened 2026-02-20
[BE] PR 2/3: Point History Retrieval Logic (#216)

Lucy-SD:
## Feature: Implementation of points history and accumulation logic

**closes sub-issue: #https://github.com/orgs/IT-Academy-BCN/projects/26/views/2?pane=issue&itemId=158024051&issue=IT-Academy-BCN%7Cita-challenges%7C216**

### 📝 Description
Implementation of the core business logic to retrieve a student's point history. This service processes raw database entries into a structured format suitable for growth charts.

### ✅ Acceptance Criteria
* Logic calculates `totalPoints` as the sum of all historical entries.
* Entries are sorted in decreasing order by date.
* Dates are returned in ISO 8601 format.

### 🔧 Technical Details
* **DTOs:** `PointsHistoryResponseDto` (contains `totalPoints` and a list of `PointHistoryEntryDto`).
* **Service Logic:** `PointsServiceImp` consumes the `Flux<UserScoreDocument>` from the repository, using `collectList()` operator to process the data reactively. In a single transformation step, it calculates the arithmetic sum of `totalPoints` and maps the documents to the `PointHistoryEntryDto`, returning a `Mono<PointsHistoryResponseDto>`.
* **Exceptions:** Proper handling of `UserNotFound` using the `common.exception` package.
* **Mapping:** Creation of `GamificationMapper`to centralize the conversion logic between `UserScoreDocument` and `PointEntryDto`, ensuring consistency across the gamification module.

### 💰 Value Provided
Transforms raw data into "Progress Information," allowing students to visualize their growth over time.
Promotes code reusability by providing a shared mapper for other stories (Ranking/Submission) within the same challenge microservice.

### 🔓Unlocks
Unblocks the final API Controller: sub-issue [#217](https://github.com/orgs/IT-Academy-BCN/projects/26/views/2?pane=issue&itemId=158025144&issue=IT-Academy-BCN%7Cita-challenges%7C217); and provides the data structure for Frontend graph components.

### 🔗 Dependencies
Chained to #215 - (PR #1107)
Closes #https://github.com/orgs/IT-Academy-BCN/projects/26/views/2?pane=issue&itemId=158024051&issue=IT-Academy-BCN%7Cita-challenges%7C216

### 🧪 Test Implementations
* **Unit Tests:** Mocking the Repository to verify the sum calculation and chronological sorting logic in the Service.

### 🔄 Acceptance Criteria (DoD)
* No hardcoded data in the Service.
* Code coverage for logic > 80%.
* PR description links to the previous shared PR.

### 👨‍💻 Technical Debts
* `totalPoints` is calculated on-the-fly. For high-volume users, an "Aggregated Totals" table might be needed in the future.
* Username Enrichment: Currently, the username is retrieved from the history. If the history is empty, the field returns an empty string. It is recommended to extend IChallengeJwtFacade to extract the username directly from the JWT in future iterations.

### DIFF
--- a/itachallenge-challenge/src/main/java/com/itachallenge/gamification/mapper/GamificationMapper.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/gamification/mapper/GamificationMapper.java
@@ -1,0 +1,18 @@
+package com.itachallenge.gamification.mapper;
+
+import com.itachallenge.challenge.dto.gamification.PointEntryDto;
+import com.itachallenge.gamification.document.UserScoreDocument;
+import org.springframework.stereotype.Component;
+
+@Component
+public class GamificationMapper {
+
+    public PointEntryDto toPointEntryDto(UserScoreDocument document) {
+        if (document == null) return null;
+
+        return PointEntryDto.builder()
+                .createdAt(document.getCreatedAt() != null ? document.getCreatedAt().toString() : "")
+                .points(document.getPoints() != null ? document.getPoints() : 0)
+                .build();
+    }
+}
--- a/itachallenge-challenge/src/main/java/com/itachallenge/gamification/mapper/GamificationMapper.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/gamification/mapper/GamificationMapper.java
@@ -1,16 +1,16 @@
 package com.itachallenge.gamification.mapper;
 
-import com.itachallenge.challenge.dto.gamification.PointEntryDto;
+import com.itachallenge.challenge.dto.gamification.PointHistoryEntryDto;
 import com.itachallenge.gamification.document.UserScoreDocument;
 import org.springframework.stereotype.Component;
 
 @Component
 public class GamificationMapper {
 
-    public PointEntryDto toPointEntryDto(UserScoreDocument document) {
+    public PointHistoryEntryDto toPointEntryDto(UserScoreDocument document) {
         if (document == null) return null;
 
-        return PointEntryDto.builder()
+        return PointHistoryEntryDto.builder()
                 .createdAt(document.getCreatedAt() != null ? document.getCreatedAt().toString() : "")
                 .points(document.getPoints() != null ? document.getPoints() : 0)
                 .build();
--- a/itachallenge-challenge/src/main/java/com/itachallenge/gamification/service/PointsService.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/gamification/service/PointsService.java
@@ -7,6 +7,6 @@ import java.util.UUID;
 
 public interface PointsService {
 
-    Mono<PointsHistoryDto> getUserPointsHistory(UUID userId);
+    Mono<PointsHistoryDto> getUserPointsHistory(String username);
 }
 
--- a/itachallenge-challenge/src/main/java/com/itachallenge/gamification/service/PointsService.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/gamification/service/PointsService.java
@@ -1,12 +1,12 @@
 package com.itachallenge.gamification.service;
 
-import com.itachallenge.challenge.dto.gamification.PointsHistoryDto;
+import com.itachallenge.challenge.dto.gamification.PointsHistoryResponseDto;
 import reactor.core.publisher.Mono;
 
 import java.util.UUID;
 
 public interface PointsService {
 
-    Mono<PointsHistoryDto> getUserPointsHistory(UUID userId);
+    Mono<PointsHistoryResponseDto> getUserPointsHistory(UUID userId);
 }
 
--- a/itachallenge-challenge/src/main/java/com/itachallenge/gamification/service/PointsService.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/gamification/service/PointsService.java
@@ -7,6 +7,6 @@ import java.util.UUID;
 
 public interface PointsService {
 
-    Mono<PointsHistoryDto> getUserPointsHistory(String username);
+    Mono<PointsHistoryDto> getUserPointsHistory(UUID userId);
 }
 
--- a/itachallenge-challenge/src/main/java/com/itachallenge/gamification/service/PointsService.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/gamification/service/PointsService.java
@@ -7,6 +7,6 @@ import java.util.UUID;
 
 public interface PointsService {
 
-  Mono<PointsHistoryDto> getUserPointsHistory(UUID uuid);
+  Mono<PointsHistoryDto> getUserPointsHistory(UUID userId);
 }
 
--- a/itachallenge-challenge/src/main/java/com/itachallenge/gamification/service/PointsService.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/gamification/service/PointsService.java
@@ -7,6 +7,6 @@ import java.util.UUID;
 
 public interface PointsService {
 
-  Mono<PointsHistoryDto> getUserPointsHistory(UUID userId);
+    Mono<PointsHistoryDto> getUserPointsHistory(UUID userId);
 }
 
--- a/itachallenge-challenge/src/main/java/com/itachallenge/gamification/service/PointsService.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/gamification/service/PointsService.java
@@ -1,0 +1,12 @@
+package com.itachallenge.gamification.service;
+
+import com.itachallenge.challenge.dto.gamification.PointsHistoryDto;
+import reactor.core.publisher.Mono;
+
+import java.util.UUID;
+
+public interface PointsService {
+
+  Mono<PointsHistoryDto> getUserPointsHistory(UUID uuid);
+}
+
--- a/itachallenge-challenge/src/main/java/com/itachallenge/gamification/service/PointsServiceImpl.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/gamification/service/PointsServiceImpl.java
@@ -23,7 +23,7 @@ public class PointsServiceImpl implements PointsService {
 
     @Override
     public Mono<PointsHistoryResponseDto> getUserPointsHistory(UUID userId) {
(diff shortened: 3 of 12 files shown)


Lucy-SD:
## All Tests Passed

image

pantalois:
@Lucy-SD 

First, regarding the tests, I see that you have only added tests for the service implementation. I think it would also be helpful to add some small tests for the DTOs and the mapper, as I noticed those have not been included.

Additionally, I would suggest renaming the DTOs. Based on the project architecture, it seems we are following the RequestDTO and ResponseDTO naming convention. It might be clearer if the DTO names explicitly reflect whether they are request or response objects.

Apart from that, I downloaded your code, created a small controller, and set up the collection in my local database to verify that the logic works correctly beyond the existing tests. Everything is functioning as expected.

So once you make those adjustments, I believe we can consider the issue ready, pending confirmation of the first linked one.

Let´s go!!!

Lucy-SD:
> @Lucy-SD
> 
> First, regarding the tests, I see that you have only added tests for the service implementation. I think it would also be helpful to add some small tests for the DTOs and the mapper, as I noticed those have not been included.
> 
> Additionally, I would suggest renaming the DTOs. Based on the project architecture, it seems we are following the RequestDTO and ResponseDTO naming convention. It might be clearer if the DTO names explicitly reflect whether they are request or response objects.
> 
> Apart from that, I downloaded your code, created a small controller, and set up the collection in my local database to verify that the logic works correctly beyond the existing tests. Everything is functioning as expected.
> 
> So once you make those adjustments, I believe we can consider the issue ready, pending confirmation of the first linked one.
> 
> Let´s go!!!


Hi Lois, I added some mapper and DTO tests. I also added a filter to the `getUserPointsHistory` method so that duplicate challengeIds are ignored (although the save task will also include filtering, I think it's good to have a double-check); as well as it's corresponding test.

Regarding the DTO names, I updated them to `PointsHistoryResponseDto` and `PointHistoryEntryDto` to be more consistent with the app (but the entry DTO is actually an entry, not a request).

Hope this leaves this PR clean and neat = )

## All tests still passing after update = )

image

pantalois (approved):
Lucy! Looks good to me, good job.

carlosPc1987 (approved):
Approved.
Nice work on the point history logic, the mapper, and the test coverage, it’s clear and consistent with the rest of the module.
I mentioned the merge with #1108 only as a heads-up for whoever merges into the base branch: both PRs add different methods to the same service and repository, so the final merge will simply combine them. There’s no dependency between the two, and no extra work on your side—your PR is good to go as is. Thanks for the implementation.

itachallenge-challenge/src/main/java/com/itachallenge/gamification/service/PointsServiceImpl.java
```diff
@@ -0,0 +1,49 @@
...
+@RequiredArgsConstructor
+public class PointsServiceImpl implements PointsService {
+
+    private final UserScoreRepository userScoreRepository;
+    private final GamificationMapper gamificationMapper;
+
+    @Override
+    public Mono<PointsHistoryResponseDto> getUserPointsHistory(UUID userId) {
+        return userScoreRepository.findByUserIdOrderByCreatedAtDesc(userId)
+                .distinct(UserScoreDocument::getChallengeId)
```
ivilarop, line 27: The rule is:

Is a maximum of 1 record per (userId, challengeId) correct?

With that rule, the cleanest and simplest solution is:

✅ Recommended decision:

Add a unique index per (user_id, challenge_id) to UserScoreDocument

Remove the `distinct()` function from the GET request (because there will no longer be duplicates)

Why?

- `distinct()` in read operations is a workaround; the rule should be implemented in persistence.
- The total will simply be the sum of the points earned for the user's documents.
- This prevents data corruption from the source (double inserts, retries, etc.).

Handling DuplicateKeyException as a clean 409 response is ideal; if out of scope for this PR, we can track it as tech debt, but the unique index should still be added to prevent duplicates at the source.

itachallenge-challenge/src/main/java/com/itachallenge/gamification/mapper/GamificationMapper.java
ivilarop, line 1: I would remove the GamificationMapper for now. The mapping is very simple (2 fields + null handling) and is only used in this service, so the extra class doesn't reduce complexity; it just adds a navigation jump and clutter. Furthermore, the name GamificationMapper is too generic and breaks naming/semantics: it looks like a mapper for the entire module, but in reality, it only maps point history entries (UserScoreDocument -> PointHistoryEntryDto).

Proposal: move the mapping to a private method in PointsServiceImpl (or to the stream itself using a lambda function), and if future reuse or further transformations arise (ISO date format, derived fields, etc.), then extract a PointsHistoryMapper with a specific name.

itachallenge-challenge/src/main/java/com/itachallenge/gamification/service/PointsService.java
ivilarop, line 1: Currently, we're mixing naming conventions within the gamification: UserScoreDocument/UserScoreRepository (Score) vs. PointsService (Points). This makes it difficult to understand the module's core concept.

Since we already use UserScore as a persistence aggregate and the service will likely grow beyond history (e.g., total score, ranking, etc.), I suggest aligning the naming conventions and renaming PointsService → UserScoreService.

This maintains a consistent mental model across the entire package and prevents the terminological drift of "Score vs. Points" as the module evolves.

itachallenge-challenge/src/test/java/com/itachallenge/challenge/dto/gamification/PointsHistoryResponseDtoTest.java
```diff
@@ -0,0 +1,24 @@
+package com.itachallenge.challenge.dto.gamification;
+
+import org.junit.jupiter.api.Test;
+import java.util.List;
+import static org.assertj.core.api.Assertions.assertThat;
+
+class PointsHistoryResponseDtoTest {
+
+    @Test
+    void testDtoStructure() {
```
ivilarop, line 10: This test only checks Lombok-generated builder/getters. It increases coverage but doesn’t protect behavior. If we want to validate the API contract, a Jackson serialization test (e.g., date mapping) would provide real value.

itachallenge-challenge/src/test/java/com/itachallenge/gamification/service/PointsServiceImplTest.java
```diff
@@ -0,0 +1,149 @@
...
+        var result = pointsServiceImpl.getUserPointsHistory(userId);
+
+        StepVerifier.create(result)
+                .expectNextMatches(response ->
+                        response.getTotalPoints() == 35)
+                .verifyComplete();
+    }
+
+    @Test
+    void givenMultipleScores_whenGetUserPointsHistory_thenHistoryListIsCorrectlyMapped() {
```
ivilarop, line 52: Valuable test for history mapping, but it currently mocks the mapper (so we’re not validating the real mapping). If we inline/remove the mapper, this test can assert the real DTO fields without extra mocks.

itachallenge-challenge/src/test/java/com/itachallenge/gamification/service/PointsServiceImplTest.java
```diff
@@ -0,0 +1,149 @@
...
+
+        var result = pointsServiceImpl.getUserPointsHistory(userId);
+
+        StepVerifier.create(result)
+                .expectNextMatches(response -> response.getTotalPoints() == 10)
+                .verifyComplete();
+    }
+
+    @Test
+    void givenValidScores_whenGetUserPointsHistory_thenHistoryIsReturnedInCorrectOrder() {
```
ivilarop, line 106: Blocker: the test setup contradicts findByUserIdOrderByCreatedAtDesc by returning the scores in ascending order. Either mock the Flux in descending order (latest first) and assert the service preserves it, or explicitly sort in the service if ordering must be guaranteed regardless of repository order.

Lucy-SD:
This PR was accidently clossed. It's changes appear here:

https://github.com/IT-Academy-BCN/ita-challenges-backend/pull/1118

itachallenge-challenge/src/main/java/com/itachallenge/gamification/service/PointsServiceImpl.java
Lucy-SD: Yes, this was intended to be just a double check, but since Carlos is implementing the unique index in UserScoreDoc, this becomes redundant. Therefore, it was removed.

itachallenge-challenge/src/main/java/com/itachallenge/gamification/mapper/GamificationMapper.java
Lucy-SD: The idea behind having a separate mapper was to be able to reuse it for ranking and future implementations. At this stage of development, I understand that it's not necessary. Removed.

itachallenge-challenge/src/main/java/com/itachallenge/gamification/service/PointsService.java
Lucy-SD: Clear. Done.

itachallenge-challenge/src/test/java/com/itachallenge/challenge/dto/gamification/PointsHistoryResponseDtoTest.java
Lucy-SD: Done.

itachallenge-challenge/src/test/java/com/itachallenge/gamification/service/PointsServiceImplTest.java
Lucy-SD: Ok.

itachallenge-challenge/src/test/java/com/itachallenge/gamification/service/PointsServiceImplTest.java
Lucy-SD: Fixed. It was decided to keep the repository stored in descending order (since it is a more logical storage strategy), ordering it in the service to ensure that the frontend receives the requested ascending order for the graphs.

PR #1111, opened 2026-02-22
[BE] PR 2.1/3: Ranking Controller and sorting logic (#207)

pantalois:
## 📄 Description ##

⚠️ This PR depends on the approval of the PR [feature/190-1/gamification-persistence-infra.](https://github.com/IT-Academy-BCN/ita-challenges-backend/pull/1107)

The context of this implementation is the requirement to generate a user ranking displaying:

- Username + Accumulated points
- Descending order (users with the highest score at the top)
- The endpoint returns an HTTP 200 OK when the operation is successfully completed.

## 🔨 Technical Decisions ##

The implementation follows a clean Spring Boot layered architecture, maintaining separation of concerns:
- Controller: exposes the REST endpoint.
- Service: acts as an intermediary layer between the controller and the repository.
- Repository: contains the data access logic.

The sorting logic is not performed in memory. Instead, it is executed directly in the database using a MongoDB aggregation pipeline, which:

- Delegates processing to the database engine.
- Improves performance.
- Reduces load on the application layer.

Created Endpoint
```
GET http://localhost:8762/itachallenge/api/v1/users/ranking
```

Behavior:

- Retrieves the user ranking.
- Returns the list sorted in descending order by points.
- Responds with 200 OK when the execution is successful.

## 🔎 Testing ##

Tests have been added for all newly created components (controller, service, and repository) to ensure correct behavior and proper coverage.

This PR is ready for review once the base branch mentioned above is approved.

## 😓 Technical Debt ##

Pagination was intentionally omitted at this stage. The current dataset is bounded and small enough that full-list retrieval does not impact latency meaningfully. It is noted as tech debt and can be introduced when the dataset grows or if performance metrics indicate a bottleneck.

## 📸 Screenshoots  ##

image

image

image

image

image

image

image


Lucy-SD (changes requested):
"Hi Lois ¡! Great work implementing aggregation in MongoDB. Delegating the calculation to the database is a smart move for performance right now. I have a few observations to ensure we meet the acceptance criteria:

* **Aggregation by ID:** In the repository, we're grouping by username. We should use `user_id` to avoid collisions or problems if a user changes their alias.


* **Error handling:** I've seen that in the pull request description you talk about a tech debt, but in a reactive environment like ours, letting a database exception explode without an `.onErrorResume` operator or clear handling can leave connections hanging or return unprofessional generic 500 errors. Suggestion: even if he's the "MVP," he should add basic handling to ensure that if the database fails, the client receives a structured error message.

* **Pagination**: Since the requirement asks for optimization for "the entire list of students," basic pagination is often necessary to meet the "low latency" requirement for long lists. At the very least, you should explain why you chose not to paginate at this stage.

* **Conflict solving:** Probably since last updates somo problemes came up. Please solve them = )

itachallenge-challenge/src/main/java/com/itachallenge/gamification/exception/ServiceException.java
```diff
@@ -0,0 +1,7 @@
+package com.itachallenge.gamification.exception;
+
+public class ServiceException extends RuntimeException {
```
Apani13: I think that the naming of this exception could be more self explanatory, I feel that right now it sounds a bit generic :)

Apani13 (approved):
I have downloaded the branch and ran all the tests on my local environment 👌

I only have one suggestion related to the naming of an exception class, a side from that, I think everything els lloks fine!

I approved  the PR already since my suggestion does not afect any file in the code :)

## Good Job!

itachallenge-challenge/src/main/java/com/itachallenge/gamification/exception/ServiceException.java
pantalois: Hi @Apani13! You're right, I added a bit more context by modifying the name. I also refactored all the related files that used that class.

pantalois: If for you is okey we can resolve this conversation.

Apani13: Absolutly! :D

## Good Job!

itachallenge-challenge/src/test/java/com/itachallenge/gamification/repository/UserScoreRepositoryTest.java
```diff
@@ -50,4 +53,46 @@ void givenExistingScores_whenFindByUsername_thenReturnsSortedByDescendingDates()
                 .expectNextMatches(s -> s.getPointsEarned() == 1)
                 .verifyComplete();
     }
+
```
Lucy-SD, line 60: I think we have an important issue here. If we are agregating by userId, but then use randomUUID for each doc created, we are recording scores from two different users . . .

itachallenge-challenge/src/main/java/com/itachallenge/gamification/repository/UserScoreRepository.java
```diff
@@ -1,12 +1,24 @@
...
 
 import java.util.UUID;
 
+@Repository
 public interface UserScoreRepository extends ReactiveMongoRepository<UserScoreDocument, UUID> {
 
     Flux<UserScoreDocument> findByUserIdOrderByCreatedAtDesc(UUID userId);
+
+    @Aggregation(pipeline = {
+            "{ $group: { _id: '$user_id', username: { $first: '$username' }, points: { $sum: '$points_earned' } } }",
```
Lucy-SD: Maybe we should use {$last: '$username'} if we want the most recent username to be shown ¿?

Maybe we could generate a technical debt: the ranking should return the current username from the users/identity microservice to avoid inconsistent data in the points history . . .

Lucy-SD (changes requested):
Hi Lois, I can see you've done some nice improvements. Just a couple of notes added. I believe that the test issue should be resolved before merge, while the naiming issue could be addressed in the future . . .

Likewise, great improvements = )

PR #1146, opened 2026-04-08
[BE] PR1/3 — Prepare weekly leaderboard data aggregation (#280)

carlosPc1987:
https://github.com/IT-Academy-BCN/ita-challenges-backend/pull/1146

📅 Weekly leaderboard data — technical description

📌 Scope
- Add groundwork for the weekly leaderboard without changing public API contracts.
- Add `WeeklyWindow` helper for ISO week bounds (Monday 00:00 inclusive → Sunday inclusive through end of day, `LocalTime.MAX` / 23:59:59.999999999).
- Add `UserScoreRepository.aggregateUserScoresByPeriod(fromInclusive, toInclusive)`.
- Leave the existing global aggregation (`aggregateUserScores`) unchanged.
- Add/extend tests for: weekly window calculation; inclusion/exclusion of records inside/outside the weekly range.

🚫 Out of scope
- Controller changes.
- Public route/DTO changes.
- HTTP/JSON contract changes.
- League assignment and level calculation (PR2).
- OpenAPI updates (PR3).

🧪 Tests
- `WeeklyWindowTest`: ISO window (incl. year/week boundary cases).
- `UserScoreRepositoryIntegrationTest`: period derived from `WeeklyWindow`; data inside and outside the range; edge case at the last instant of Sunday (`toInclusive`); `ReactiveMongoTemplate.insert()` where fixed timestamps are needed (`@CreatedDate` auditing).
- - `UserScoreRepositoryIntegrationTest`: also covers **zero results** for `aggregateUserScoresByPeriod` when the collection is empty (regression guard).
- Local build and tests passed.

📝 Technical notes
- Weekly period follows the ISO calendar (Monday–Sunday).
- Range generation is intended for the service layer; the repository only filters/aggregates by `created_at`.
- Week bounds are computed in the reference `ZonedDateTime`’s zone; the service should use the agreed league timezone.
- If the PO confirms a different week definition, it will be adjusted in PR2/PR3.
- **Date/time library choice:** PR1 keeps `WeeklyWindow` on `java.time` only (no new dependencies; narrow ISO Monday–Sunday bounds with tests). Follow-up: evaluate **ThreeTen-Extra** / `YearWeek` if PR2/PR3 adds heavier timezone/DST or more aggregation windows — see review discussion.

📄 CHANGELOG
Not updated in this PR (no public API changes or user-visible behavior).

image

### DIFF
--- a/itachallenge-challenge/src/main/java/com/itachallenge/gamification/util/WeeklyWindow.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/gamification/util/WeeklyWindow.java
@@ -1,0 +1,37 @@
+package com.itachallenge.gamification.util;
+
+import java.time.DayOfWeek;
+import java.time.LocalDateTime;
+import java.time.LocalTime;
+import java.time.ZonedDateTime;
+import java.time.temporal.TemporalAdjusters;
+import java.util.Objects;
+
+public class WeeklyWindow {
+    private final LocalDateTime fromInclusive;
+    private final LocalDateTime toInclusive;
+
+    public WeeklyWindow(LocalDateTime fromInclusive, LocalDateTime toInclusive) {
+        this.fromInclusive = Objects.requireNonNull(fromInclusive, "fromInclusive must not be null");
+        this.toInclusive = Objects.requireNonNull(toInclusive, "toInclusive must not be null");
+    }
+
+    public static WeeklyWindow fromReferenceDateTime(ZonedDateTime referenceDateTime) {
+        Objects.requireNonNull(referenceDateTime, "referenceDateTime must not be null");
+        ZonedDateTime startOfWeek = referenceDateTime
+                .with(TemporalAdjusters.previousOrSame(DayOfWeek.MONDAY))
+                .with(LocalTime.MIN);
+        ZonedDateTime endOfWeek = startOfWeek
+                .plusDays(6)
+                .with(LocalTime.MAX);
+        return new WeeklyWindow(startOfWeek.toLocalDateTime(), endOfWeek.toLocalDateTime());
+    }
+
+    public LocalDateTime getFromInclusive() {
+        return fromInclusive;
+    }
+
+    public LocalDateTime getToInclusive() {
+        return toInclusive;
+    }
+}
--- a/itachallenge-challenge/src/test/java/com/itachallenge/gamification/repository/UserScoreRepositoryIntegrationTest.java
+++ b/itachallenge-challenge/src/test/java/com/itachallenge/gamification/repository/UserScoreRepositoryIntegrationTest.java
@@ -97,6 +97,19 @@ class UserScoreRepositoryIntegrationTest {
     }
 
     @Test
+    void givenEmptyDatabase_whenAggregateUserScoresByPeriod_thenReturnsEmpty() {
+        userScoreRepository.deleteAll().block();
+
+        WeeklyWindow window = WeeklyWindow.fromReferenceDateTime(
+                ZonedDateTime.of(2026, 4, 8, 12, 0, 0, 0, ZoneId.of("Europe/Madrid")));
+        LocalDateTime from = window.getFromInclusive();
+        LocalDateTime to = window.getToInclusive();
+
+        StepVerifier.create(userScoreRepository.aggregateUserScoresByPeriod(from, to))
+                .verifyComplete();
+    }
+
+    @Test
     void givenScoresInsideAndOutsidePeriod_whenAggregateUserScoresByPeriod_thenReturnsOnlyInsideRange() {
         userScoreRepository.deleteAll().block();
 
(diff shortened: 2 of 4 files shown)


eariasvalor (changes requested):
Great job, just one issue to address before merging.

* `@CreatedDate` may override fixed dates in tests**

`UserScoreDocument` uses `@CreatedDate`, so Spring overwrites `createdAt` with `LocalDateTime.now()` on save, ignoring the builder value. The test passes today by coincidence (2026-04-08 falls within the range) but will fail tomorrow.

Use `ReactiveMongoTemplate.insert()` instead of `saveAll()` for documents with fixed dates — it bypasses the auditing lifecycle. `ReactiveMongoTemplate` was already injected in this class before this PR removed it.

JordiCasas87 (commented):
Hello Carlos,
Everything looks correct to me. Just a couple of small questions (not sure if they are relevant, so I’m raising them just in case):

WeeklyWindow.java:
fromReferenceDateTime converts a ZonedDateTime to LocalDateTime, dropping timezone information. Could this affect boundary filtering if different timezones are involved, or is the system expected to run under a single timezone?

UserScoreRepositoryIntegrationTest.java:
The test uses 23:59:59 as the upper boundary, while the logic uses LocalTime.MAX. It might be worth adding a boundary case to fully align with the intended behavior.

These are just minor suggestions from my side. I’ll be happy to approve once the issue raised by Esther is resolved.

carlosPc1987:
Hi Esther, @eariasvalor 

Thanks for the review — I’ve pushed an update that addresses your comment. The period aggregation test now uses ReactiveMongoTemplate.insert() for the documents with fixed createdAt values (and re-injects ReactiveMongoTemplate), so @CreatedDate auditing no longer overwrites the timestamps from saveAll(). The scenario should stay stable regardless of the run date.

Let me know if that works for you.

Hi Jordi , @JordiCasas87 

Thanks — good catches. Weekly boundaries are computed in the reference ZonedDateTime’s zone, then converted to LocalDateTime for storage/query; multi-TZ edge cases depend on us always passing the agreed “league” zone from the caller. I’ve aligned the period test with WeeklyWindow’s toInclusive (LocalTime.MAX) and added a test for a score timestamped at that exact upper bound.

eariasvalor (approved):
Fixed as suggested — `mongoTemplate.insert()` correctly bypasses `@CreatedDate` auditing. The boundary test is a nice addition. LGTM

JordiCasas87 (approved):
All good after the changes. Thanks for addressing the feedback.

itachallenge-challenge/src/main/java/com/itachallenge/gamification/util/WeeklyWindow.java
```diff
@@ -0,0 +1,37 @@
+package com.itachallenge.gamification.util;
+
+import java.time.DayOfWeek;
+import java.time.LocalDateTime;
+import java.time.LocalTime;
+import java.time.ZonedDateTime;
+import java.time.temporal.TemporalAdjusters;
+import java.util.Objects;
+
+public class WeeklyWindow {
```
ivilarop, line 10: Before keeping `WeeklyWindow` as a custom utility class, check whether the project already has or could benefit from a library that handles ISO week calculations natively — for example **ThreeTen-Extra**, which provides `YearWeek` and handles Monday-to-Sunday boundaries, timezones and edge cases out of the box.

Custom `util` classes tend to become a catch-all over time, and date/time logic is exactly the kind of problem where a well-maintained library reduces risk and eliminates the need for bespoke edge case handling — as the timezone issue caught during review already illustrated.

If no suitable library is available or adding one is not justified, document the reasoning and keep `WeeklyWindow` — but make that a conscious decision rather than a default.

ivilarop (dismissed):

itachallenge-challenge/src/main/java/com/itachallenge/gamification/util/WeeklyWindow.java
carlosPc1987: Thanks @ivilarop — agreed on the risk of ad-hoc date/time code.

Decision for this PR: I’m keeping WeeklyWindow on java.time only on purpose: this PR is meant to stay a small foundation change without adding dependencies, and the behaviour we need is narrow (ISO week Monday–Sunday with inclusive bounds), already covered by tests including year/week boundaries and the Sunday upper bound.

I’ve documented that choice in the PR description under Technical notes (ThreeTen-Extra / YearWeek as a follow-up if PR2/PR3 introduces heavier timezone/DST rules or more aggregation windows), so it’s an explicit trade-off rather than a default.

itachallenge-challenge/src/test/java/com/itachallenge/gamification/repository/UserScoreRepositoryIntegrationTest.java
```diff
@@ -92,6 +96,81 @@ void givenEmptyDatabase_whenAggregateUserScores_thenReturnZero() {
                 .verifyComplete();
     }
 
+    @Test
```
kenan-it-academy, line 112: Just as a bit of extra safety (a bit of paranoia regarding future regressions) I would also like a `.aggregateUserScoresByPeriod()` Zero test included. Other than that, this PR seems well crafted and ready to go once that's done!

kenan-it-academy (changes requested):
There's a test I'd like added. Other than that, this PR seems pretty clean!

eariasvalor (approved):
Everything looks good to me!

JordiCasas87 (approved):
Looks good to me. Requested changes have been addressed.

kenan-it-academy (approved):

PR #1149, opened 2026-04-09
PR 2/3 — feat(gamification): weekly leagues split in LeaderboardService (PR2)

carlosPc1987:
## Summary
Adds **weekly league assignment** in the service layer: users are ranked by **weekly points** for the **current ISO week** and split into **Gold (10)**, **Silver (20)**, and **Bronze (remaining)**, with **alphabetical tie-break** on equal points.

**No controller, routes, or public JSON contract changes** in this PR (PR3 will expose HTTP/OpenAPI).

Parent: [BE] Gamification: league ranking system (weekly → Gold / Silver / Bronze) #1146 

---

## Weekly window (agreed & documented for Iván’s review)
- **Definition:** **ISO week** (Monday start → Sunday end), consistent with frontend usage.
- **Who builds the range:** the **service** computes `[fromInclusive, toInclusive]` via `WeeklyWindow.fromReferenceDateTime(ZonedDateTime.now(Europe/Madrid))` and passes it to the repository.
- **Repository responsibility:** only **filters + aggregates** by `created_at` within that range (`aggregateUserScoresByPeriod`), unchanged from PR1.

If the PO requests a different week definition later, we adjust in a follow-up (same split as discussed with Iván: document first, then implement).

---

## Business rules (aligned with PO/mentor)
- **Recalculation cadence:** weekly reset on the agreed cycle (Sunday / end of ISO week — same discussion as daily); ranking uses **weekly points** in that window.
- **League sizes:** Gold **10**, Silver **20**, Bronze **all remaining** users with weekly score in that week.
- **Ordering:** higher `total_points` first.
- **Tie-break:** **case-insensitive alphabetical** by `username`.

---

## What changed
- `LeaderboardService`: new `getWeeklyLeagues()`.
- `LeaderboardServiceImpl`: weekly aggregation + sort + split into three lists.
- `WeeklyLeaguesResult`: DTO for service output (`gold`, `silver`, `bronze` as `List<LeaderboardEntryDto>`), mirroring existing leaderboard entry shape (`username`, `total_points` via `LeaderboardEntryDto`).
- Unit tests for split + tie-break + error mapping.

---

## Scope / acceptance criteria
- Aligned with updated (#1146 ): PR2 covers the weekly league **split + ordering** in the service layer only.  
- **Position, level (from full history), and HTTP/query contract** are **out of PR2** — see PR3 / PO follow-up.  
(User story acceptance criteria updated to match.)

---

## Out of scope (PR3)
- REST endpoint, query params, OpenAPI.
- Historical global leaderboard (`GET /itachallenge/api/v1/leaderboard`) — **must remain unchanged**.

---

## Testing
- [x] `./gradlew :itachallenge-challenge:test --tests "com.itachallenge.gamification.service.LeaderboardServiceImplTest"`
- Test coverage highlights
· Weekly league split sizes (Gold/Silver/Bronze).
· Tie-break behavior on equal points.
· Error mapping in service.
· Repository period aggregation:
   inside/outside range,
   empty database returns empty result,
· Sunday last instant included (toInclusive),
   explicit ISO boundary coverage: Sunday included / Monday excluded.
---

## CHANGELOG
Not updated (no user-visible API change in this PR).

image


eariasvalor (changes requested):
Hey, good work! The logic is clean and well-structured. A few things to address before approving:
1. Missing fields in the response
The acceptance criteria require each entry to include username, position, weeklyPoints and level (computed at runtime from full history). Currently LeaderboardEntryDto only carries username and total_points. Was this descoped with the PO?
2. Minor code style issue in mapToEntryDto — missing space before !=null. Sonar will likely flag it.
3. WeeklyLeaguesResult is currently in the service package — since PR3 will expose this via HTTP, would it make more sense to place it in dto directly to avoid an extra mapping layer?

carlosPc1987:
Thanks @eariasvalor — great review.

I’ve fixed the small style issue in mapToEntryDto (!= null spacing).
position / level: for the scope we aligned with PO/mentor for this delivery, PR2 focuses on the weekly league split (Gold/Silver/Bronze) using weekly points and alphabetical tie-break. Explicit rank position and level are not part of this service PR; they imply extra rules (e.g. level from full history) and should be confirmed with PO. If required, we can add them in PR3 (HTTP contract) or a follow-up once criteria are explicit.

WeeklyLeaguesResult in the service package: I’m keeping it as a service-layer model on purpose; PR3 will map it to the public HTTP DTOs so we don’t mix domain/service output with the API contract.

ISO week / date logic: following Iván’s feedback on PR1, we’re consciously using java.time + WeeklyWindow for the ISO Monday–Sunday window with tests for edge cases, and we’ll track evaluating something like ThreeTen-Extra / YearWeek if date-time complexity grows in later PRs — so the choice is explicit, not accidental.

eariasvalor:
Thanks for the detailed response, everything makes sense! One last thing: given that position, level, and the query param approach have been descoped or adjusted based on PO alignment, would it make sense to update the user story to reflect the actual agreed scope? That way the acceptance criteria stay in sync with what's actually being delivered and future reviewers have a clear reference.

carlosPc1987:
I’ll update the user story / issue (#280) so the acceptance criteria match what we’re actually shipping in this chain: weekly leagues split (Gold / Silver / Bronze), weekly points + alphabetical tie-break, without explicit rank position, level-from-full-history, or extra query-param behaviour in PR2, with the HTTP contract and any remaining fields handled in PR3 (or follow-ups once the PO locks the details). That should make it easier for the next reviewers too.

eariasvalor:
Hey, just checked the user story and it hasn't been updated yet — position, level, the query param approach on GET /leaderboard, and the rest of the original acceptance criteria are still there unchanged. Could you update it before we approve?

carlosPc1987:
Hey @eariasvalor 

Thanks for flagging it.

I’ve updated the user story so the acceptance criteria match what we’re delivering:

Issue: https://github.com/IT-Academy-BCN/ita-challenges/issues/298

Changes: title + description aligned with the PO/mentor agreement (weekly ISO window, Gold 10 / Silver 20 / Bronze rest, alphabetical tie-break, no level in this sprint, PR2 = service split only, PR3 = HTTP / gold[] silver[] bronze[], global leaderboard unchanged).

Let me know if anything still looks off on your side.

eariasvalor (approved):
User story updated and aligned with what was shipped. LGTM! ✅

Andrescalvo22 (changes requested):
Good work overall — the structure and separation of concerns look solid 👍

A few points worth reviewing before approval:

1. Duplicate field assignment
There is a duplicated `.totalPoints(...)` assignment in `mapToEntryDto`. One of them should be removed.

2. Null safety in sorting
The tie-break comparator is not null-safe. If a username is null, it may cause a NullPointerException. Consider using a null-safe comparator.

carlosPc1987:
Thanks for the review  @Andrescalvo22 — addressed in the latest push.

totalPoints: mapToEntryDto only sets totalPoints once; if you still saw a duplicate, it should be from an older revision of the diff.
Null-safe tie-break: the weekly sort now uses Comparator.nullsLast(String.CASE_INSENSITIVE_ORDER) on username (points were already using nullsLast on the primary sort).
Please let me know if you’d like anything else adjusted.

Andrescalvo22 (approved):
It looks fine to me! Approved!

itachallenge-challenge/src/test/java/com/itachallenge/gamification/repository/UserScoreRepositoryIntegrationTest.java
```diff
@@ -185,6 +185,42 @@ void givenScoreAtSundayLastNanosecond_whenAggregateUserScoresByPeriod_thenInclud
                 .verifyComplete();
     }
 
+    @Test
+    void givenScoresAtSundayEndAndMondayStart_whenAggregateUserScoresByPeriod_thenSundayIncludedAndMondayExcluded() {
```
kenan-it-academy, line 189: I think my only comment is that this test is too long. I would recommend using a helper function to reduce the footprint of the builders within the test, to improve readability.

kenan-it-academy (approved):
There is a minor readability change I would appreciate, but I'm not going to block the PR for it in this case. Just something I'll likely bring up later 🙃

PR #1153, opened 2026-04-09
feat(gamification): implement read-only endpoint for available activity types (#305)

### LINKED ISSUE #305: Score: creación de compilador e intérprete Java
Es necesario crear en la capa de Servicios un servicio que, recibido un String sea capaz de:
* Compilar el código y devolver mensaje de error en caso de error de compilación (objeto Jackson JSON con error message del compiler)
* Interpretarlo con valores de entrada y verificar que los valores de salida son correctos.
   * Valores de salida correctos: debe devolver objeto Jackson JSON según se ve en el file solution.user.txt
   * Valores de salida incorrectos: mensaje de error tipo.

**REFERENCIAS:**

* https://docs.oracle.com/javase%2F7%2Fdocs%2Fapi%2F%2F/javax/tools/JavaCompiler.html
* https://medium.com/@Mark.io/running-arbitrary-java-code-on-the-fly-d5003ebd46a8
* https://datacadamia.com/lang/java/dynamic

JordiCasas87:
Closes #305

## Description

This PR implements a read-only endpoint to expose the available activity types defined in `ActivityType`, so the frontend can consume them dynamically without hardcoding values.

The new endpoint added is:

- `GET /itachallenge/api/v1/activity-types`

It returns the list of available activity types currently defined in the system:
- `CODE_REVIEW`
- `PRESENTATION`
- `CHALLENGE_COMPLETED`

## Changes

- Added `ActivityTypeController` to expose the new read-only endpoint
- Added `ActivityTypeService` and `ActivityTypeServiceImpl`
- Added `ActivityTypeResponseDto` for the endpoint response
- Updated `CHANGELOG.md`
- Updated `postman/ITA-Challenges.postman_collection.json`

## Scope

Included:
- Read-only retrieval of available activity types
- Swagger/OpenAPI exposure for the new endpoint
- Unit and controller tests
- Postman collection update

Not included:
- Point registration logic
- Score persistence changes
- Ranking or history changes

## Validation

- Local service test passed
- Local controller tests passed
- Endpoint verified manually in Swagger with expected response

TestPR2Service
TestPR2Controller
SwaggerENDPOINT

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,17 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### [itachallenge-challenge-3.5.0-RELEASE] - 2026-04-09
+
+[Gamification]
+
+### Added
+
+#### Activity Types API
+- Added `ActivityType` enum as the shared source of truth for available point-generating activity types.
+- Added `GET /itachallenge/api/v1/activity-types` endpoint to expose available activity types for frontend consumption.
+- Added service, DTO, and controller support for activity type retrieval, along with unit and controller tests.
+
 ## [itachallenge-challenge-3.4.0 - RELEASED] 2026-03-23
 
 [Gamification]
--- a/itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/gamification/ActivityTypeController.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/gamification/ActivityTypeController.java
@@ -29,7 +29,7 @@ public class ActivityTypeController {
             responses = {
                     @ApiResponse(
                             responseCode = "200",
-                            description = "OK - empty array if none.",
+                            description = "OK - returns an object with an empty activityTypes list if none.",
                             content = @Content(
                                     mediaType = "application/json",
                                     schema = @Schema(implementation = ActivityTypeResponseDto.class)
--- a/itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/gamification/ActivityTypeController.java
+++ b/itachallenge-challenge/src/main/java/com/itachallenge/challenge/controller/gamification/ActivityTypeController.java
@@ -1,0 +1,49 @@
+package com.itachallenge.challenge.controller.gamification;
+
+import com.itachallenge.challenge.dto.MessageDto;
+import com.itachallenge.challenge.dto.gamification.ActivityTypeResponseDto;
+import com.itachallenge.gamification.service.ActivityTypeService;
+import io.swagger.v3.oas.annotations.Operation;
+import io.swagger.v3.oas.annotations.media.Content;
+import io.swagger.v3.oas.annotations.media.Schema;
+import io.swagger.v3.oas.annotations.responses.ApiResponse;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+import reactor.core.publisher.Mono;
+
+@Slf4j
+@RestController
+@RequiredArgsConstructor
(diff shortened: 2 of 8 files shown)


andyrouge81 (changes requested):
Nice work on this PR, everything looks clean and well aligned with the objective of exposing a read-only endpoint for activity types. The use of the `ActivityType` enum as the single source of truth is a solid approach, and both controller and service layers are simple and well structured. Tests and Postman collection are also a great addition 

I just have a couple of small suggestions to improve it further:

In the Swagger documentation, the description mentions “empty array if none”, but the endpoint actually returns an object with an empty list (`{ "activityTypes": [] }`). It might be worth adjusting the wording for better accuracy.
In `ActivityTypeServiceImplTest`, the test currently checks fixed values using `containsExactly(...)`. Since the service is based on `ActivityType.values()`, it could be more robust to assert dynamically against the enum values instead. This way, the test won’t break if new activity types are added in the future.

These are just minor improvements, overall the implementation looks solid and ready to go 🚀

Good job Jordi!

JordiCasas87:
Thanks for the feedback!

I've applied the suggested changes:
- Updated the Swagger description for better accuracy
- Adjusted the test to assert dynamically against the enum values

Let me know if everything looks good now 🙂

andyrouge81 (approved):
Looks good to me,

The implementation is clean, aligned with the objective, and tests are well covered. Nice improvement using ActivityType.values() as the source of truth.

Just a very minor detail: @RequiredArgsConstructor in ActivityTypeServiceImpl seems unnecessary since there are no injected dependencies. Not blocking at all, just a small cleanup suggestion.

Good job Jordi!

eariasvalor (approved):
Agreed with the above, both suggestions are valid — especially the dynamic assertion in the controller test. Clean implementation overall, approved! ✅

JordiCasas87:
Thanks for the review and approval, Esther! 🙂

Andrescalvo22 (changes requested):
Nice work on this PR 👏 The endpoint is clean, easy to understand, and follows a good separation of concerns between controller, service, and DTO. The use of WebFlux (Mono) is consistent, and the tests cover the main scenarios (success, empty response, and error), which is great to see.

I do have a few suggestions that could help improve the implementation:

1. DTO typing (important)
Right now the response is using List<String> for activityTypes. It would be better to use the enum directly (List<ActivityType>) instead of mapping it to String.

The main reason for this is that using the enum makes the contract much stronger and safer:
- It avoids potential issues with typos or invalid string values
- It ensures type safety across the application
- It makes the API more self-documented (both for developers and in Swagger)
- Jackson can handle enum serialization automatically, so there's no need to convert them manually

2. Service mapping
Related to the previous point, the service currently does:
.map(Enum::name)

If you switch the DTO to use ActivityType, you can return the enum values directly without converting them to String, simplifying the code.

But in order of that, Good Job!

eariasvalor:
@Andrescalvo22 Good points, though I'd argue the current List<String> approach seems actually intentional and valid — DTOs are meant to decouple the internal domain from the HTTP contract, so converting via .map(Enum::name) keeps that boundary clean. If ActivityType changes internally, the API contract stays unaffected. For me it's more of a design preference than a clear improvement.

JordiCasas87:
Thanks for the suggestion!

After reviewing it with the team, we decided to keep the current approach using List<String> in the DTO to maintain the API contract decoupled from the domain, as discussed.

Let me know if that works for you 👍

eariasvalor (approved):
Everything looking good!

Andrescalvo22 (approved):
It sounds good to me, Approved!

kenan-it-academy:
So just to confirm, the purpose of this PR is to:

- Add the following endpoint: `GET /itachallenge/api/v1/activity-types`
- Make it return the following list:
    - CODE_REVIEW
    - PRESENTATION
    - CHALLENGE_COMPLETED

Is this correct? Cause it seems WAY too large of a PR to accomplish this @JordiCasas87

JordiCasas87:
Just to clarify, the PR size is now correct.

The previous large diff was caused by unrelated commits from `develop`, which have been removed.

The PR now only includes the intended changes:

* 8 files
* ~240 lines

This reflects the actual scope of the endpoint implementation.

kenan-it-academy (approved):
Simple and to the point, this all makes sense to me

PR #1151, opened 2026-04-09
PR3/3 — feat(gamification): expose weekly leagues leaderboard endpoint (#280)

carlosPc1987:
## Summary
PR 3/3 

This PR exposes weekly leagues through the HTTP layer by adding a dedicated endpoint and response DTO, while keeping the existing global leaderboard contract unchanged.

**Tracking:** Umbrella [ita-challenges#280](https://github.com/IT-Academy-BCN/ita-challenges/issues/280) — detailed AC in [ita-challenges#298](https://github.com/IT-Academy-BCN/ita-challenges/issues/298).

**Stack:** This PR targets `feature/280-pr2-weekly-leagues-service` (PR2 chain). Depends on PR1 work merged via the stacked branch / related PRs in this repo.

---

## What this PR adds

- New endpoint:
  - `GET /itachallenge/api/v1/leaderboard/weekly`
- New response DTO:
  - `WeeklyLeaguesResponseDto` with:
    - `gold`
    - `silver`
    - `bronze`
- Controller wiring:
  - `LeaderboardController` now maps `LeaderboardService.getWeeklyLeagues()` into the HTTP response.
- OpenAPI annotations for the new endpoint.

---

## Contract / behavior

- Existing endpoint remains unchanged:
  - `GET /itachallenge/api/v1/leaderboard` (global historical ranking)
- New endpoint returns weekly split:
  - Gold: top 10
  - Silver: next 20
  - Bronze: remaining
- Entries reuse existing convention through `LeaderboardEntryDto` (`username`, `total_points`).

---

## Out of scope

- Weekly split calculation logic (already handled in PR2).
- Weekly data window/repository aggregation logic (already handled in PR1).
- Any breaking changes to existing global leaderboard API.

---

## Tests

- **`LeaderboardControllerTest`**
  - success response with `gold/silver/bronze`
  - empty arrays response
  - service error -> `5xx`

- **`LeaderboardServiceImplTest`**
  - split into leagues preserving repository order
  - boundaries:
    - `<10` users -> only `gold`
    - `=10` users -> only `gold`
    - `=30` users -> `gold` + `silver`, empty `bronze`
  - repository error -> service maps to `5xx` domain exception

- **`UserScoreRepositoryIntegrationTest`**
  - tie-break validation in Mongo sort (`totalPoints desc`, `username asc`) for:
    - global aggregation
    - period aggregation

- **`WeeklyWindowTest`**
  - mid-week and year-boundary window correctness
  - null input safety (`NullPointerException` expected)

[...]

PR #1154, opened 2026-04-10
feat: implement unified point assignment logic in UserScoreService

Andrescalvo22:
## 📝 Description

This PR implements the unified business logic for assigning and persisting user points across all point-generating actions in the system.

Instead of introducing a new service, the existing `UserScoreServiceImpl` has been extended to act as the single entry point for point persistence, ensuring consistency with the unified scoring model defined in PR 1.

---

## 🔧 Scope

This PR focuses on centralizing point assignment logic and making it reusable for both academy activities and challenge completions.

The service now supports:

* Academy activities (`CODE_REVIEW`, `PRESENTATION`)
* Challenge completions (`CHALLENGE_COMPLETED`)

---

### Unified Point Registration

* Added method:
  `registerPoints(UUID userId, ActivityType type, UUID challengeId)`

* The method:

  * Receives `userId`, `ActivityType`, and optional `challengeId`
  * Retrieves point values directly from `ActivityType`
  * Validates `challengeId` based on activity type:

    * Required for `CHALLENGE_COMPLETED`
    * Must be null for other activity types
  * Creates a `UserScoreDocument`
  * Persists it via `UserScoreRepository`

---

## 🧠 Point Assignment Strategy

Point values are now defined within the `ActivityType` enum:

* `CHALLENGE_COMPLETED` → 20 points
* `CODE_REVIEW` → 5 points
* `PRESENTATION` → 10 points

This centralizes scoring logic within the domain model and removes the need for external configuration.

---

## 🧪 Testing

Unit tests in `UserScoreServiceImplTest` have been extended to cover the new logic.

Coverage includes:

* Successful point persistence for both challenge and non-challenge activities
* Validation of `challengeId` constraints
* Error handling for invalid inputs
* Verification of repository interactions

All tests are passing successfully.

---

## 🔗 Integration

This method is intended to be used by the submission flow (#275):

* When a challenge is completed, `SubmissionService` will call `registerPoints`
* `activityType` must be set to `CHALLENGE_COMPLETED`
* `challengeId` must be provided

---

## ✅ Acceptance Criteria

* `UserScoreServiceImpl` extended with unified point assignment method ✔️
* Supports both activity-based and challenge-based scoring ✔️
* Points retrieved from `ActivityType` enum ✔️
* `UserScoreDocument` created and persisted correctly ✔️
* `challengeId` handled according to activity type ✔️
* Unit tests added and passing ✔️
* Code follows project conventions ✔️

---

## 👨‍💻 Technical Debt

* Duplicate activity validation (e.g. preventing multiple point assignments for the same action) is not yet implemented


eariasvalor (changes requested):
Hey, good work overall! Two things worth checking before approving:
1. username is not set in registerPoints
The field is optional in the document, but since the leaderboard falls back to "Anonymous" when username is null, all points registered through this method would show up as "Anonymous" in the ranking. I don't think this should happen. Should username be received as a parameter here?
2. id is not set
Since id is a UUID (not a String), MongoDB won't autogenerate it. You'll need to explicitly set it with UUID.randomUUID(), as done elsewhere in the project.

Andrescalvo22:
@eariasvalor Good catch, thanks!

I’ve updated the method to include `username` as a parameter so leaderboard entries are properly attributed. 
Also added explicit UUID generation for the document id to ensure correct Mongo persistence.

Tests have been updated accordingly.

image

andyrouge81 (commented):
Looks good overall — the PR now aligns much better with the intended unified point registration flow, and the fixes for `username` and explicit UUID assignment address the issues raised in the previous review.

A couple of non-blocking points I’d still highlight for future improvement:

The changelog still references the old `registerPoints(UUID userId, ActivityType type, UUID challengeId`) signature, but the actual method now also includes `username`. It would be good to keep that documentation aligned with the final implementation.
The new tests cover the main happy paths and `challengeId` constraints, which is great, but they currently only verify that `save(...)` was called. In the future, it could be worth asserting the actual persisted UserScoreDocument fields (`id, username, activityType, pointsEarned, challengeId`) to make the tests stronger.

That said, the implementation is consistent with the scope described in the PR, and the current version looks good to me. Nice work.

eariasvalor (changes requested):
Hey Andrés! Before we approve, could you address the two points Andy mentioned?

1. Update the CHANGELOG to reflect the current `registerPoints` signature (including `username`).
2. Add the missing test cases for the null validation branches (`userId`, `type`, `username`) and the repository error path.

This will also help with the Sonar coverage on #1151. Once done, we're good to go!

Andrescalvo22:
@eariasvalor @andyrouge81 Thanks for the feedback!

I’ve updated the CHANGELOG to reflect the current registerPoints signature (including username), and added the missing test cases:

- null validation branches (userId, type, username)
- repository error path

This also improves Sonar coverage for this PR.

Let me know if anything else should be adjusted 🙂

itachallenge-challenge/src/main/java/com/itachallenge/gamification/service/UserScoreServiceImpl.java
```diff
@@ -48,4 +49,39 @@ private PointsHistoryResponseDto buildHistoryResponse(List<UserScoreDocument> do
...
+    @Override
+    public Mono<Void> registerPoints(UUID userId, String username, ActivityType type, UUID challengeId) {
+
+        if (type == null) {
+            return Mono.error(new IllegalArgumentException("ActivityType cannot be null"));
+        }
+
+        if (username == null || username.isBlank()) {
+            return Mono.error(new IllegalArgumentException("username cannot be null or blank"));
+        }
```
eariasvalor: Hi Andrés, as agreed, this validation should be removed to align with the decision of making userName nullable. Also, this has to be documented in the PR description for future consideration.

eariasvalor (changes requested):

eariasvalor (approved):
Code, tests and CHANGELOG are now all aligned — username nullable is consistent across the board, and the conflict with PR #299 is resolved. LGTM! ✅

andyrouge81 (approved):
Looks good to me now.

The main points raised during the review have been addressed:

* `username` / UUID handling was corrected
* tests were extended for validation branches and repository error path
* CHANGELOG was updated
* nullable `username` is now aligned across code, tests, and documentation

Given the latest changes and the current review alignment, I think this PR is in a good shape to move forward. Nice work.

andyrouge81 (commented):
Nice work overall — the unified point assignment logic and validations look solid 👍

One small design observation: `assignPoints(...)` seems to act mainly as a thin wrapper around `registerPoints(...)`, delegating all the logic and just returning the points value. It might be worth clarifying its intent (e.g. if it’s meant only for non-challenge activities) or considering whether a single entry point would be enough to avoid potential confusion in usage.

Also, just a quick note on the `CHANGELOG.md`: it still mentions `registerPoints(UUID userId, String username, ...)`, while the current service signature no longer includes `username`. It might be worth aligning the documentation with the final implementation to avoid confusion.

Not blocking at all, just a thought for future maintainability 🙂

Andrescalvo22:
@andyrouge81 Thanks for the feedback, I will change the CHANGELOG!

PR #1155, opened 2026-04-10
feat(challenge): persist score for completed submissions (#299)

andyrouge81:
## What
This PR refactors completed reto submission scoring so that score persistence goes through the centralized gamification path.

When a submission reaches `SUBMITTED_COMPLETE`, the submission flow now delegates score registration to `UserScoreService.registerPoints(...)` instead of persisting score entries through a separate submission-specific repository path.

## Why
Card `#275` requires completed reto submissions to generate score records in the shared `user_score_history` model using `activityType = CHALLENGE_COMPLETED`, while keeping compatibility with existing score consumers such as:
- `GET /users/{userId}/scores/history`
- `GET /leaderboard`

The previous approach introduced a parallel persistence path in submission scoring. This PR removes that duplication and restores a single score persistence flow.

## Scope implemented in this PR
- detect completed submissions in `SubmissionServiceImpl`
- route completed submission scoring through `UserScoreService.registerPoints(...)`
- remove the submission-specific score persistence bridge
- keep score persistence aligned with the shared gamification model
- preserve `activityType = CHALLENGE_COMPLETED`
- preserve `challengeId` for reto completion records
- keep compatibility with history and leaderboard consumers
- update service tests accordingly

## Scope decision
Although the original card description also mentioned returning `pointsEarned` in the submission response, that part is intentionally not included in this PR.

After alignment with the PO, it was decided not to add `pointsEarned` to the submission DTO in this iteration. This PR focuses on the score persistence flow only.

## Technical note
The submission flow does not currently provide `username`, while the centralized score service expects it for most score registrations.

To keep a single persistence path without reintroducing duplicated submission-specific logic, the centralized score registration flow was adjusted to allow `username = null` specifically for `ActivityType.CHALLENGE_COMPLETED`.

This keeps score persistence centralized while avoiding a second repository-based implementation in the submission flow.

### Automated
- updated `SubmissionServiceImplTest`
- updated `UserScoreServiceImplTest`
- removed tests tied to the deleted submission-specific score recorder path

### Manual
Validated locally with Swagger + Mongo:
- completed submission flow returns success
- a new score record is created in `user_score_history`
- `scores/history` reflects the new persisted score
- `activityType = CHALLENGE_COMPLETED` is persisted
- `challengeId` is persisted for reto completion records
- `created_at` is present in persisted records

## Self-check
- [x] I tested my changes locally and verified the implemented flow
- [x] The PR has a clear and focused purpose
- [x] There are no unrelated changes from previous branches within the intended scope
- [x] The submission scoring flow now uses a single persistence path
- [x] Existing score consumers were checked for compatibility

## Notes
- Draft PR for early review
- Not intended to merge until dependent review/approval is complete
- `CHANGELOG.md` was intentionally cleaned up and is not updated in this PR, following the current project guidelines
- Remaining discussion points include whether `username` should later be sourced in the submission flow and whether stronger score-level idempotency should be added in the centralized persistence layer

## Current review alignment
This PR addresses the review concern about having two persistence paths for score registration.

There is now a single persistence path through `UserScoreService.registerPoints(...)`.

The remaining open point is idempotency. The current submission flow already prevents modifying submissions once they are completed, which reduces duplicate risk in normal usage, but there is not yet a dedicated score-level safeguard in the centralized persistence path.

Captura de pantalla 2026-04-12 a las 11 19 03
Captura de pantalla 2026-04-12 a las 11 19 30
Captura de pantalla 2026-04-12 a las 11 20 11


JordiCasas87 (commented):
I think this is an important point to address before merging.

Score persistence is currently happening through two different paths: UserScoreService.registerPoints(...) (with validations and business rules) and direct repository usage in SubmissionScoreRecorderImpl. This duplicates logic and may lead to inconsistent data (e.g. username is required in one flow but null in the other).

To keep a single source of truth, it would be better to delegate score persistence to UserScoreService.registerPoints(...) instead of saving directly via the repository.

Also, score recording is triggered whenever a submission reaches SUBMITTED_COMPLETE, with no apparent safeguard against duplicates (e.g. retries or repeated actions). It would be worth ensuring idempotency, for example by checking if a score already exists before persisting a new one.

I understand this PR depends on ongoing changes in the shared score model, so this might be partially influenced by that. Just flagging it to help ensure both sides stay aligned once those changes are finalized.

andyrouge81:
Update after refactor:

The submission scoring flow no longer uses a separate repository-based bridge. Completed submissions now route score persistence through `UserScoreService.registerPoints(...)`, so there is a single persistence path aligned with the shared gamification model.

At this point, the remaining failing CI checks appear to be outside the scope of this PR:
- `ChallengeServiceImplTest.getChallengesByTopic_WhenErrorOccurs_ReturnsError`
- `ResourceServiceImplTest.createResource_WithFailedConversion_ShouldThrowError`

These failures belong to unrelated challenge/resource service areas, so I am intentionally not mixing those fixes into this PR.

JordiCasas87 (approved):
Looks good to me.

Nice one centralizing everything through UserScoreService.registerPoints(...), much cleaner now.

Just keep in mind idempotency when you align this with the other PR.

andyrouge81:
I investigated the current failing CI checks locally and was able to reproduce them outside the submission scoring flow touched in this PR.

The failures appear to come from unrelated reactive tests in `ChallengeServiceImpl` (and likely the same area for the second failing test), not from the submission scoring changes introduced here. After adjusting that reactive flow locally, the failing tests passed on my machine.

I’m intentionally not including that fix in this PR to avoid mixing scopes, since this PR is focused on routing completed submission scoring through the centralized `UserScoreService.registerPoints(...)` path.

So at this point, the remaining CI failure looks external to the scope of this PR rather than caused by the submission scoring refactor itself.
Captura de pantalla 2026-04-13 a las 16 38 05

andyrouge81:
Final update from my side:

I aligned the shared gamification-related files with the implementation already being used in the companion branch, in order to reduce divergence between both PRs and avoid keeping two incompatible versions of the same shared score flow while trying to get Sonar/CI back to green.

On my side, I also revalidated the submission flow locally:
- completed submission scoring is being persisted
- the submission path goes through the centralized score service
- persistence was verified manually in Swagger / Mongo

That said, at this point the PR has grown too much in touched files and corrected lines compared to the original scope of card #275. Because of that, I don’t think it is a good idea to keep pushing more changes on top of this branch.

My recommendation would be to stop here, realign on scope, and decide whether to:
- reduce this PR back to the submission persistence change only, or
- rebuild the change cleanly from a smaller base branch

andyrouge81:
Update from my side:

I removed the out-of-scope activity endpoint files that had been introduced while trying to align shared gamification code, since they were not part of card #275 and were adding noise to the branch:
- `ActivityController`
- `ActivityPointsRequest`
- `ActivityPointsResponse`
- related DTO/controller tests

I validated that cleanup separately in a backup branch, and the module still builds successfully locally after removing those files. I’m also attaching a local test run screenshot as evidence that the module tests pass in my environment.

At this point, the remaining CI / Sonar failure is no longer related to those out-of-scope activity files. The current failing checks are now pointing to `UserScoreRepositoryIntegrationTest` and the `activityType` non-null constraint in `UserScoreDocument`.

So from my side:
- the submission scoring persistence flow has been revalidated locally
- the out-of-scope activity endpoint noise has been removed
- the remaining blocker is now isolated to the shared gamification repository/integration test area rather than the removed activity endpoint files
- 
Captura de pantalla 2026-04-15 a las 11 49 10

andyrouge81:
Quick update: the previous `UserScoreRepositoryIntegrationTest` issue has already been addressed, and the current failing CI checks are now pointing again to tests outside the scope of this PR, specifically in `ChallengeServiceImpl`.

andyrouge81:
Update from my side:

I aligned the shared gamification code and related tests with the implementation currently being used by my teammates, so the submission scoring flow and the shared score contract are now consistent within this PR branch.

This includes aligning both production code and supporting tests around the centralized score persistence path used by submission completion.

At this point, the branch content has been updated accordingly, but Sonar / CI is still pending because the PR still needs to be fully aligned with the current `develop` branch state.

